PR #4, opened 2019-10-31
Add Ruby Gem example

koogawa:
Add Ruby Gem example to README 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,16 @@ jobs:
     key: ${{ runner.os }}-maven
 ```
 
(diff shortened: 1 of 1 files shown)


aried3r:
Hey! That's awesome to see :)

I was wondering, is that the complete example because I saw that Rails is then setting the cache path in the next step. Is that implied (also for npm and others) or should this be added as a necessary step?

https://github.com/rails/rails/commit/55f9c8a147701f864602a16f0a6eb30054806b25

PR #13, opened 2019-10-31
Move examples to their own page

joshmgross:
Until we have more comprehensive docs at https://help.github.com/, we need examples in this repo. Since README.md was getting cluttered, I've moved them to a new location.

Also updated the `Node - npm` and `Java - Maven` examples to match the other formats.

Resolves https://github.com/actions/cache/issues/9

Issue #3, opened 2019-10-31
Python example


Sytten:
Could you add an example for a python workspace

hugovk:
# pip + pre-commit

Here's an example for caching two directories. Both on a single OS/Python:

```yml
name: Lint

on: [push, pull_request]

jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: [3.7]

    steps:
      - uses: actions/checkout@v1

      - name: pip cache
        uses: actions/cache@preview
        with:
          path: ~/.cache/pip
          key: lint-${{ matrix.python-version }}-pip

      - name: pre-commit cache
        uses: actions/cache@preview
        with:
          path: ~/.cache/pre-commit
          key: lint-${{ matrix.python-version }}-pre-commit
```

You have to be careful the keys are unique between different workflows to avoid collisions!
(See also https://github.com/actions/cache/issues/17.)

PR #14, opened 2019-10-31
Add pip (with virtual environments) example for Python project

Surgo:
This PR adds example for cache Python packages with virtual environments.

Default install path is like `/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/`.
So use virtual environment to install packages into workspace and cache it.

## First time (no cache)

Screenshot 2019-11-01 00 43 47

## Second time (use cache)

Screenshot 2019-11-01 00 47 49

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -120,7 +120,7 @@ uses: actions/cache@preview
       ${{ runner.os }}-mix-
 ```
 
-### Python - pip
+## Python - pip
 
 Use with virtual environments.
 
--- a/examples.md
+++ b/examples.md
@@ -10,7 +10,7 @@
 - [Ruby - Gem](#ruby---gem)
 - [Go - Modules](#go---modules)
 - [Elixir - Mix](#elixir---mix)
-- [Python - pip](#python---pip)
+- [Python - pip, virtuelenv](#python---pip,-virtuelenv)
 
 ## Node - npm
 
@@ -111,6 +111,7 @@ uses: actions/cache@preview
       ${{ runner.os }}-go-
 ```
 
+<<<<<<< HEAD
 ## Elixir - Mix
 ```yaml
 - uses: actions/cache@preview
@@ -122,6 +123,9 @@ uses: actions/cache@preview
 ```
 
 ## Python - pip
+=======
+## Python - pip, virtuelenv
+>>>>>>> Add virtualenv for title and fix wrong format
 
 Use with virtual environments.
 
@@ -129,11 +133,7 @@ Use with virtual environments.
 - uses: actions/cache@preview
   with:
     path: .venv
-    key: ${{ runner.os }}-pip-${{ hashFiles(format('{0}/{1}', github.workspace, '**/requirements.txt')) }}
+    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
     restore-keys: |
       ${{ runner.os }}-pip-
-- run: |
-    python -m venv .venv
-    source .venv/bin/activate
-    python -m pip install -r requirements.txt
 ```
--- a/examples.md
+++ b/examples.md
@@ -123,7 +123,7 @@ uses: actions/cache@preview
 
 ## Python - pip, virtualenv
 
-Use with virtual environments.
+If caching the virtualenv, you have to activate the environment for each tasks.
 
 ```yaml
 - uses: actions/cache@preview
--- a/examples.md
+++ b/examples.md
@@ -125,8 +125,17 @@ uses: actions/cache@preview
 
 ```yaml
 - uses: actions/cache@preview
+  if: startsWith(runner.os, 'Linux')
   with:
     path: ~/.cache/pip
+    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+    restore-keys: |
+      ${{ runner.os }}-pip-
+
+- uses: actions/cache@preview
+  if: startsWith(runner.os, 'macOS')
+  with:
+    path: ~/Library/Caches/pip
     key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
     restore-keys: |
       ${{ runner.os }}-pip-
--- a/examples.md
+++ b/examples.md
@@ -129,7 +129,7 @@ Use with virtual environments.
 - uses: actions/cache@preview
   with:
     path: .venv
-    key: ${{ runner.os }}-pip-${{ hashFiles(format('{0}/{1}', github.workspace, 'requirements.txt')) }}
+    key: ${{ runner.os }}-pip-${{ hashFiles(format('{0}/{1}', github.workspace, '**/requirements.txt')) }}
     restore-keys: |
       ${{ runner.os }}-pip-
 - run: |
--- a/examples.md
+++ b/examples.md
@@ -10,6 +10,7 @@
 - [Ruby - Gem](#ruby---gem)
 - [Go - Modules](#go---modules)
 - [Elixir - Mix](#elixir---mix)
+- [Python - pip](#python---pip)
 
 ## Node - npm
 
--- a/examples.md
+++ b/examples.md
@@ -10,7 +10,7 @@
 - [Ruby - Gem](#ruby---gem)
 - [Go - Modules](#go---modules)
 - [Elixir - Mix](#elixir---mix)
-- [Python - pip, virtuelenv](#python---pip,-virtuelenv)
+- [Python - pip, virtuelenv](#python---pip-virtuelenv)
 
 ## Node - npm
 
@@ -111,7 +111,6 @@ uses: actions/cache@preview
       ${{ runner.os }}-go-
 ```
 
-<<<<<<< HEAD
 ## Elixir - Mix
 ```yaml
 - uses: actions/cache@preview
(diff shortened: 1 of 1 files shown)


imbsky:
The example has been moved to `examples.md`. See #13.

examples.md
```diff
@@ -95,4 +95,21 @@ uses: actions/cache@preview
     key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
     restore-keys: |
       ${{ runner.os }}-gem-
-```
\ No newline at end of file
+```
+
+### Python - pip
```
imbsky: Unify format.

```suggestion
## Python - pip
```

imbsky:
Please add this to the top.

```md
- [Python - pip](#python---pip)
```

examples.md
```diff
@@ -95,4 +95,21 @@ uses: actions/cache@preview
...
+
+## Python - pip
+
+Use with virtual environments.
+
+```yaml
+- uses: actions/cache@preview
+  with:
+    path: .venv
+    key: ${{ runner.os }}-pip-${{ hashFiles(format('{0}/{1}', github.workspace, 'requirements.txt')) }}
```
imbsky:
```suggestion
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```

examples.md
```diff
@@ -95,4 +96,21 @@ uses: actions/cache@preview
...
+    restore-keys: |
+      ${{ runner.os }}-pip-
+- run: |
+    python -m venv .venv
+    source .venv/bin/activate
+    python -m pip install -r requirements.txt
```
imbsky: Remove run section.

Surgo: But `.venv` is created by `python -m venv .venv` command 🤔

imbsky: This is just an example of the cache action API, not how to write a Python workflow setup.

imbsky: See examples in other languages/ecosystems.

imbsky: For an example of using virtualenv, changing to the minimum example code and title to `Python - pip, virtualenv` should enough. (imo)

Surgo: Ummm... `.venv` is not default package install path.
Default path is`/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/` for  `actions/setup-python@v1` with `python-version: 3.7`
Is `/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/` better cache path in examples?

imbsky: See the comment above.

imbsky:
Here is correct format.

```yml
key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```

examples.md
```diff
@@ -8,6 +8,7 @@
 - [Swift, Objective-C - Carthage](#swift-objective-c---carthage)
 - [Swift, Objective-C - CocoaPods](#swift-objective-c---cocoapods)
 - [Ruby - Gem](#ruby---gem)
+- [Python - pip, virtuelenv](#python---pip,-virtuelenv)
```
imbsky:
```suggestion
- [Python - pip, virtuelenv](#python---pip-virtuelenv)
```

imbsky:
If you correct the title link, overall it seems good!

imbsky:
Great! Looks good to me!

examples.md
```diff
@@ -8,6 +8,7 @@
 - [Swift, Objective-C - Carthage](#swift-objective-c---carthage)
 - [Swift, Objective-C - CocoaPods](#swift-objective-c---cocoapods)
 - [Ruby - Gem](#ruby---gem)
+- [Python - pip, virtuelenv](#python---pip-virtuelenv)
```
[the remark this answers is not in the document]
imbsky: Good catch.

philipslan:
One thing to note about caching the virtualenv is that users will have to manually activate the environment for each task. That might be worth adding.

Surgo:
Cache download cache instead of virtualenv like [Travis-CI](https://docs.travis-ci.com/user/caching/#pip-cache)

examples.md
```diff
@@ -108,3 +109,14 @@ uses: actions/cache@preview
     restore-keys: |
       ${{ runner.os }}-go-
 ```
+
+## Python - pip
+
+```yaml
+- uses: actions/cache@preview
+  with:
+    path: ~/.cache/pip
```
hugovk: This only works for Linux, see https://github.com/actions/cache/issues/3#issuecomment-548436296.

Surgo: Oh.. thank you for your advice :)

examples.md
```diff
@@ -119,3 +120,23 @@ uses: actions/cache@preview
...
+      ${{ runner.os }}-pip-
+
+- uses: actions/cache@preview
+  if: startsWith(runner.os, 'macOS')
+  with:
+    path: ~/Library/Caches/pip
+    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+    restore-keys: |
+      ${{ runner.os }}-pip-
+```
```
philipslan: There should also probably be an example for windows here. The path I believe is `path: ~\AppData\Local\pip\Cache` from [pip docs](https://pip.pypa.io/en/stable/reference/pip_install/#caching).

hugovk: Confirmed: https://github.com/hugovk/pypistats/runs/288196149

imbsky:
```yml
- uses: actions/cache@preview
  if: startsWith(runner.os, 'Windows')
  with:
    path: ~\AppData\Local\pip\Cache
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

joshmgross:
With a step like:
```yaml
    - name: Get pip cache
      id: pip-cache
      run: |
        python -c "from pip._internal.locations import USER_CACHE_DIR; print('::set-output name=dir::' + USER_CACHE_DIR)"
```
We can get the cache directory from `pip`, so that we don't need to hardcode it and we can work universally.

Once https://github.com/actions/cache/issues/39 is fixed, we can remove the extra Windows action as we'll be able to use `/` instead of `\` inside `hashFiles` (that fix should be rolling out soon)

Thus the example would be something like

```yaml
- name: Get pip cache
   id: pip-cache
   run: |
     python -c "from pip._internal.locations import USER_CACHE_DIR; print('::set-output name=dir::' + USER_CACHE_DIR)"

- uses: actions/cache@v1
  if: startsWith(runner.os, 'Windows')
  with:
    path: ${{ steps.pip-cache.outputs.dir }}
    key: ${{ runner.os }}-pip-${{ hashFiles('**\requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-

- uses: actions/cache@v1
  if: !startsWith(runner.os, 'Windows')
  with:
    path: ${{ steps.pip-cache.outputs.dir }}
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

hugovk:
Note the underscore before internal, and the word internal: pip has no public API. 

* https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program

And it did change fairly recently:

* https://stackoverflow.com/a/48956368/724176

joshmgross:
@hugovk Good point, we shouldn't recommend using an internal API in our docs.

@Surgo could you add a Windows example following like:

```yaml
- uses: actions/cache@v1
  if: startsWith(runner.os, 'Windows')
  with:
    path: ~\AppData\Local\pip\Cache
    key: ${{ runner.os }}-pip-${{ hashFiles('**\requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

If you're busy or unable to continue this PR, please let me know and I can pick it up. Thanks again for all the work you've done!

shadowmoose:
Not sure if this is useful for anyone here or at Issue #3, but I've cobbled together a cross-platform cache workflow that uses one preliminary step to find the pip library location, using built-in Python libraries & supported methods. 

It seems to work on all my projects. It also avoids adding extra skipped build steps.

```yml
name: CacheTest

on:
  push

jobs:
  run-tests:
    name: ${{ matrix.os }}, ${{ matrix.python-version }}
    runs-on: ${{ matrix.os }}-latest
    strategy:
      max-parallel: 3
      matrix:
        os: [ubuntu, windows, macOS]
        python-version: [3.5, 3.6, 3.7]

    steps:
      - uses: actions/checkout@v1
      - name: Set up Python
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
      
      - name: Get Python Location
        run: |
          echo "::set-output name=path::$(python -c 'import site; print(site.getusersitepackages())')"
        id: get_location

      - name: Activate Cache
        uses: actions/cache@preview
        with:
          path: ${{ steps.get_location.outputs.path }}
          key: pytest-${{ matrix.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/**equirement*.txt') }}

      - name: Install Dependencies
        run: |
          python -m pip install --upgrade pip setuptools wheel
          pip install --upgrade -r requirements.txt --user
```

joshmgross:
Resuming the work for PR at https://github.com/actions/cache/pull/86

Surgo:
Sorry my late response 🙇

Issue #9, opened 2019-10-31
Create the examples directory


imbsky:
Currently, PR's have been submitted to add language specific examples (#4, #5, #8, #10), and probably more PRs will be submitted if you approve those. So, if you plan to approve it, you should create an examples directory to keep clean README. What do you think?

chrispat:
We will have more comprehensive documentation going up next week and will have some samples.  I don't want to bloat this repo with a bunch of different samples.

imbsky:
Yes, I agree completely. If we have basic API usage documentation, we don't need an example for each language.

ad-m:
@chrispat, It may be helpful to document an example:
- using Docker, due to its specific, own cache management,
- using cache and "jobs. <job_id> .needs" to separate build from testing performed in parallel (using "jobs.<job_id>.strategy").

chrispat:
@ad-m due to the way docker does its caching I don't think you would want to use this feature to cache docker layers.

ad-m:
In this case, the Docker cache issue in GitHub Actions remains open. Should this be an issue in this repository or in repository ```@actions/toolkit```?

joshmgross:
@ad-m I think it's fine to leave that in `@actions/toolkit`, as it's a general improvement to Docker actions. If the recommended step is to just cache before using every docker action, then that might be better if done by default.

PR #18, opened 2019-10-31
Add Go modules example

crazy-max:
Workflow example: https://github.com/crazy-max/echo-ipfilter/commit/715af2dd90fbf9b9559aa21e6071a4913634c891/checks?check_suite_id=290272723

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -96,3 +96,14 @@ uses: actions/cache@preview
     restore-keys: |
       ${{ runner.os }}-gem-
 ```
+
+## Go - Modules
+
+```yaml
+- uses: actions/cache@preview
+  with:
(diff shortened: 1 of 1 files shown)


imbsky:
Please add this to the top.

```yml
- [Go - Modules](#go---modules)
```

crazy-max:
@imbsky @joshmgross Done
But maybe we could use the [TOC action](https://github.com/marketplace/actions/toc-generator)?

imbsky:
This example is temporary and will be removed in the future.

crazy-max:
@imbsky Understood, you plan to place each example in a dedicated folder maybe?

imbsky:
See #9.

PR #21, opened 2019-10-31
Prevent commands from executing during tests

joshmgross:
When running tests that check for error cases, calls to `core.error` or `core.warning` are made, which internally execute commands on the runner before being output to stdout (see [`IssueCommand`](https://github.com/actions/toolkit/blob/master/packages/core/src/command.ts#L19))

The `::warning::` and `::error::` commands create annotations, which pollute the run logs and are misleading since these are expected failures from testing.

From: https://github.com/actions/toolkit/issues/118, we can globally mock `process.stdout.write` to alter any `::` commands so that they aren't mistakenly executed during testing.


jest.config.js
```diff
@@ -8,4 +8,15 @@ module.exports = {
...
   verbose: true
-}
\ No newline at end of file
+}
+
+const processStdoutWrite = process.stdout.write.bind(process.stdout)
+process.stdout.write = (str, encoding, cb) => {
+  // We don't want :: commands to be executed by the runner during tests
+  // Replace any :: with :
+  if (!str.match(/^::/)) {
```
yacaovsnc: Do we want anything written out to stdout by the actual process?  What if we just drop everything on the floor?  Any value in capturing logs?

joshmgross: I think it's helpful to have some output for debugging test errors or writing tests

yacaovsnc: In that case, I would either:
1. Add some prefix to the line, to indicate clearly this is from the actual runner, so that I can easily grep for testing output or proram output; or
2. Pipe everything to a different place, such as a file.

Personally, I don't like to see my testing result and the actual program's output mingled together.   If we add some prefix, will it prevent `::error::`s?

yacaovsnc: Also, what about `stderr`?

joshmgross: Good point, I've switched to dropping all of the actual program's output (aka any `::` command)
Dropping absolutely everything resulting in losing the `--coverage` table, and the jest `--silent` option doesn't handle silencing `process.stdout.write`

yacaovsnc: Seems like we don't want to mess with stdout too much then.

PR #25, opened 2019-10-31
Fix README.md

peter7z:
`restore-keys` had incorrect indentation.


nathany:
Wow, macOS tests still queued? They really need a `[skip ci]` option for readme updates.

joshmgross:
>Wow, macOS tests still queued? They really need a [skip ci] option for readme updates.

Good point 😄 
Removing those here: https://github.com/actions/cache/pull/28
I'll follow up with the team on why `macOS-latest` is seeing delays

PR #26, opened 2019-10-31
Remove cache checksum debug - close #24

### LINKED ISSUE #24: Remove cache checksum debug
For debugging cache integrity, code was added to calculate the checksum of the cache during restore and save (see https://github.com/actions/cache/blob/master/src/restore.ts#L100)

Since we can't conditionally run this in debug mode only, it's run on every use of the action and is a waste of time. We should remove that block of code.

ad-m:


joshmgross:
Thanks! We calculate the checksum at post save action as well, see https://github.com/actions/cache/blob/master/src/save.ts#L70 
Could you remove that as well?

ad-m:
>  Could you remove that as well?

I'll try! I hope that I will be able to make this change, convince the GitHub team to accept the change I propose and I will contribute to the development of GitHub.

PR #27, opened 2019-10-31
Readme: fix restore-keys indentation

nathany:
This caused me some pain trying to identify the issue, but the other examples all have `restore-keys` indented.


peter7z:
Duplicate of #25

PR #35, opened 2019-11-01
Update action name to Dependency Caching

joshmgross:
For publishing to the marketplace, we need a more unique name that isn't taken by an existing action, user or organization name. Verified in the marketplace that a "Dependency Caching" doesn't already exist, and user/org names can't contain spaces as far as I'm aware.


joshmgross:
Published to the marketplace, didn't need a new name: https://github.com/marketplace/actions/cache

PR #37, opened 2019-11-02
Add docker layer caching example

peter-evans:
This is a working example for Docker layer caching. Not sure if it will be considered a bit too complicated for these examples. The complication is mainly due to splitting the image into chunks to workaround the 200MB limit.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -174,7 +174,7 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
     if [ -d "docker-cache" ]; then
       cat docker-cache/x* > my-image.tar
       docker load < my-image.tar
-      rm -fr docker-cache
+      rm -rf docker-cache
     fi
 - name: Build image
   if: steps.cache.outputs.cache-hit != 'true'
--- a/examples.md
+++ b/examples.md
@@ -1,6 +1,7 @@
 # Examples
 
 - [C# - Nuget](#c---nuget)
+- [Docker](#docker)
 - [Elixir - Mix](#elixir---mix)
 - [Go - Modules](#go---modules)
 - [Java - Gradle](#java---gradle)
@@ -156,4 +157,30 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
     key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
     restore-keys: |
       ${{ runner.os }}-pods-
+```
+
+## Docker
+
+```yaml
+- uses: actions/cache@preview
+  id: cache
+  with:
+    path: docker-cache
+    key: ${{ runner.os }}-docker-${{ hashFiles('**/Dockerfile') }}
+    restore-keys: |
+      ${{ runner.os }}-docker-
+- name: Load cached Docker layers
+  run: |
+    if [ -d "docker-cache" ]; then
+      cat docker-cache/x* > my-image.tar
+      docker load < my-image.tar
+      rm -fr docker-cache
+    fi
+- name: Build image
+  if: steps.cache.outputs.cache-hit != 'true'
+  run: |
+    docker build --cache-from my-image -t my-image .
+    docker save my-image > my-image.tar
+    mkdir docker-cache
+    split -b 100m my-image.tar docker-cache/x
 ```
--- a/examples.md
+++ b/examples.md
@@ -166,7 +166,7 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
   id: cache
   with:
     path: docker-cache
-    key: ${{ runner.os }}-docker-${{ hashFiles('**/Dockerfile') }}
+    key: ${{ runner.os }}-docker-${{ github.sha }}
     restore-keys: |
       ${{ runner.os }}-docker-
 - name: Load cached Docker layers
--- a/examples.md
+++ b/examples.md
@@ -162,7 +162,7 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
 ## Docker
 
 ```yaml
-- uses: actions/cache@preview
+- uses: actions/cache@v1
   id: cache
   with:
     path: docker-cache
--- a/examples.md
+++ b/examples.md
@@ -180,7 +180,7 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
   if: steps.cache.outputs.cache-hit != 'true'
   run: |
     docker build --cache-from my-image -t my-image .
-    docker save my-image > my-image.tar
+    docker save my-image $(docker history -q my-image | awk '!/<missing>/{print}') > my-image.tar
     mkdir docker-cache
     split -b 100m my-image.tar docker-cache/x
 ```


abatilo:
Just wanted to say, this is amazing and thank you for figuring this out.

abatilo:
Is there a way to use this for docker images that are built and ran as actions themselves?

Like when you use:
`uses: ./yourpath`

peter-evans:
@abatilo I don't think that is possible with this method, but the alternative for that is to pre-build container actions and link to the container in `action.yml`.

```
runs:
  using: 'docker'
  image: 'docker://myorg/my-image:1.0.0'
```

abatilo:
@peter-evans
But either you'll have to push it somewhere public since you can't do private repos yet, or you are spending the time to pre-build every run anyways, no?

steebchen:
What does the `if: steps.cache.outputs.cache-hit != 'true'` part do exactly? Wouldn't it skip the actual build action even though just the Dockerfile hashes are the same? Maybe I don't understand it, but it seems like this not what you usually want, for example for running tests you would want to execute the last step regardless of what was cached or not.

peter-evans:
@abatilo Yes, I think you're right. Currently, if you want to keep a container action private I don't think you have a choice but to rebuild it each time.

peter-evans:
> What does the if: steps.cache.outputs.cache-hit != 'true' part do exactly? Wouldn't it skip the actual build action even though just the Dockerfile hashes are the same? Maybe I don't understand it, but it seems like this not what you usually want, for example for running tests you would want to execute the last step regardless of what was cached or not.

@steebchen The build is skipped if there is a cache hit because the exact Docker image will then be already loaded into docker. The tests should run fine I think. If there is not an exact match (for example one extra layer is added) then the cache is restored and the build uses layers that match from cache.

steebchen:
I may have understood wrong but I am pretty sure that the cache checks are for the cache key, and your cache key only looks for Dockerfiles. This means that in this case, your tests would run on the first push, and then would __only__ run again when you change a dockerfile, not if you change your tests or code (code changed, but your cache check only checks Dockerfiles, thus the tests won't run).

The official example for caching looks like this:

```yaml
  - uses: actions/cache@preview
    id: cache
    with:
      path: path/to/dependencies
      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles')}}
  
  - name: Install Dependencies
    run: /install.sh
```

This makes sense, because when lockfiles are not changed, we can be sure we already have the exact dependencies, though we still need to run our actual tests regardless of whether the install dependencies steps was run (miss) or not (hit).

tuler:
Having the same Dockerfile does not mean you don’t need to rebuild. It depends on what the Dockerfile does. If for example you do a `COPY . .`, a change to any file will change the docker image.

peter-evans:
@steebchen @tuler You are both right, I've realised caching on the Dockerfile doesn't work. In fact, it's actually not necessary to have a cache hit at all, because the cache restore provides the cached layers docker can use during the build.

It might make sense to get a cache hit in some case though. What about commit SHA?
```
key: ${{ runner.os }}-docker-${{ github.sha }}
```

tuler:
In my test I used the sha. But it really depends though. Keep in mind that using such key will bust the repo limit pretty quick.

They need to fix this limit thing, by charging for used space, or providing a way for user-provided cloud space.

Until then I’ll continue using my own caching strategy at AWS S3.

peter-evans:
Updated the cache key to use the commit SHA. I guess that works fine for the purposes of a generic example.

tuler:
I believe that caching a single layer won't help much.
The correct solution is to cache all non missing layers.

```
docker save my-image $(docker history -q my-image | awk '!/<missing>/{print}') > my-image.tar
```

The `awk` command is to filter out `<missing>` layers.

peter-evans:
@tuler I started with that solution, but it didn't seem to make any difference in my tests. It still was able to fetch all layers of the image from cache with plain `docker save my-image`. Perhaps I misunderstand and it won't work correctly in all cases?

tuler:
@peter-evans you are correct, it will work.

Saving and loading the intermediate images would just replace the pulling (from docker hub typically).

Another side-note is that if you use multi-stage builds you have to save/load each stage image separately, which can be done using `docker build --target builder -t my-image:builder` (for a `builder` stage).

peter-evans:
Updated to cache all non missing layers.

@tuler I think we can agree that this is far from a perfect solution and supporting Docker caching natively would be most ideal. However, for the purposes of a basic working example using `actions/cache` do you think this is acceptable now? I think it might be too complicated to try and cover every use case (e.g. multi-stage builds).

tuler:
@peter-evans agree, although I doubt this will work on real world scenarios with the current limitations. I wish they change that before going live in a couple weeks.

examples.md
```diff
@@ -157,3 +158,29 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
...
+      docker load < my-image.tar
+      rm -fr docker-cache
+    fi
+- name: Build image
+  if: steps.cache.outputs.cache-hit != 'true'
+  run: |
+    docker build --cache-from my-image -t my-image .
+    docker save my-image $(docker history -q my-image | awk '!/<missing>/{print}') > my-image.tar
+    mkdir docker-cache
+    split -b 100m my-image.tar docker-cache/x
```
YashMathur: Had a question about the splitting strategy - wouldn't the entire `docker-cache` directory be cached as an individual cache item? 

If that's the case, and if the size of the `docker-cache` directory is over 400mb, then splitting the tar wouldn't help much unless the split files are put into separate folders that are cached using different keys.

peter-evans: I don't think it's necessary to have cache hits because when falling back to cache restore it loads the layers into Docker to be used in its own `--cache--from` mechanism. That's why this example caches the whole directory against the commit SHA.

FabianKoestring:
Is there a way to use this for docker-compose?

adamgeorgeson:
@FabianKoestring I've been testing this approach out with `docker-compose` this week and the following works

#### `workflow.yml`
```
jobs:
  test:
    name: Build and test
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v1
        with:
          fetch-depth: 1
      - name: Load Cache Keys
        uses: actions/cache@v1
        id: cache
        with:
          path: docker-cache
          key: ${{ runner.os }}-docker-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-docker-
      - name: Load cached Docker layers
        run: |
          if [ -d "docker-cache" ]; then
            cat docker-cache/x* > my-app.tar
            docker load < my-app.tar
            rm -rf docker-cache
            rm -rf my-app.tar
          fi
      - name: Build image
        run: |
          docker-compose build test
          docker save my-app $(docker history -q my-app | awk '!/<missing>/{print}') > my-app.tar
          mkdir docker-cache
          split -b 100m my-app.tar docker-cache/x
          rm -rf my-app.tar
      - name: Run Test Suite
        if: success()
        run: docker-compose run --rm test
```

#### `docker-compose.yml`
```
services:
  app: &app
    build:
      ...
      cache_from:
        - ruby:2.6.5
        - my-app:latest
      dockerfile: ./Dockerfile
      args:
        ...
    image: my-app:latest
```

examples.md
sousousore1: docker-cache (total of over 400mb)
└xaa (100mb)
└xab (100mb)
└xac (100mb)
└xad (100mb)
└xae (100mb)

YashMathur: @sousousore1 thanks for explaining. That was the point I was trying to make in my comment. 

I ended up creating a new folder for each file I was going to cache using some bash helper functions

```
folder-a (400mb)
 └ xaa
folder-b (400mb)
 └ xab
folder-c (400mb)
 └ xac
```

After restoring cache, I combined them into one file `cat folder-*/* > docker-cache`

peter-evans: Apologies, I misunderstood the original question. Thanks for explaining. I'll try and reproduce this.

examples.md
```diff
@@ -157,3 +158,29 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
...
+    path: docker-cache
+    key: ${{ runner.os }}-docker-${{ github.sha }}
+    restore-keys: |
+      ${{ runner.os }}-docker-
+- name: Load cached Docker layers
+  run: |
+    if [ -d "docker-cache" ]; then
+      cat docker-cache/x* > my-image.tar
+      docker load < my-image.tar
+      rm -fr docker-cache
```
ScottBrenner:
```suggestion
      rm -rf docker-cache
```

Conventional order?

examples.md
albertorm95: Cache size was increase to 2GB #140

jaschaio:
@tuler you mentioned that you were using your own caching strategy with AWS S3. Any chance you could share an example before docker layer caching will be natively supported?

tuler:
> @tuler you mentioned that you were using your own caching strategy with AWS S3. Any chance you could share an example before docker layer caching will be natively supported?

I download and upload build artifacts from/to S3.

https://github.com/cartesi/machine-emulator/blob/master/.github/workflows/build.yml#L16
https://github.com/cartesi/machine-emulator/blob/master/.github/workflows/build.yml#L29

examples.md
shunkakinoki: Cache size was increased to 5GB

examples.md
hunj: `-fr` as in "for real" :joy:

peter-evans:
This was mostly just a fun experiment for me and I'm not sure that it is actually a good solution for reasons that have been brought out in the comments. It may not even be necessary anymore due to recent cache size increases.

I’m going to close this PR now and I’ll leave it up to those of you that actually have some use for this kind of solution to pick it up and improve it where necessary.

examples.md
tmarwen: Cache size was increased to `10GB`.

PR #38, opened 2019-11-02
Add Haskell Example

hvr:
Cabal's [Nix-style Local Builds](https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html) caches built dependencies in `~/.cabal/store`; caching this folder often makes a difference between  15+ minute build-times or a speedy 1-2 minute job.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -9,6 +9,7 @@
 - [Swift, Objective-C - CocoaPods](#swift-objective-c---cocoapods)
 - [Ruby - Gem](#ruby---gem)
 - [Go - Modules](#go---modules)
+- [Haskell - Cabal](#haskell---cabal)
 
 ## Node - npm
 
@@ -107,4 +108,13 @@ uses: actions/cache@preview
     key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
     restore-keys: |
       ${{ runner.os }}-go-
+```
+
+## Haskell - Cabal
+
+```yaml
+- uses: actions/cache@preview
+  with:
+    path: ~/.cabal/store
+    key: ${{ runner.os }}-haskell-${{ matrix.ghc }}
 ```
--- a/examples.md
+++ b/examples.md
@@ -118,3 +118,5 @@ uses: actions/cache@preview
     path: ~/.cabal/store
     key: ${{ runner.os }}-haskell-${{ matrix.ghc }}
 ```
+
+Note: `matrix.ghc` denotes the Haskell compiler version (see also https://github.com/actions/setup-haskell)
--- a/examples.md
+++ b/examples.md
@@ -113,7 +113,7 @@ uses: actions/cache@preview
 ## Haskell - Cabal
 
 ```yaml
-- uses: actions/cache@preview
+- uses: actions/cache@v1
   with:
     path: ~/.cabal/store
     key: ${{ runner.os }}-haskell-${{ matrix.ghc }}


examples.md
```diff
@@ -108,3 +109,12 @@ uses: actions/cache@preview
...
       ${{ runner.os }}-go-
 ```
+
+## Haskell - Cabal
+
+```yaml
+- uses: actions/cache@preview
+  with:
+    path: ~/.cabal/store
+    key: ${{ runner.os }}-haskell-${{ matrix.ghc }}
```
joshmgross: What does `matrix.ghc` represent?

Typically we want the key to contain some reference to the dependencies so that we get a new key and cache when dependencies change (see the other examples with `hashFiles`).

examples.md
```diff
@@ -108,3 +109,12 @@ uses: actions/cache@preview
     restore-keys: |
       ${{ runner.os }}-go-
 ```
+
+## Haskell - Cabal
+
+```yaml
+- uses: actions/cache@preview
```
```suggestion
- uses: actions/cache@v1
```

examples.md
hvr: > What does matrix.ghc represent?

the version of the GHC compiler used; with Haskell it's customary to test various versions of the compiler, whose cached artifacts are incompatible to each other, hence using the GHC version as cache-index/key

Fwiw, the concept of "lockfiles" is a different paradigm from the Nix-style store (which is really a hash-indexed cache already) used by cabal.

joshmgross: Yep, you have two main options.

If the path is constant under the home directory, you can use `~` which the action will resolve to `$HOME` on macOS/Linux and `%USERPROFILE%` on Windows.

If it's more complicated, you can use an extra step beforehand to determine the cache location and output that. See https://github.com/actions/cache/blob/master/examples.md#php---composer for an example of that.

hvr: Yeah, using %USERPROFILE% on windows is not a good idea, because it's prone to hitting the anachronistic PATH_MAX limits much more likely; so we we typically set it to something like `C:\SR` to workaround the limit until all tooling is able to safely use more recent Windows' long paths APIs.

The composer approach sounds great; maybe long-term the `setup-haskell` action could provide such an output variable (unfortunately I lack the JavaScript skills to teach `setup-haskell` various things I've already inlined into a workflow file...)

Thanks for the pointer!

joshmgross: I believe the hosted windows runners should have long path support enabled when initialized via:
```powershell
# See https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation
Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled' -Value 1
```

>maybe long-term the setup-haskell action could provide such an output variable
That's a great idea!

>unfortunately I lack the JavaScript skills to teach setup-haskell various things I've already inlined into a workflow file.

Feel free to file an issue https://github.com/actions/setup-haskell, I'm sure the feedback would be welcomed!

joshmgross:
@hvr Is this PR ready to merge?

kutyel:
@joshmgross this can be closed now cause #148 😉

joshmgross:
Closing in favor of #148, thanks for your contributions!

hvr:
@joshmgross @kutyel Btw, why wasn't *this* PR merged? Please see my comments #148 of which I'm quite a bit skeptical :-)

PR #42, opened 2019-11-03
Add Elixir Mix example

ghostrick:
Add Elixir Mix example to examples.md

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -112,7 +112,7 @@ uses: actions/cache@preview
 
 ## Elixir - Mix
 ```yaml
-- uses: actions.cache@preview
+- uses: actions/cache@preview
   with:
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -108,3 +109,13 @@ uses: actions/cache@preview
     restore-keys: |
       ${{ runner.os }}-go-
 ```
+
+## Elixir - Mix
+```yaml
+- uses: actions.cache@preview
```
joshmgross: Should be `uses: actions/cache@preview` (`/` instead of `.`)

ghostrick: Sorry, typo😫

ghostrick: @joshmgross  Fixed it.

PR #56, opened 2019-11-05
Encode cache keys

### LINKED ISSUE #53: 404 when key contains slash
I think this has the same root cause as #43, except I am using my own custom string with slash (`tmp release/benchmark`) and receives 404 when trying to save the cache.

https://github.com/alexlamsl/UglifyJS2/runs/287169751

alexlamsl:
Reading the restore portion of this action:
https://github.com/actions/cache/blob/ce4a52af49a216b8cf2738812233f8a51dee225d/src/cacheHttpClient.ts#L18-L20

The key-separating commas are getting encoded alongside most non-alphanumeric characters as well, so the middleware must be processing the raw key values instead of encoded ones, which it apparently cannot handle.

So let's encode the keys on top to ensure smooth operations - that way we can remove the restrictions on using commas in key names as well.

Fixes #53

### DIFF
--- a/src/restore.ts
+++ b/src/restore.ts
@@ -17,11 +17,15 @@ async function run() {
         );
         core.debug(`Cache Path: ${cachePath}`);
 
-        const primaryKey = core.getInput(Inputs.Key, { required: true });
+        const keys = [
+            core.getInput(Inputs.Key, { required: true }),
+            ...core
+                .getInput(Inputs.RestoreKeys)
+                .split(/\s*\n\s*/)
+                .filter(x => x)
+        ].map(key => encodeURIComponent(key).replace(/%/g, " "));
+        const primaryKey = keys[0];
         core.saveState(State.CacheKey, primaryKey);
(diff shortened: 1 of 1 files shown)


alexlamsl:
@joshmgross I can confirm this now works for the job described in #53

Saving new entry: https://github.com/alexlamsl/UglifyJS2/runs/289304672
new

Restoring entry: https://github.com/alexlamsl/UglifyJS2/runs/289304904
reuse

joshmgross:
My concern with this is that we're manipulating user input, leading to keys that are not expected or conflict with existing keys. There are also security concerns with essentially double encoding the inputs, see https://www.owasp.org/index.php/Double_Encoding

I'd rather fix `/` not working on the server-side, or via a different encoding scheme/API (we have no issues with using encoded `/` in keys as a query parameter.

alexlamsl:
Fair enough - I shall await for the fix then 😉

Issue #6, opened 2019-10-31
The cache limit is too small


smorimoto:
I understand that Issue is not related to the code of this repository, but I would like to discuss with many people, so I open Issue here. (and I know community forums exist. but many people probably don't know yet.)

First, I really appreciate the GitHub team for adding the cache feature to GitHub Actions. That's great for us! But, In recent years, `node_modules` is too large. 200MB can't cover it. It's the same in other languages. For example, using esy to install the opam packages, it can easily exceed 800MB. Is there a way to increase the cache limit? or if individual cache limits are removed, it becomes a relatively realistic limit. I know that if the file size is too large, save/restore may insanely slow down. but it shouldn't be limited on the cache action side.

cemo:
> Individual caches are limited to 200MB and a repository can have up to 2GB of caches. Once the 2GB limit is reached, older caches will be evicted based on when the cache was last accessed.

Yes please remove individiual cache limit. 2GB per repo is reasonable.

dentuzhik:
For our middle-sized react-native project, archived node_modules directory is 260 MB and CocoaPods directory is 202 MB

sidmitra:
Typically python virtualenv are also >500MB these days. These are cached to avoid re-compiling some modules each time like lxml, numpy etc.

imbsky:
In my opinion, 2GB per repository is still small for medium ~ large project.

imbsky:
The file size limit is defined [here](https://github.com/actions/cache/blob/master/src/save.ts#L57), so I think we can easily remove the individual limit. but, we can't do anything about the limit per repository. 2GB is probably enough for JavaScript developer, but not so good for a native developer like me.

hazcod:
I would appreciate a higher limit (e.g. 5BG) for enterprise/paying customers as wel.

imbsky:
Well, my guess is that you will be able to increase the cache limit by charging like Git LFS.

chrispat:
We are working on the long term plan for how we will enable larger limits.  Charging for it like we do for packages or actions artifacts is something we are considering.

lpil:
The compiled deps for my small Rust project is 500 MB, the cache will need to be considerably larger to support Rust.

imbsky:
You will need more if you use OCaml. Anyway, 200MB/2GB is not enough.

andrewhampton:
[...]

PR #61, opened 2019-11-05
Bump cache limit to 400MB

joshmgross:
We've updated the server-side to allow for increasing the individual cache limit to 400MB. We're continuing to work on increasing this limit.

Related: https://github.com/actions/cache/issues/6

Release notes: This will be part of release `v1.0.1`, if you're pinned to `v1` you'll get this updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ See [Examples](examples.md)
 
 ## Cache Limits
 
-Individual caches are limited to 200MB and a repository can have up to 2GB of caches. Once the 2GB limit is reached, older caches will be evicted based on when the cache was last accessed.  Caches that are not accessed within the last week will also be evicted.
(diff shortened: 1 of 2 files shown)


ad-m:
Is it possible for this value to be passed to the agent or downloaded by the agent before attempting to send data? This will allow future configuration of project-specific limits or other limits set by the administrator of the GitHub Enterprise instance.

If data is stored by Azure Storage, this step could also be used to initiate direct upload of data in Azure Storage by using service-level shared access signatures (SAS) ak. pre-signed URL ( https://docs.microsoft.com/en-us/azure/storage/common/storage-stored-access-policy-define-dotnet) with the same level of security, but less network traffic transmission.

ad-m:
Oops, it looks like Azure doesn't support the limit on the size of files for SAS authenticated requests.

jahed:
The documentation needs to be updated here too: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows

zeke:
Fantastic news! We can use this in github/help-docs now!

```
$ du -hs node_modules
250M	node_modules
```

PR #62, opened 2019-11-05
Add unit tests for restore

joshmgross:
Adding unit tests for `restore.ts`, which mocks out most of the HTTP Client and file system utils. WIll follow up in a future PR with tests for save, the client, and the utils.


__tests__/restore.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    expect(stateMock).toHaveBeenCalledWith("CACHE_KEY", key);
+    expect(warningMock).toHaveBeenCalledTimes(0);
+    expect(failedMock).toHaveBeenCalledTimes(0);
+
+    expect(infoMock).toHaveBeenCalledWith(
+        `Cache not found for input keys: ${key}.`
+    );
+});
+
+test("restore with no server error", async () => {
```
yacaovsnc: I am a bit confused - this says "no server error", but the client throws a new `Error`, is this "restore with server error should fail"?

__tests__/restore.test.ts
```diff
@@ -0,0 +1,336 @@
...
+
+    const infoMock = jest.spyOn(core, "info");
+    const warningMock = jest.spyOn(core, "warning");
+    const failedMock = jest.spyOn(core, "setFailed");
+    const stateMock = jest.spyOn(core, "saveState");
+
+    const cacheEntry: ArtifactCacheEntry = {
+        cacheKey: key,
+        scope: "refs/heads/master",
+        archiveLocation: "https://www.example.com/download"
```
yacaovsnc: nit: Could use something like `www.actionscache.test/download` per [RFC 2606](https://tools.ietf.org/html/rfc2606#section-2)

__tests__/restore.test.ts
joshmgross: Yep, copy/paste typo from another test 😄 

Should be "restore with server error should fail"

PR #68, opened 2019-11-07
Provide better errors for unsupported event types

### LINKED ISSUE #64: Unsupported event types should be blocked as verification errors
Currently, only `push` and `pull_request` events work with the cache service, other events will fail server authorization with an error like "No scopes with read permission were found on the request."

We should check the event (from env variable `GITHUB_EVENT_NAME`) and provide a more helpful error message when the event is not supported.

joshmgross:
Currently, only push and pull_request events work with the cache service, other events will fail server authorization with an error like "No scopes with read permission were found on the request."

Failing early during verification makes it clear why reading and restoring the cache is failing.

Resolves #64

### DIFF
--- a/src/restore.ts
+++ b/src/restore.ts
@@ -11,6 +11,12 @@ import * as utils from "./utils/actionUtils";
 async function run() {
     try {
         // Validate inputs, this can cause task failure
+        if (!utils.isValidEvent()) {
+            core.setFailed(
+                `Event Validation Error: The event type ${process.env["GITHUB_EVENT_NAME"]} is not supported. Only \`push\` and \`pull_request\` events are supported at this time.`
+            );
+        }
+
         let cachePath = utils.resolvePath(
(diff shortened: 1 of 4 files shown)


src/restore.ts
```diff
@@ -11,6 +11,12 @@ import * as utils from "./utils/actionUtils";
 async function run() {
     try {
         // Validate inputs, this can cause task failure
+        if (!utils.isValidEvent()) {
+            core.setFailed(
+                `Event Validation Error: The event type ${process.env["GITHUB_EVENT_NAME"]} is not supported. Only \`push\` and \`pull_request\` events are supported at this time.`
```
dhadka: Nit: Replace with `Events.Key`

dhadka: You could also make the supported events a list so there's only one spot that needs to be updated when we add new events.

Issue #67, opened 2019-11-07
Recommend caching ~/.npm rather than node_modules


billyjanitsch:
The readme currently provides [an example workflow](https://github.com/actions/cache/tree/4b0709a0d5bb7cd20f2ca9b9eb3dd620875faec9#example-workflow) that suggests caching a project's `node_modules` folder.

This is generally not recommended: see [here](https://docs.npmjs.com/cli/ci.html#example), [here](https://stackoverflow.com/questions/42521884/should-i-have-travis-cache-node-modules-or-home-npm), [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#nodejsnpm), etc. It also doesn't integrate well with npm's suggested CI workflow -- which is to cache `~/.npm` and use `npm ci` -- because `npm ci` always removes node_modules if it exists so caching it strictly slows down the build.

For reference, Azure and Travis both cache `~/.npm` rather than `node_modules` by default.

Was this a conscious design decision? If not, would you consider revising the readme to suggest using the shared cache?

joshmgross:
We'll update the example, thanks for the tip!

FND:
I'm not sure whether this belong here, but GitHub's help pages still use `node_modules` as an example: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#using-the-cache-action

Starting out, that's what I wanted to use as a reference until I - pretty much by accident - stumbled across this change. So it might be worth updating that article as well.

joshmgross:
@FND Thanks! We're working on getting that doc example fixed, cc: @chrispat

PR #69, opened 2019-11-07
Remove recommendation to cache node_modules

joshmgross:
Resolves https://github.com/actions/cache/issues/67

It's not recommended to cache `node_modules`, so while it may work for basic examples it can cause issues and doesn't work with `npm ci`. `npm` has a folder that we should cache instead, see https://docs.npmjs.com/cli/cache#cache

Added simple examples for Ubuntu, macOS, and Windows as well as a cross-platform scenario that should work for all three.

Also updated the Example Workflow to be a more generic example of caching

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 ### Example workflow
 
 ```yaml
-name: Example Caching with npm
+name: Caching Primes
 
 on: push
 
@@ -39,22 +39,19 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
-    - name: Cache node modules
+    - name: Cache Primes
+      id: cache-primes
       uses: actions/cache@v1
       with:
-        path: node_modules
-        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-        restore-keys: |
-          ${{ runner.os }}-node-
+        path: prime-numbers
+        key: ${{ runner.os }}-primes
 
-    - name: Install Dependencies
-      run: npm install
+    - name: Generate Prime Numbers
+      if: steps.cache-primes.outputs.cache-hit != 'true'
+      run: /generate-primes.sh -d prime-numbers
     
-    - name: Build
-      run: npm run build
-
-    - name: Test
-      run: npm run test
+    - name: Use Prime Numbers
+      run: /primes.sh -d prime-numbers
 ```
 
 ## Ecosystem Examples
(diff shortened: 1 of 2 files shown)


billyjanitsch:
Thanks @joshmgross, much appreciated!

johnterickson:
Since you cited Pipeline Caching docs, I wanted to mention that we are in the process of changing the example to cache node_modules.  We're not seeing any real impact in caching npm's cache - i.e. it's not the fetching of packages that is slow.

That said, based on the research and links, we'll want to include the node version in the key.

johnterickson:
In summary, I'm thinking:
1. Include `node --version` (and `npm --version`?) in the cache key
2. We want to avoid `npm install` and use `npm ci` instead
3. Include `packages.json` in the key because
> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

4. On cache hit don't call `npm install` or `npm ci`

joshmgross:
@johnterickson Thanks for the info! Definitely seems like caching `node_modules` is fastest if it works

I would be happy to talk through caching scenarios.

>We want to avoid npm install and use npm ci instead

Worth mentioning that `npm ci` is quite a bit faster than `npm install`, but will wipe `node_modules` before running installs.

>On cache hit don't call `npm install` or `npm ci`

I don't remember how pipelines caching works, but this will break for partial cache hits (like on matches restore keys instead). This also assumes a good key, which will be fine if `package-lock.json` is included.

Based on https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#conditioning-on-cache-restoration, you'd just want this in the scenario of `true`,  and to install on `false` or `inexact`

johnterickson:
You can follow @fadnavistanmay and me tinkering over here: https://github.com/fadnavistanmay/azure-pipelines-caching-yaml/blob/1b02274e7f9d77eed993aee65ae3e6bdc2ec1f1f/node.js-with-npm/azure-pipelines.yml

Build 1: https://dev.azure.com/codesharing-su0/cachesandbox/_build/results?buildId=16725&view=results
Windows 2017: 5m15s

Build 2 (lockfile hit for node_modules): https://dev.azure.com/codesharing-su0/cachesandbox/_build/results?buildId=16736&view=results
Windows 2017: 0m57s

Build 3 (lockfile miss; .npm/cache fallback hit): https://dev.azure.com/codesharing-su0/cachesandbox/_build/results?buildId=16737&view=results
Windows 2017: 4m57s

This data point (package and lock file from vs-code repo) seems to indicate little-to-no value in caching the npm cache folder. That said more data points would be good.

Issue #60, opened 2019-11-05
Yarn restores cache, but does not install from it


glambert:
Hey team!

First, awesome job on this feature, it will immensely help our CI speed for our JavaScript projects, kudos! 

I've been running on the "over the limit" error for a yarn project with workspaces enabled:

```
Post job cleanup.
/bin/tar -cz -f /home/runner/work/_temp/3c08f6f0-f11f-4d8f-bed5-d491e7d8d443/cache.tgz -C /home/runner/.cache/yarn .
##[warning]Cache size of 231440535 bytes is over the 200MB limit, not saving cache.
```

But when I run the same `tar` command locally, I get a 100.3 MB bundle. Is there anything I'm missing here?

Here's my workflow:

```yaml
name: Test
on:
  push:
    branches:
      - '**'
    tags:
      - '!**'
jobs:
  test:
    name: Test, lint, typecheck and build
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@master
      - name: Dump GitHub context
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
        run: echo "$GITHUB_CONTEXT"
      - name: Use Node.js 10.16.0
        uses: actions/setup-node@v1
        with:
          node-version: 10.16.0
      - name: Cache yarn node_modules
        uses: actions/cache@v1
        with:
          path: ~/.cache/yarn
          key: ${{ runner.OS }}-yarn-${{ hashFiles('**/yarn.lock') }}
          restore-keys: |
            ${{ runner.os }}-yarn-
      - name: Install
        run: yarn install --frozen-lockfile
        # ...
```

Thanks a lot!

glambert:
As a follow-up, if I use `path: node_modules` it does save the cache successfully:

```
Cache yarn node_modules1s
Cache not found for input keys: ["Linux-yarn-db9ae0b40b59cc89d84b533d906606822c72106dfc77455cf7eee198d6e55858","Linux-yarn-"].
Run actions/cache@v1
  with:
    path: node_modules
    key: Linux-yarn-db9ae0b40b59cc89d84b533d906606822c72106dfc77455cf7eee198d6e55858
    restore-keys: Linux-yarn-
  
Cache not found for input keys: ["Linux-yarn-db9ae0b40b59cc89d84b533d906606822c72106dfc77455cf7eee198d6e55858","Linux-yarn-"].
```

...

```
Post Cache yarn node_modules27s
Cache saved successfully
Post job cleanup.
[...]

Issue #16, opened 2019-10-31
Cache multiple paths


hugovk:
It can be useful to cache multiple directories, for example from different tools like pip and pre-commit.

With Travis CI (ignoring their `pip: true` shortcut):

```yml
cache:
  directories:
    - $HOME/.cache/pip
    - $HOME/.cache/pre-commit
```

The Actions equivalent requires adding a duplicated step for each directory, something like:

```yml
      - name: pip cache
        uses: actions/cache@preview
        with:
          path: ~/.cache/pip
          key: ${{ matrix.python-version }}-pip

      - name: pre-commit cache
        uses: actions/cache@preview
        with:
          path: ~/.cache/pre-commit
          key: ${{ matrix.python-version }}-pre-commit
```

Perhaps also allow multiple directories in a single step, something like this?

```yml
      - name: pip cache
        uses: actions/cache@preview
        with:
          path:
            - ~/.cache/pip
            - ~/.cache/pre-commit
          key: ${{ matrix.python-version }}
```

chrispat:
We made a scoping call for the v1 to not support multiple paths.  This is somethign we do plan to look at  for the future.

nathany:
Same requirement for Elixir where we commonly cache multiple paths:

```
paths:
  - _build
  - deps
```

mrmckeb:
This will definitely be required for many open source projects. We want to trial this on some Storybook projects, but caching for a monorepo (Yarn) would be needed for that - along with #6.

But that's not to discourage this great first-step.

Siedlerchr:
This would be very important, for example if you build a web app for android in node, you will add node_modules and also gradle to the cache.

dbrockman:
Would it be possible to use a glob pattern for path?

The user case that I have in mind is caching the typescript build output in a monorepo.

For example:
```
path: packages/*/{build,tsconfig.tsbuildinfo}
```

imbsky:
Is anyone already working on this?

mrmckeb:
It might also be good if this didn't require config for something like Yarn, where you could just say `workspaces: true`.

jcornaz:
>It might also be good if this didn't require config for something like Yarn

I don't think it is good for this project to have tool-specific behavior. I think it should be agnostic of whatever tools we use.

imbsky:
I agree with @jcornaz.

mrmckeb:
[...]

PR #70, opened 2019-11-07
Update yarn cache example

joshmgross:
Caching `~/.cache/yarn` doesn't work for all scenarios, and in the case of https://github.com/actions/cache/issues/60 contained extra files not needed in the cache.

Using `yarn cache dir` retrieves the correct directory based on the current yarn config, see https://yarnpkg.com/lang/en/docs/cli/cache/

For the hosted runners this resolves to:
- windows-latest: `C:\Users\runneradmin\AppData\Local\Yarn\Cache\v6`
- macOS-latest: `/Users/runner/Library/Caches/Yarn/v6`
- ubuntu-latest: `/home/runner/.cache/yarn/v6`

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -80,12 +80,17 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
 ```
 
 ## Node - Yarn
+The yarn cache directory will depend on your operating system and version of `yarn`. See https://yarnpkg.com/lang/en/docs/cli/cache/ for more info.
 
 ```yaml
+- name: Get yarn cache
+  id: yarn-cache
+  run: echo "::set-output name=dir::$(yarn cache dir)"
+
 - uses: actions/cache@v1
   with:
-    path: ~/.cache/yarn
-    key: ${{ runner.os }}-yarn-${{ hashFiles(format('{0}{1}', github.workspace, '/yarn.lock')) }}
+    path: ${{ steps.yarn-cache.outputs.dir }}
+    key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock')) }}
     restore-keys: |
       ${{ runner.os }}-yarn-
 ```
--- a/examples.md
+++ b/examples.md
@@ -90,7 +90,7 @@ The yarn cache directory will depend on your operating system and version of `ya
 - uses: actions/cache@v1
   with:
     path: ${{ steps.yarn-cache.outputs.dir }}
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -80,12 +80,17 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
...
+- name: Get yarn cache
+  id: yarn-cache
+  run: echo "::set-output name=dir::$(yarn cache dir)"
+
 - uses: actions/cache@v1
   with:
-    path: ~/.cache/yarn
-    key: ${{ runner.os }}-yarn-${{ hashFiles(format('{0}{1}', github.workspace, '/yarn.lock')) }}
+    path: ${{ steps.yarn-cache.outputs.dir }}
+    key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock')) }}
```
erictaylor: Extra closing parentheses.
```suggestion
    key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
```

imbsky:
Why set the yarn cache directory instead of `node_modules`? If I understand correctly, it's intended for development and offline environments.

joshmgross:
@imbsky It's not recommended to cache `node_modules`, as you can miss install steps, break dependencies between node versions, and cause other issues. See https://github.com/actions/cache/issues/67 for more info

oscar-b:
Yarn recommends caching both .yarn cache and node_modules. 

See the GitLab example: https://yarnpkg.com/lang/en/docs/install-ci/

imbsky:
@joshmgross I see.

@oscar-b I didn't know that, thanks for the useful link!

oscar-b:
The tricky part is handling yarn workspaces, which uses multiple node_modules folders. But that should be covered by multi-folder caching I guess.

mrmckeb:
@oscar-b I think that it would be better not to require users to do that. Imagine projects like Babel and Storybook - they'd need huge scripts just for caching.

imbsky:
Isn't that why he is making such a proposal https://github.com/actions/cache/issues/16#issuecomment-551202039 ?

joshmgross:
Going to stick with the yarn cache, for now, once we add multiple paths support (see #16) then we can update the example to add `node_modules`. With workspaces, we can't assume `node_modules` is the correct location (could be `workspace-a/node_modules`), so at the very least the yarn cache will cover all installs.

PR #75, opened 2019-11-08
Update action's description

ethomson:


action.yml
```diff
@@ -1,5 +1,5 @@
-name: 'Cache'
-description: 'Cache dependencies and build outputs to improve workflow execution time'
+name: 'Cache Artifacts'
```
joshmgross: @chrispat Do we want to change the name? I'm worried that "Artifacts" is overloaded and the docs specifically separate the two concepts: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#comparing-artifacts-and-dependency-caching

As a user, "cache artifacts", "upload artifacts", and "download artifacts" sound like they should be very similar yet "cache" is fundamentally different in use and function.

ethomson: We're not changing the name right now.

joshmgross: @ethomson "Cache Artifacts" is what will show up in the Marketplace with the `action.yml` change to `name: 'Cache Artifacts'`

ethomson: Sorry, to clarify: we're not going to merge this PR as-is.  We don't want a name change.

joshmgross:
@ethomson and @chrispat, do we still want these description updates? And have we verified if an Action name update will break marketplace links?

chrispat:
Yes we should make these changes and changing the name will not break the marketplace.

PR #86, opened 2019-11-12
Add pip examples

### LINKED ISSUE #3: Python example

joshmgross:
Picking up from https://github.com/actions/cache/pull/14

Resolves #3 

Thanks @surgo for starting this work!

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -8,6 +8,7 @@
 - [Node - npm](#node---npm)
 - [Node - Yarn](#node---yarn)
 - [PHP - Composer](#php---composer)
+- [Python - pip](#python---pip)
 - [Ruby - Gem](#ruby---gem)
 - [Rust - Cargo](#rust---cargo)
 - [Swift, Objective-C - Carthage](#swift-objective-c---carthage)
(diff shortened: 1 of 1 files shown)


hugovk:
Another idea, similar to @shadowmoose's https://github.com/actions/cache/pull/14#issuecomment-552178396:

> You can provide `--cache-dir` or set the `PIP_CACHE_DIR` environment variable to whatever path you like and it should be respected.

https://github.com/pypa/pip/pull/6391#issuecomment-552227976

joshmgross:
@hugovk I saw that, thanks for following up with the pip team!

The only concern with setting the cache directory is that the user will have to determine where they want their cache, and requires setting the cache directory for every pip command they run (which could be easy to forget).

hugovk:
You're welcome!

The env var might help with that, but I agree and definitely prefer to let pip do its thing with no changes.

hugovk:
> Edit: Requested a `pip cache dir` command at [pypa/pip#7350](https://github.com/pypa/pip/issues/7350).

Now released in pip 20.1! https://github.com/actions/cache/pull/285

PR #89, opened 2019-11-12
Fix lint on Windows

joshmgross:
- Fix lint on windows
- Remove namespaces in favor of enums, see https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-namespace.md


imbsky:
Great!

PR #93, opened 2019-11-13
Add unit tests for actionUtils

joshmgross:
Following up from https://github.com/actions/cache/pull/62

Adding unit tests for utilities in this action (`actionUtils`)


__tests__/actionUtils.test.ts
```diff
@@ -0,0 +1,226 @@
...
+});
+
+test("isExactKeyMatch with undefined cache entry returns false", () => {
+    const key = "linux-rust";
+    const cacheEntry = undefined;
+
+    expect(actionUtils.isExactKeyMatch(key, cacheEntry)).toBe(false);
+});
+
+test("isExactKeyMatch with undefined cache entry key returns false", () => {
```
yacaovsnc: ...with `empty` cache entry key...

__tests__/actionUtils.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const homedir = jest.requireActual("os").homedir();
+    const homedirMock = jest.spyOn(os, "homedir");
+    homedirMock.mockImplementation(() => {
+        return homedir;
+    });
+
+    const resolvedPath = actionUtils.resolvePath(filePath);
+
+    const expectedPath = path.join(homedir, ".cache/yarn");
```
yacaovsnc: Does this work on windows?  `/` isn't replaced with `\`?

joshmgross: This will replace `/` with `\` on Windows, which matches the behavior of `actionUtils.resolvePath`

yacaovsnc: Oh oh, I missed it's `path` join.

PR #108, opened 2019-11-20
Remove validation failures and warning annotations

joshmgross:
With https://github.com/actions/cache/pull/68, unsupported event types throw errors instead of failing during save. This breaks multi-event workflows that use the cache (such as a workflow that runs on schedules and push), as the scheduled build will now break due to the cache step failing.

While an `if` statement could be used, requiring that for a common workflow scenario can get verbose, especially as more events are added.

Example using `if`:
```yaml
   - name: My Cache
      if: github.event_name == 'push' || github.event_name == 'pull_request'
      uses: actions/cache@v1
      with:
        path: .github
        key: ${{ runner.os }}-github
```

This PR changes the validation error into a warning and returns early.

Also updated to prefix warnings with a `[warning]` and log them normally,  which will allow the highlighting in the logs without creating annotations.
Example using the `[warning]` prefix:
Screen Shot 2019-11-20 at 1 27 21 PM

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -162,6 +162,16 @@ test("getCacheState with valid state", () => {
     expect(getStateMock).toHaveBeenCalledTimes(1);
 });
 
+test("logWarning logs a message with a warning prefix", () => {
+    const message = "A warning occurred.";
+
+    const infoMock = jest.spyOn(core, "info");
+
+    actionUtils.logWarning(message);
+
+    expect(infoMock).toHaveBeenCalledWith(`[warning]${message}`);
+});
+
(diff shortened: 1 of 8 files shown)


whiskeysierra:
Will this be released soon?

joshmgross:
@whiskeysierra yep!

whiskeysierra:
Great to hear! Will you also update v1 to the latest version?

joshmgross:
@whiskeysierra Yep!

We're following the [Action Versioning Guidelines](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md), so each new release will be an individual tag (like `v1.0.3`) and after final verification, the `v1` tag is pushed to match the latest release. They're both out now.

Issue #112, opened 2019-11-21
Update Ruby-Gem example to set path


dhadka:
The Ruby-Gem example suggests caching the path `vendor/bundle`, but we must also tell the bundler to use this path: `bundle install --path vendor/bundle`.

See for example https://github.com/rails/rails/commit/55f9c8a147701f864602a16f0a6eb30054806b25

Issue identified by https://github.com/actions/cache/issues/111

PR #113, opened 2019-11-22
Update Ruby example in documentation to specify bundler path

evancloutier:
As per the description in #112, we need to specify a path when configuring bundler for dependency installation.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -223,9 +223,9 @@ When dependencies are installed later in the workflow, we must specify the same 
 
 ```yaml
 - name: Bundle install
-        run: |
-          bundle config path vendor/bundle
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -219,6 +219,14 @@ Replace `~/.cache/pip` with the correct `path` if not using Ubuntu.
     restore-keys: |
       ${{ runner.os }}-gem-
 ```
+When dependencies are installed later in the workflow, we must specify the same path for the bundler.
+
+```yaml
+- name: Bundle install
+        run: |
```
dhadka: These lines are indented 6 extra spaces.

PR #124, opened 2019-12-10
Add Another C# Example to use personal cache folder

### LINKED ISSUE #115: NuGet cache on macOS is unexpectedly over cache limit
See [gitfool/Cake.Dungeon](https://github.com/gitfool/Cake.Dungeon) workflow builds for commit [3cc36dc](https://github.com/gitfool/Cake.Dungeon/commit/3cc36dcc81bfcaaf49ebf5fd5ea3275979e642b9):
- [Linux](https://github.com/gitfool/Cake.Dungeon/runs/318467284)
- [Mac](https://github.com/gitfool/Cake.Dungeon/runs/318467288)
- [Windows](https://github.com/gitfool/Cake.Dungeon/runs/318467275)
- [Docker](https://github.com/gitfool/Cake.Dungeon/runs/318470817)

All NuGet caches are ~15 MB, except for the Mac which [warns](https://github.com/gitfool/Cake.Dungeon/runs/318467288#step:10:3) it's ~**4275** MB and so does not save it.

FWIW, the same [build](https://dev.azure.com/gitfool/Cake.Dungeon/_build/results?buildId=346&view=logs&j=a82d9e72-1de4-5bd6-c188-90296135479e&t=6bea8dfa-9275-4f28-821b-965590ba6931&l=31) works as expected on Azure Pipelines.

nogic1008:
Resolved #115

`macos-latest` host runner includes the .NET Core SDK and large libraries such as Xamarin.
Therefore, `~/.nuget/packages` folder is so large that it cannot be cached by `actions/cache`.

This PR changes the C# example to use a personal cache folder.

This is my example workflow and it works fine.
https://github.com/nogic1008/dotnet-ci-example/blob/master/.github/workflows/dotnetcore.yml

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -1,6 +1,6 @@
 # Examples
 
-- [C# - Nuget](#c---nuget)
+- [C# - NuGet](#c---nuget)
 - [Elixir - Mix](#elixir---mix)
 - [Go - Modules](#go---modules)
 - [Java - Gradle](#java---gradle)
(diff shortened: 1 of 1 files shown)


chrispat:
Can you make this an alternate example?  I think both are valid options and we should indicate the difference between the approaches.  For example, I have seen a number of cases where moving the .nuget folder using the NUGET_PACKAGES environment variable breaks targets files embedded in packages that assume they are located at %USERPROFILE%/.nuget.

nogic1008:
@chrispat 
I fixed document to treat "Use Personal Cache Folder" example as another example.

examples.md
```diff
@@ -26,6 +26,21 @@ Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/packa
       ${{ runner.os }}-nuget-
 ```
 
+Depending on the environment, huge packages might be pre-installed in the global cache folder.
+If you do not want to include them, consider to move the cache folder like below.
+>Note: This workflow does not work for projects that require files to be placed in the global.
```
chrispat:
```suggestion
>Note: This workflow does not work for projects that require files to be placed in user profile package folder
```

Issue #91, opened 2019-11-13
Fails on self-hosted Windows runner with "tar --force-local is not supported"


zarenner:
While the hosted virtual environments work just fine, trying to run on a self-hosted windows runner fails today if gnu tar is not already on the path (e.g. via a mingw install or similar).

On newer versions of Windows, BSD tar is included, so you get an error like:
```
C:\Windows\system32\tar.exe -cz --force-local -f C:/Users/zarenner/actions-runner/_work/_temp/8b5e8d5e-a823-42dd-a103-7b47503f2fdb/cache.tgz -C C:/Users/zarenner/actions-runner/_work/zarenner-testperf/zarenner-testperf .
tar.exe: Option --force-local is not supported
Usage:
  List:    tar.exe -tf <archive-filename>
  Extract: tar.exe -xf <archive-filename>
  Create:  tar.exe -cf <archive-filename> [filenames...]
  Help:    tar.exe --help
##[warning]The process 'C:\Windows\system32\tar.exe' failed with exit code 1
```

Note that this isn't especially high priority since:
1. People don't usually wipe self-hosted runners each run making hosted caching less important, and
2. Unless the self-hosted runner happens to co-located with the caching service storage, significant performance gains are unlikely anyways

PR #126, opened 2019-12-12
Use BSD tar on windows

joshmgross:
Resolves https://github.com/actions/cache/issues/91

On the Windows Hosted Runners, GNU tar is prepended to the path from the installation of minGW (http://www.mingw.org/). This version of tar doesn't work out of the box with Windows file paths, and required using `--force-local` and `\` file manipulation to work correctly.

On self-hosted runners, there's no guarantee that `tar` on the PATH will be GNU tar, and BSD tar will fail when encountering the extra `--force-local` option.

This change enforces using the BSD Tar on Windows, allowing for self-hosted runners to use this action. Self-hosted runners without the system tar will fail, but this is better than the current situation where no self-hosted runners work.

Most of the changes in this PR are just moving the `tar` code into its own module to isolate testing and reduce repeated code.

The only behavior change is in `getTarPath`

I verified these changes with a new cache as well as using a cache created by the current `v1` to verify compatibility: https://github.com/joshmgross/cache-canary/runs/346337525

### DIFF
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -1,18 +1,16 @@
 import * as core from "@actions/core";
-import * as exec from "@actions/exec";
-import * as io from "@actions/io";
 import * as path from "path";
 import * as cacheHttpClient from "../src/cacheHttpClient";
 import { Events, Inputs } from "../src/constants";
 import { ArtifactCacheEntry } from "../src/contracts";
 import run from "../src/restore";
+import * as tar from "../src/tar";
 import * as actionUtils from "../src/utils/actionUtils";
 import * as testUtils from "../src/utils/testUtils";
 
-jest.mock("@actions/exec");
-jest.mock("@actions/io");
-jest.mock("../src/utils/actionUtils");
 jest.mock("../src/cacheHttpClient");
+jest.mock("../src/tar");
+jest.mock("../src/utils/actionUtils");
 
 beforeAll(() => {
     jest.spyOn(actionUtils, "resolvePath").mockImplementation(filePath => {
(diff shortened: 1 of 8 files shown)


yacaovsnc:
Do you know if bsd tar will be available on all windows versions?  Including windows servers?

joshmgross:
>Do you know if bsd tar will be available on all windows versions? Including windows servers?

Based on https://docs.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows, it was only added in Build 17063, so older versions of Windows won't have it.
I'm not sure if the Windows server will have it. From speaking with @ericsciple, the various npm packages for `tar` can sometimes be problematic.

The current behavior is that no self-hosted runners will work with `actions/cache` since it's expecting a GNU `tar` prepended to the Path, which is unique to our hosted runner images. This at least allows all self-hosted runners using newer versions of Windows to use this action.

1480c1:
Would it be possible to do a 1.0.4 release with this in it? I am trying to set up self-hosted windows runners for a project, but it always fails to cache anything because of the `--force-local` option not being available on bsdtar and cache ignores the mingw-w64 tar I have first in `PATH`. I tried to do `uses: actions/cache@master` but it did not get the `dist` folder even though it got everything else.

joshmgross:
@1480c1 There will be a release soon with this change. In the meantime, if you fork the repository and run `npm run release`, you should be able to build and commit the `dist` folder and use that fork of the action.

1480c1:
Okay, thank you. Don't know much about npm yet so I wasn't sure how to solve the issue on my own.

PR #128, opened 2019-12-13
Chunked Cache Upload APIs

joshmgross:
To enable larger cache sizes, we're updating our internal APIs to prevent network timeouts that occur from the large single upload stream of the previous iteration.

Bumping the per cache limit to match the per repo limit, as that's the only limit we have now.

### DIFF
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -12,7 +12,7 @@ import {
     ArtifactCacheEntry,
     CommitCacheRequest,
     ReserveCacheRequest,
-    ReserverCacheResponse
+    ReserveCacheResponse
 } from "./contracts";
 import * as utils from "./utils/actionUtils";
 
@@ -113,7 +113,7 @@ export async function reserveCache(key: string): Promise<number> {
     const reserveCacheRequest: ReserveCacheRequest = {
         key
     };
-    const response = await restClient.create<ReserverCacheResponse>(
+    const response = await restClient.create<ReserveCacheResponse>(
         "caches",
         reserveCacheRequest,
         getRequestOptions()
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -147,11 +147,11 @@ async function uploadChunk(
     data: NodeJS.ReadableStream,
     start: number,
     end: number
-): Promise<IRestResponse<void>> {
+): Promise<void> {
     core.debug(
         `Uploading chunk of size ${end -
-        start +
-        1} bytes at offset ${start} with content range: ${getContentRange(
+            start +
+            1} bytes at offset ${start} with content range: ${getContentRange(
             start,
             end
         )}`
@@ -173,14 +173,16 @@ async function uploadChunk(
 
     const response = await uploadChunkRequest();
     if (isSuccessStatusCode(response.statusCode)) {
-        return response;
+        return;
     }
 
     if (isRetryableStatusCode(response.statusCode)) {
-        core.debug(`Received ${response.statusCode}, retrying chunk at offset ${start}.`);
+        core.debug(
+            `Received ${response.statusCode}, retrying chunk at offset ${start}.`
+        );
         const retryResponse = await uploadChunkRequest();
         if (isSuccessStatusCode(retryResponse.statusCode)) {
-            return retryResponse;
+            return;
         }
     }
 
@@ -197,11 +199,11 @@ async function uploadFile(
     // Upload Chunks
     const fileSize = fs.statSync(archivePath).size;
     const resourceUrl = getCacheApiUrl() + "caches/" + cacheId.toString();
-    const responses: IRestResponse<void>[] = [];
     const fd = fs.openSync(archivePath, "r");
 
     const concurrency = Number(process.env["CACHE_UPLOAD_CONCURRENCY"]) ?? 4; // # of HTTP requests in parallel
-    const MAX_CHUNK_SIZE = Number(process.env["CACHE_UPLOAD_CHUNK_SIZE"]) ?? (32 * 1024 * 1024); // 32 MB Chunks
+    const MAX_CHUNK_SIZE =
+        Number(process.env["CACHE_UPLOAD_CHUNK_SIZE"]) ?? 32 * 1024 * 1024; // 32 MB Chunks
     core.debug(`Concurrency: ${concurrency} and Chunk Size: ${MAX_CHUNK_SIZE}`);
 
     const parallelUploads = [...new Array(concurrency).keys()];
@@ -212,7 +214,10 @@ async function uploadFile(
         await Promise.all(
             parallelUploads.map(async () => {
                 while (offset < fileSize) {
-                    const chunkSize = Math.min(fileSize - offset, MAX_CHUNK_SIZE)
+                    const chunkSize = Math.min(
+                        fileSize - offset,
+                        MAX_CHUNK_SIZE
+                    );
                     const start = offset;
                     const end = offset + chunkSize - 1;
                     offset += MAX_CHUNK_SIZE;
@@ -222,14 +227,13 @@ async function uploadFile(
                         end,
                         autoClose: false
                     });
-                    responses.push(
-                        await uploadChunk(
-                            restClient,
-                            resourceUrl,
-                            chunk,
-                            start,
-                            end
-                        )
+
+                    await uploadChunk(
+                        restClient,
+                        resourceUrl,
+                        chunk,
+                        start,
+                        end
                     );
                 }
             })
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -1,7 +1,7 @@
 import * as core from "@actions/core";
 import * as fs from "fs";
 import { BearerCredentialHandler } from "typed-rest-client/Handlers";
-import { HttpClient } from "typed-rest-client/HttpClient";
+import { HttpClient, HttpCodes } from "typed-rest-client/HttpClient";
 import { IHttpClientResponse } from "typed-rest-client/Interfaces";
 import {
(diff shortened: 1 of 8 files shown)


src/contracts.d.ts
```diff
@@ -4,3 +4,16 @@ export interface ArtifactCacheEntry {
...
+export interface CommitCacheRequest {
+    size: number;
+}
+
+export interface ReserveCacheRequest {
+    key: string;
+    version?: string;
+}
+
+export interface ReserverCacheResponse {
```
zarenner: Typo

src/cacheHttpClient.ts
```diff
@@ -83,46 +97,127 @@ async function pipeResponseToStream(
...
+    if (cacheId < 0) {
+        throw new Error(`Unable to reserve cache.`);
+    }
 
-    const response = await restClient.uploadStream<void>(
-        "POST",
-        postUrl,
-        stream,
-        requestOptions
+    // Upload Chunks
```
zarenner: Part of the reason for this API change was so that we could attempt to reserve the cache before even tarring up the cache. Are we still planning to do that?

joshmgross: Yep, I'll move the reserve call to before the tar creation

zarenner:
Implement a (very small number) of retries as discussed, or did you decide not to?

joshmgross:
@zarenner I've added retries to the upload chunk requests

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+    archivePath: string
+): Promise<void> {
+    // Upload Chunks
+    const fileSize = fs.statSync(archivePath).size;
+    const resourceUrl = getCacheApiUrl() + "caches/" + cacheId.toString();
+    const responses: IRestResponse<void>[] = [];
+    const fd = fs.openSync(archivePath, "r");
+
+    const concurrency = 4; // # of HTTP requests in parallel
+    const MAX_CHUNK_SIZE = 32000000; // 32 MB Chunks
```
zarenner: might be nice if these were variables to allow overriding/testing easily.

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+                        restClient,
+                        resourceUrl,
+                        chunk,
+                        start,
+                        end
+                    )
+                );
+            }
+        })
+    );
```
zarenner: Interesting way to do concurrency limiting, but seems fine since node is single-threaded.

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+
+    fs.closeSync(fd);
+
+    const failedResponse = responses.find(
+        x => !isSuccessStatusCode(x.statusCode)
```
zarenner: Don't you throw in `uploadChunk` on invalid response? When will `uploadChunk` return a failed response?

src/restore.ts
```diff
@@ -60,7 +60,7 @@ async function run(): Promise<void> {
 
         try {
             const cacheEntry = await cacheHttpClient.getCacheEntry(keys);
-            if (!cacheEntry) {
+            if (!cacheEntry || !cacheEntry?.archiveLocation) {
```
zarenner: nit: If you're using `?.` can't you just omit the first condition?

src/save.ts
```diff
@@ -47,19 +56,20 @@ async function run(): Promise<void> {
...
+        const fileSizeLimit = 2 * 1024 * 1024 * 1024; // 2GB per repo limit
         const archiveFileSize = utils.getArchiveFileSize(archivePath);
         core.debug(`File Size: ${archiveFileSize}`);
         if (archiveFileSize > fileSizeLimit) {
             utils.logWarning(
                 `Cache size of ~${Math.round(
-                    archiveFileSize / (1024 * 1024)
-                )} MB (${archiveFileSize} B) is over the 400MB limit, not saving cache.`
+                    archiveFileSize / (1024 * 1024 * 1024)
+                )} GB (${archiveFileSize} B) is over the 2GB limit, not saving cache.`
```
zarenner: nit: Seems more useful to me to give the size in MB still rather than GB, rather than "Cache size of ~2 GB..."

src/cacheHttpClient.ts
joshmgross: If you have suggestions then I'm open to changing this, this feels a little *hacky*

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+    let offset = 0;
+    await Promise.all(
+        parallelUploads.map(async () => {
+            while (offset < fileSize) {
+                const chunkSize =
+                    offset + MAX_CHUNK_SIZE > fileSize
+                        ? fileSize - offset
+                        : MAX_CHUNK_SIZE;
+                const start = offset;
+                const end = offset + chunkSize - 1;
```
yacaovsnc: What happens when we try to upload a size 0 file?

src/cacheHttpClient.ts
yacaovsnc: nit: Would `32 * 1024 * 1024` be easier to read in this case?

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+            requestOptions
+        );
+    };
+
+    const response = await uploadChunkRequest();
+    if (isSuccessStatusCode(response.statusCode)) {
+        return response;
+    }
+
+    if (isRetryableStatusCode(response.statusCode)) {
```
yacaovsnc: Probably should log a retry debug line.  If we see a client `retrying` many times, maybe an indication of some other issues?

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+    const parallelUploads = [...new Array(concurrency).keys()];
+    core.debug("Awaiting all uploads");
+    let offset = 0;
+    await Promise.all(
+        parallelUploads.map(async () => {
+            while (offset < fileSize) {
+                const chunkSize =
+                    offset + MAX_CHUNK_SIZE > fileSize
+                        ? fileSize - offset
+                        : MAX_CHUNK_SIZE;
```
pjquirk: Nit: I think this is a little easier to parse, but YMMV:
```js
const chunkSize = Math.min(fileSize - offset, MAX_CHUNK_SIZE)
```

src/restore.ts
```diff
@@ -77,7 +77,10 @@ async function run(): Promise<void> {
             utils.setCacheState(cacheEntry);
 
             // Download the cache from the cache entry
-            await cacheHttpClient.downloadCache(cacheEntry, archivePath);
+            await cacheHttpClient.downloadCache(
+                cacheEntry?.archiveLocation,
```
pjquirk: Nit, the `?` isn't necessary here

src/save.ts
```diff
@@ -47,19 +56,20 @@ async function run(): Promise<void> {
...
             );
             return;
         }
 
-        await cacheHttpClient.saveCache(primaryKey, archivePath);
+        core.debug("Saving Cache");
```
pjquirk: Include `cacheId` and/or `archivePath` in this debug?

src/save.ts
```diff
@@ -34,6 +34,15 @@ async function run(): Promise<void> {
             return;
         }
 
+        core.debug("Reserving Cache");
+        const cacheId = await cacheHttpClient.reserveCache(primaryKey);
+        if (cacheId < 0) {
```
pjquirk: Double-checking, zero is valid here?

dhadka: The cacheId starts incrementing from 1, so the server would never return 0.

src/cacheHttpClient.ts
```diff
@@ -83,46 +107,183 @@ async function pipeResponseToStream(
...
+                        chunk,
+                        start,
+                        end
+                    )
+                );
+            }
+        })
+    );
+
+    fs.closeSync(fd);
```
pjquirk: Does this need to be in a `finally` block? Not sure what the behavior is if, for example, `createReadStream` fails.

src/cacheHttpClient.ts
joshmgross: Good idea, thanks!

src/save.ts
joshmgross: Added `cacheId` to the debug output, we output `archivePath` further up (Archive Path is just a temp directory anyway, the user doesn't need to know it)

src/save.ts
joshmgross: Changed this to explicitly test `if (cacheId == -1)` which is the default return value in the method when no cache ID was sent back.

src/cacheHttpClient.ts
joshmgross: Yep! Since any failed upload will throw, this was never closed in that event

src/cacheHttpClient.ts
joshmgross: Added a debug line below

src/cacheHttpClient.ts
joshmgross: In that case:
`offset = 0`, `chunkSize = fileSize - offset = 1`, `end = 0`, which is valid since the content range is inclusive

src/cacheHttpClient.ts
joshmgross: Yep, removed this section

src/cacheHttpClient.ts
joshmgross: Added env variables to control these two settings, with a fallback to the default values

Issue #59, opened 2019-11-05
Sbt / ivy2 example


niklasf:
Please provide an example for a Scala project built using sbt.

YashMathur:
There may be a better way to do this that involves hashing the files to break it up into multiple caches, but here's what I'm doing right now -
```
- name: Cache Scala artifacts
  uses: actions/cache@v1
  id: cache
  with:
    path: ~/.ivy2/cache
    key: ${{ runner.os }}-ivy2
    restore-keys: |
      ${{ runner.os }}-ivy2
```

tgodzik:
We were looking into it on our PR https://github.com/scalameta/metals/pull/1050, however it seems that jar dependencies can take up a lot of space, so we are unable to use it properly.

joshmgross:
Added with https://github.com/actions/cache/pull/134

PR #134, opened 2019-12-19
Add Scala - SBT example

daniel-shuy:


examples.md
```diff
@@ -11,6 +11,7 @@
 - [Python - pip](#python---pip)
 - [Ruby - Gem](#ruby---gem)
 - [Rust - Cargo](#rust---cargo)
+- [Scala - SBT](#scala---sbt)
```
dhadka: Please also add a link to this in README.md under Implementation Examples.

joshmgross:
Awesome, thanks for this! This will resolve https://github.com/actions/cache/issues/59

daniel-shuy:
Ah, just noticed https://github.com/actions/cache/issues/59#issuecomment-552846834. This example only works for SBT < 1.3 (with Ivy), or SBT 1.3+ with Coursier disabled (which will fallback to Ivy). Seems like the cache size limit is too small for Coursier at the moment.

dhadka:
@daniel-shuy Will this example work with Coursier as-is once the cache size limit is increased?  We expect to rollout the size increase early next year.

daniel-shuy:
> 
> 
> @daniel-shuy Will this example work with Coursier as-is once the cache size limit is increased? We expect to rollout the size increase early next year.

Unfortunately no, the configuration for Coursier is very different

Issue #63, opened 2019-11-06
Allow using the cache for other than push and pull_request events


Xilonz:
When triggering my action with `on: ['deployment']`.  It throws a warning about having no read permissions. 

It doesn't happen with `on: push`.

```
##[debug]Resolved Keys:
##[debug]["Linux-yarn-...hash...","Linux-yarn-"]
##[debug]Cache Url: https://artifactcache.actions.githubusercontent.com/...hash.../
##[warning]No scopes with read permission were found on the request.
::set-output name=cache-hit,::false
```

jahed:
Saw a similar warning for [one of my **scheduled** workflows](https://github.com/jahed/node-terraform/commit/df5413194663618e77730ec8f55f7dc913985629/checks?check_suite_id=298460992). The same workflow ran fine 2 days ago (November 4). It failed yesterday too (November 5) so may be an issue on GitHub's end. Possibly the way permissions are granted to use the cache.

joshmgross:
The cache is currently limited to `push` and `pull_request` events. We're looking into opening it up to more event types in the future.

I've filed an issue internally to get the [docs](https://help.github.com/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows) updated with this information.

jahed:
That makes sense. The runs that did work for me were triggered on `push`.

Instead of updating documentation, may I suggest improving the error message? To say that the action is not supported on the event that triggered it. Saves people having to dig through documentation for special cases.

Ignigena:
Definitely would love to see this open up to other event types, particularly `deployment` and `deployment_status`. Our workflow is running smoke tests after a deployment happens. Since we can't use the caching action it takes a few minutes to download and install the necessary dependencies every time it's run.

Being able to cache this like we do with our `push` and `pull_request` events would save a ton of time.

joshmgross:
Since this issue has all the context, I've changed it to focus on us opening the cache up for more events. Thanks for the feedback!

I'm tracking the confusing error message here: https://github.com/actions/cache/issues/64

nikolaeu:
`repository_dispatch` would be really nice to have covered as well :)

chemitaxis:
Hi @joshmgross yes please, open to other events likes `on: ["deployment"]`

We are using this workaround in this moment, but caching is not working...

https://github.community/t5/GitHub-Actions/GitHub-Actions-Manual-Trigger-Approvals/td-p/31504

Thanks!!

[...]

PR #147, opened 2020-01-10
feat: add support event type `schedule`

bluelovers:
https://github.com/actions/cache/issues/63


joshmgross:
@bluelovers This event validation on the cache client mirrors what's done on the server, so removing it here will just result in an error when creating and reading the cache. Work is needed on the backend server to support the `schedule` event

bluelovers:
allright ... where can request this?

joshmgross:
#63 is the tracking issue

PR #148, opened 2020-01-10
Add Haskell - Cabal example

kutyel:
This has been stolen from @robrix and @patrickt's configurations 😄

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -3,6 +3,7 @@
 - [C# - NuGet](#c---nuget)
 - [Elixir - Mix](#elixir---mix)
 - [Go - Modules](#go---modules)
+- [Haskell - Cabal](#haskell---cabal)
 - [Java - Gradle](#java---gradle)
 - [Java - Maven](#java---maven)
 - [Node - npm](#node---npm)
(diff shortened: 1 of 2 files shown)


examples.md
```diff
@@ -62,6 +63,28 @@ steps:
       ${{ runner.os }}-go-
 ```
 
+## Haskell - Cabal
+
+We cache the elements of the Cabal store separately, as the entirety of `~/.cabal` can grow very large for projects with many dependencies.
```
joshmgross: How large can this get?

kutyel: Probably several GBs in some cases... 🤷‍♂

examples.md
```diff
@@ -62,6 +63,28 @@ steps:
...
+
+```yaml
+- uses: actions/cache@v1
+  name: Cache ~/.cabal/packages
```
hvr: Caching this folder is a very bad idea as it'll eat up half a GiB of cache space for very little benefit; this is why I intentionally did not include this in #38 and also why `haskell-ci` avoids caching that folder for Travis CI

examples.md
```diff
@@ -62,6 +63,28 @@ steps:
...
+    key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-packages
+- uses: actions/cache@v1
+  name: Cache ~/.cabal/store
+  with:
+    path: ~/.cabal/store
+    key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store
+- uses: actions/cache@v1
+  name: Cache dist-newstyle
+  with:
+    path: dist-newstyle
```
hvr: Caching this folder is also of modest benefit; it mostly helps for large projects whose local artifacts (i.e. not the build-deps) are costly to build, otherwise you risk accumulating garbage and wasting the limited cache folder space (also you might risk running into subtle corruption issues when updating `cabal` as the `dist-newstyle` folder isn't properly versioned); this is why I intentionally avoided this folder being included in #38 as well as in general for the `dist-newstyle` folder.

robrix:
cc @tclem ☝️

examples.md
tclem: This is super helpful information and I think definitely work a comment in the example docs as to the tradeoffs.

We could also add cabal version to the cache key for to help with the issue of updating `cabal` for projects where caching `dist-newstyle` is appropriate.

PR #151, opened 2020-01-14
Add renv examples

uribo:
The [renv](https://rstudio.github.io/renv/index.html) is a dependency management toolkit for R.
Sample repository is here https://github.com/uribo/renv_ghaction/actions


kevinushey:
@uribo, thanks for putting this together! I might recommend simplifying this -- one can set the path to the `renv` root directory for cached artifacts with, for example:

```
RENV_PATHS_ROOT = ~/.renv
```

and this would also ensure a stable cache path across different operating systems. Do you think setting that rather than having different OS-specific steps would be better?

uribo:
@kevinushey Thank you for your comment. 
It's great to have commonality across OS.

I think it's a good idea to use a common path (`~/.renv`).

PR #159, opened 2020-01-22
Add Swift Package Manager (SPM) example

123FLO321:
Adds example configuration for [Swift Package Manager (SPM)](https://swift.org/package-manager/) based projects.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -385,4 +385,3 @@ When dependencies are installed later in the workflow, we must specify the same 
     restore-keys: |
       ${{ runner.os }}-spm-
 ```
-
--- a/examples.md
+++ b/examples.md
@@ -16,6 +16,8 @@
 - [Scala - SBT](#scala---sbt)
 - [Swift, Objective-C - Carthage](#swift-objective-c---carthage)
 - [Swift, Objective-C - CocoaPods](#swift-objective-c---cocoapods)
+- [Swift - Swift Package Manager](#swift---swift-package-manager)
+
 
 ## C# - NuGet
 Using [NuGet lock files](https://docs.microsoft.com/nuget/consume-packages/package-references-in-project-files#locking-dependencies):
@@ -373,3 +375,15 @@ When dependencies are installed later in the workflow, we must specify the same 
     restore-keys: |
       ${{ runner.os }}-pods-
 ```
+
+## Swift - Swift Package Manager
+
+```yaml
+- uses: actions/cache@v1
+  with:
+    path: .build
+    key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
+    restore-keys: |
+      ${{ runner.os }}-spm-
+```
+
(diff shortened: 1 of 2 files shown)


examples.md
```diff
@@ -373,3 +374,14 @@ When dependencies are installed later in the workflow, we must specify the same
     restore-keys: |
       ${{ runner.os }}-pods-
 ```
+
+## Swift - Swift Package Manager
+
+```yaml
+- uses: actions/cache@v1
+  with:
+    path: .build
```
joshmgross: Is saving the entire `.build` directory recommended?

123FLO321: Swift places all build files directly in that folder.
That's also how IBMs swift-buildpack does caching (https://github.com/IBM-Swift/swift-buildpack#caching-of-the-build-directory)

abinrm:
@123FLO321  Can you please help me with this condition to check if there is a cache is present or not. My `CI.yml` looks like below 

```
name: CI

on: [push]

jobs:
  build:

    runs-on: macos-latest

    steps:
    - name: Checkout
      uses: actions/checkout@v1
    
    - name: Cache
      uses: actions/cache@v1.1.2
      with:
        path: .build
        key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
        restore-keys: |
          ${{ runner.os }}-spm-
      
    - name: Build app
      if: steps.build.outputs.cache-hit != 'true'
      run: xcodebuild -project 'PROJECTNAME.xcodeproj' -scheme 'SCHEME' -destination 'generic/platform=tvOS' -configuration Release build CODE_SIGNING_ALLOWED=NO | xcpretty
```
Every time I get the following while running this GitHub action for cache
`Cache not found for input keys: macOS-spm-d5b57c01375b10bad0ebb66287afdf2cd3efa116fa0dacf95341e3e895592293, macOS-spm-.`

examples.md
JosephDuffy: @123FLO321 Have you found this caching works for built files or artefacts? https://bugs.swift.org/browse/SR-11760 is being used to track an issue related to how SwiftPM tracks files for the build cache. There was a recent PR for this that was not merged: https://github.com/apple/swift-package-manager/pull/2774.

In my testing (see https://github.com/JosephDuffy/actions-playground/runs/954106324?check_suite_focus=true vs https://github.com/JosephDuffy/actions-playground/runs/954129569?check_suite_focus=true) I see the checkouts cache being used, but the build cache is not being used at all. This is true on stable Xcode (11.5) as was as the 12 beta.

123FLO321: @JosephDuffy the Problem there is that no Package.resolved file exists when trying to load the cache. Try adding a `swift package resolve` step before loading cache.

JosephDuffy: @123FLO321 The checkout step has completed and there's a Package.resolved file in that branch (https://github.com/JosephDuffy/actions-playground/compare/swiftpm-cache#diff-d29466d627f33a3d61709bad13c5eadd). It's a cache hit (see the second link) but SwiftPM still performs a rebuild.

The links are a little long/unclear so I'll update the comment to be clearer.

Issue #160, opened 2020-01-24
Store on Push Only Option (only restore with pull requests)


n1hility:
In order to conserve storage  under the  super small 2GB quote, it would be useful to model a workflow where caches are only stored for the target branch vs every single PR against it. In this way a workflow can store on push, which happens after every merge, and then you avoid pull requests (less useful due to their isolation) from competing for storage with the branch's cache, and the eviction algorithm picking the less efficient target.

I propose adding a parameter to skip pull request events.

dhadka:
While there are certainly usage patterns that would cause problems with the current 2 GB quote, there are some design features that should help:

1. We don't save the cache if there's a cache hit on the exact key.  So ideally the cache (key) gets into the target or master branch, which lets all PRs use that cache and avoid unnecessary duplication.

2. Eviction runs once a day and will evict the least recently used caches first.  We're not evicting as soon as a repo exceeds the 2 GB limit.  Thus, ideally the caches created by pushes are used more frequently and eviction will target the less frequently used PR caches.

dhadka:
@n1hility Also good news - we are increasing the storage limit to 5 GBs.  The limit will be increased on the backend by EOD tomorrow and we can also increase the [individual file limit](https://github.com/actions/cache/blob/e43776276fc1bf0f5f1b462661f341691905b2df/src/save.ts#L59) afterwards.

Issue #94, opened 2019-11-14
Too much configuration required


fregante:
This is the [suggested configuration](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action) to cache `npm`:

```yml
    - name: Cache node modules
      uses: actions/cache@v1
      with:
        path: node_modules
        key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
        restore-keys: |
          ${{ runner.OS }}-build-${{ env.cache-name }}-
          ${{ runner.OS }}-build-
          ${{ runner.OS }}-
```

This is the suggested configuration to cache `npm` on Travis:

```yml

```

Not a mistake. It's empty. It's [the default.](https://docs.travis-ci.com/user/caching/#npm-cache)

Can better defaults be provided? GitHub Actions workflows are 5x to 10x longer than Travis’ for the most common configurations.

chrispat:
I am curious.  Other than pure aesthetics why is shorter better?  And how would you contrast this with all of the request to have a separate save and restore actions more like CircleCI?

fregante:
You can make it both powerful and have sensible defaults. 

1. Could it detect and cache npm  if it was called without arguments?
2. Could it cache npm/yarn with `with: {npm: true}`
3. Could GitHub also wrap this action into another dependencies-specific action (which would include config for most languages)?

I’m sure many will pop up in userland soon enough if you don’t, but your docs will still suggest that wordy configuration if there isn’t an official one.

hugovk:
+1

Compare the 1,000 characters needed for Python compared with 10 for Travis CI: https://github.com/actions/cache/issues/16#issuecomment-553435250.

mrmckeb:
I think this also relates to #16. Less config is always desirable - and if it's better than CircleCI, Travis, GitLab - that's awesome.

krcroft:
> @chrispat: I am curious. Other than pure aesthetics why is shorter better?

I mean this with the most respect, but statements of such nonsense would be career-ending among experienced and wise development teams.  The best code is none at all followed next by the least amount of readily maintainable code.

In this case, we're talking about unnecessary boiler-plate: is it possible to reduce it further? I'm not sure; maybe for the given functionality offered, the amount of required parameters needed in the boiler-plate is already at a minimum.

MikailBag:
I think the only way to reduce boilerplate os to add some presets. E.g.
```yaml
[...]

Issue #55, opened 2019-11-05
Can this module be used from other JavaScript actions?


bahmutov:
I am trying to write a JavaScript action that runs `npm ci` and caches `~/.npm` and `~/.cache/Cypress` folders. I can use this action from YML file like this

```yml
# https://github.com/actions/cache
- name: Cache node modules
  uses: actions/cache@v1
  with:
    path: ~/.npm
    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
    restore-keys: |
      ${{ runner.os }}-node-

- name: Cache Cypress binary
  uses: actions/cache@v1
  with:
    path: ~/.cache/Cypress
    key: cypress-${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
    restore-keys: |
      cypress-${{ runner.os }}-node-

- name: install dependencies
  env:
    # make sure every Cypress install prints minimal information
    CI: 1
  run: npm ci
```

But I would like to have our own Cypress action that does the caching and NPM install for the user, something like

```yaml
- name: Install NPM and Cypress
  uses: @cypress/github-action
```

This means from Cypress GH action JS code I need to call `restore` and `save` files in this action. I have tried but without success, for example

```js
// after executing `npm ci` successfully
const homeDirectory = os.homedir()
const npmCacheDirectory = path.join(homeDirectory, '.npm')

core.saveState('path', npmCacheDirectory)
core.saveState('key', 'abc123')
core.saveState('restore-keys', 'abc123')

console.log('loading cache dist restore')
require('cache/dist/restore')
```
which gives me 
```
done installing NPM modules
loading cache dist restore
##[error]Input required and not supplied: path
##[error]Node run failed with exit code 1
```

I think this particular error is due to a different `core` singleton between "regular" npm module and `ncc`- bundled `actions/cache` one. But in general, do you have by any chance a plan to document how to use this module from JavaScript? I could write TypeScript and bundle it as a top level action, which should bundle `actions/cache` code I think.

byCedric:
I don't think states are accessible from other actions, [as you can see here](https://github.com/actions/toolkit/blob/master/packages/core/src/core.ts#L182). Also, it's kind of hard to use this repository's code as npm dependency due to the ncc bundled code, like you mentioned. However, I'm also interested if we can implement the caching mechanism in 3rd party actions.

[...]

PR #164, opened 2020-01-29
Add a restore-only option to support pulling without saving a cache entry

n1hility:
Implements #160 

This feature allows a workflow to minimize storage consumption of the cache by allowing jobs to selectively store. As an example, pull requests, which typically do not have shared use of caches they create, since they are isolated, can be excluded from saving using the following expression: 

` restore-only: ${{ github.event_name == 'pull_request' }}
`

### DIFF
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -401,3 +401,57 @@ test("restore with cache found for restore key", async () => {
     );
     expect(failedMock).toHaveBeenCalledTimes(0);
 });
+
+test("restore skipped with save-only", async () => {
+    const key = "node-test";
+    const restoreKey = "node-";
+    testUtils.setInputs({
+        path: "node_modules",
+        key,
+        restoreKeys: [restoreKey],
+        saveOnly: "true"
+    });
+
+    const infoMock = jest.spyOn(core, "info");
+    const failedMock = jest.spyOn(core, "setFailed");
+    const stateMock = jest.spyOn(core, "saveState");
+
+    const cacheEntry: ArtifactCacheEntry = {
+        cacheKey: restoreKey,
+        scope: "refs/heads/master",
+        archiveLocation: "www.actionscache.test/download"
+    };
+    const getCacheMock = jest.spyOn(cacheHttpClient, "getCacheEntry");
+    getCacheMock.mockImplementation(() => {
+        return Promise.resolve(cacheEntry);
+    });
+    const tempPath = "/foo/bar";
+
+    const createTempDirectoryMock = jest.spyOn(
+        actionUtils,
+        "createTempDirectory"
+    );
+    createTempDirectoryMock.mockImplementation(() => {
+        return Promise.resolve(tempPath);
+    });
+
+    const setCacheStateMock = jest.spyOn(actionUtils, "setCacheState");
+    const downloadCacheMock = jest.spyOn(cacheHttpClient, "downloadCache");
+    const setCacheHitOutputMock = jest.spyOn(actionUtils, "setCacheHitOutput");
+
+    await run();
+
+    expect(stateMock).toHaveBeenCalledWith("CACHE_KEY", key);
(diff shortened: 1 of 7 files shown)


n1hility:
The macOS failure looks environmental. I don't see it when running directly against my repo.

dhadka:
@joshmgross Any idea what's going on with the `npm ci` error on Mac?

```
npm ERR! code E401
npm ERR! Unable to authenticate, need: Basic realm="GitHub Package Registry"
```

joshmgross:
Resolved the unrelated `npm ci` issues on Mac (it was an issue with a new rollout of that image)

I'll take a look at this PR next week, if we add a restore-only option then we may want a `save-only` option as well just for parity.

n1hility:
> Resolved the unrelated `npm ci` issues on Mac (it was an issue with a new rollout of that image)
> 
> I'll take a look at this PR next week, if we add a restore-only option then we may want a `save-only` option as well just for parity.

@joshmgross thanks! I added a separate second commit for save-only in case you decide you want it.

n1hility:
FWIW we have been using this change successfully in our environment

joshmgross:
Thanks, @n1hility 

Long term I think we want to add separate `restore-cache` and `save-cache` actions to get full control over this sort of thing. Ideally, that will go with having a separate NPM package (see #55) to share the same core code.

I'll see if I can figure out a timeline for all of that. Adding additional configuration to the main cache action makes it more complicated for new users of the action (see #94) and I think most use cases should be covered by the existing eviction method of deleting the least used cache.

n1hility:
Thanks for the update @joshmgross. For verbosity reasons, I think if you added 2 additional actions, you would find yourself wanting to mix "cache" (save and restore) with "restore-cache" instead of restore-cache + save cache right next to each other, mixed with restore-cache. So ideally all 3 operations could be mixed at will.

I think the more explicit name approach as separate actions improves readability but it does potentially increase verbosity since you can't reuse expressions  within the action itself. To use an example, sharing a workflow between PRs and branches (the common case) requires that I have two code paths, each duplicating the action for  each path, instead of something like restore-only: ${{ blah}}. 

That said I probably favor the readability over some added redundancy. Care would also need to be taken to ensure the versions expressed in the workflow  between the 3 operations are compatible. (e.g. mixing v3 of save with v2 or restore  could lead to issues if the packaging format changed). That's a minor and easily corrected issue though.

boredland:
this sounds very reasonable, is there a reason NOT to merge this PR?

n1hility:
@joshmgross how do you want me to proceed on this? It looks like there is a package now?

joshmgross:
> @joshmgross how do you want me to proceed on this? It looks like there is a package now?

👋  Hey @n1hility, we do have a package now! To fully support this scenario, we'd need to separate actions so that we don't need to rely on the `post-if` in this action for saving the cache. 

I don't have a timeline for when we'd create those separate actions, as there's an additional support burden from maintaining multiple repositories of the same action.

For now, I'm fine with having a `skip-save` option that would only run the `restore` part of this action. We can ignore any `save-only` option since that wouldn't work well with the `post-if`.

jeffalder:
@n1hility This fixes our use-case. Where does this PR stand? Is there any way I can help move this forward?

Fixes #334

sakulstra:
Is there any way to help moving this forward?

vsvipul:
@n1hility  @sakulstra  @jeffalder  @boredland Now that the allowed usage per repo has been bumped up to 10 GB, is this something that still needs to be triaged and to be considered? 
If yes, please let me know about your use cases and how can it be useful for users so that I can take it forward. If not, we can go ahead and close it and consider it in the future if need arises. 

cc/ @joshmgross

vsvipul:
Closing for now. Feel free to reopen if you feel this has a valid use-case now, along with more details.

sakulstra:
@vsvipul at least for me this request was not related to space issues.

It's more a matter of wanting to seed the cache from somewhere, without overwriting it.

PR #166, opened 2020-01-31
proxy support

dakale:
Required less changes than I originally thought...

cc @joshmgross To review

Do you have any preference on how we test via canary / what order we do that in? On one hand id love to run that test before merging but I also dont want to make multiple canary changes to get the correct revision of this action and then back to the version tag. It almost seems like we should have an extra test phase that acquires the action from `@master` so we can keep up to date as changes come in but before a release is cut

### DIFF
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -4,7 +4,8 @@ import { BearerCredentialHandler } from "@actions/http-client/auth";
 import { HttpClient, HttpCodes, ITypedResponse } from "@actions/http-client";
 import {
     IHttpClientResponse,
-    IRequestOptions
+    IRequestOptions,
+    IHeaders
 } from "@actions/http-client/interfaces";
 import {
     ArtifactCacheEntry,
@@ -88,9 +89,7 @@ export async function getCacheEntry(
         return null;
     }
     if (!isSuccessStatusCode(response.statusCode)) {
-        throw new Error(
-            `Cache service responded with ${response.statusCode}`
-        );
+        throw new Error(`Cache service responded with ${response.statusCode}`);
     }
 
     const cacheResult = response.result;
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -1,16 +1,17 @@
 import * as core from "@actions/core";
 import * as fs from "fs";
 import { BearerCredentialHandler } from "@actions/http-client/auth";
-import { HttpClient, HttpCodes, ITypedResponse } from "@actions/http-client";
+import { HttpClient, HttpCodes } from "@actions/http-client";
 import {
     IHttpClientResponse,
     IRequestOptions,
-    IHeaders
+    ITypedResponse
 } from "@actions/http-client/interfaces";
 import {
     ArtifactCacheEntry,
     CommitCacheRequest,
-    ReserveCacheRequest
+    ReserveCacheRequest,
+    ReserveCacheResponse
 } from "./contracts";
 import * as utils from "./utils/actionUtils";
 
@@ -135,7 +136,7 @@ export async function reserveCache(key: string): Promise<number> {
     const additionalHeaders = {
         "Content-Type": "application/json"
     };
-    const response = await httpClient.postJson<any>(
+    const response = await httpClient.postJson<ReserveCacheResponse>(
         getCacheApiUrl("caches"),
         reserveCacheRequest,
         additionalHeaders
@@ -266,12 +267,12 @@ async function commitCache(
     httpClient: HttpClient,
     cacheId: number,
     filesize: number
-): Promise<ITypedResponse<any>> {
+): Promise<ITypedResponse<null>> {
     const commitCacheRequest: CommitCacheRequest = { size: filesize };
     const additionalHeaders = {
         "Content-Type": "application/json"
     };
-    return await httpClient.postJson<any>(
+    return await httpClient.postJson<null>(
         getCacheApiUrl(`caches/${cacheId.toString()}`),
         commitCacheRequest,
         additionalHeaders
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -1,13 +1,11 @@
 import * as core from "@actions/core";
 import * as fs from "fs";
-import { BearerCredentialHandler } from "typed-rest-client/Handlers";
-import { HttpClient, HttpCodes } from "typed-rest-client/HttpClient";
-import { IHttpClientResponse } from "typed-rest-client/Interfaces";
+import { BearerCredentialHandler } from "@actions/http-client/auth";
+import { HttpClient, HttpCodes } from "@actions/http-client";
 import {
-    IRequestOptions,
-    RestClient,
-    IRestResponse
-} from "typed-rest-client/RestClient";
+    IHttpClientResponse,
+    IRequestOptions
+} from "@actions/http-client/interfaces";
 import {
     ArtifactCacheEntry,
     CommitCacheRequest,
@@ -16,11 +14,17 @@ import {
 } from "./contracts";
 import * as utils from "./utils/actionUtils";
 
-function isSuccessStatusCode(statusCode: number): boolean {
+function isSuccessStatusCode(statusCode?: number): boolean {
+    if (!statusCode) {
+        return false;
+    }
     return statusCode >= 200 && statusCode < 300;
 }
 
-function isRetryableStatusCode(statusCode: number): boolean {
+function isRetryableStatusCode(statusCode?: number): boolean {
+    if (!statusCode) {
+        return false;
+    }
     const retryableStatusCodes = [
         HttpCodes.BadGateway,
         HttpCodes.ServiceUnavailable,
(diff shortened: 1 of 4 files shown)


joshmgross:
>Do you have any preference on how we test via canary / what order we do that in?

Not particularly. I typically branch off my working branch and push a branch that has the `ncc` release js files and then run that branch in a canary, ex: https://github.com/joshmgross/cache-canary/blob/master/.github/workflows/npm-cache.yml

I've wanted to do some sort of build of `master` canary, but it gets a little complicated (you'd need to install dependencies, build the action, then wipe dependencies so that you can verify that you can pull dependencies from the cache)

bryanmacfarlane:
For testing, see other actions like checkout, setup-ruby etc. which don't ref master, they ref `./` so it even works in your branch and PRs.

src/cacheHttpClient.ts
```diff
@@ -52,39 +56,44 @@ function createAcceptHeader(type: string, apiVersion: string): string {
...
-    const restClient = createRestClient();
+    const httpClient = createHttpClient();
     const resource = `cache?keys=${encodeURIComponent(keys.join(","))}`;
 
-    const response = await restClient.get<ArtifactCacheEntry>(
-        resource,
-        getRequestOptions()
-    );
-    if (response.statusCode === 204) {
+    const response = await httpClient.get(getCacheApiUrl(resource));
```
bryanmacfarlane: I want to add httpclient.getJson<T>().  I will do that now.

src/cacheHttpClient.ts
```diff
@@ -156,38 +165,37 @@ async function uploadChunk(
...
-    requestOptions.additionalHeaders = {
+    const additionalHeaders = {
         "Content-Type": "application/octet-stream",
         "Content-Range": getContentRange(start, end)
     };
 
-    const uploadChunkRequest = async (): Promise<IRestResponse<void>> => {
-        return await restClient.uploadStream<void>(
+    const uploadChunkRequest = async (): Promise<IHttpClientResponse> => {
+        return await httpClient.sendStream(
```
bryanmacfarlane: nice

src/cacheHttpClient.ts
```diff
@@ -253,39 +261,37 @@ async function uploadFile(
...
     filesize: number
-): Promise<IRestResponse<void>> {
-    const requestOptions = getRequestOptions();
+): Promise<IHttpClientResponse> {
     const commitCacheRequest: CommitCacheRequest = { size: filesize };
-    return await restClient.create(
-        `caches/${cacheId.toString()}`,
-        commitCacheRequest,
-        requestOptions
+    return await httpClient.post(
```
bryanmacfarlane: I want to add httpClient.postJSON<T>() method.  will do that now.

bryanmacfarlane:
I renamed the PR to clarify the goal is to support proxy.  Just want to make sure there's no other changes outside of the swapping http clients to make that happen.

src/cacheHttpClient.ts
bryanmacfarlane: e.g. I want to make sure this code happens which would have with the rest client.  https://github.com/microsoft/typed-rest-client/blob/master/lib/RestClient.ts#L198

bryanmacfarlane:
See https://github.com/actions/http-client/pull/9
if OK, I can publish then you can use here.

bryanmacfarlane:
published @actions/http-client@1.0.4

src/cacheHttpClient.ts
```diff
@@ -1,26 +1,29 @@
 import * as core from "@actions/core";
 import * as fs from "fs";
-import { BearerCredentialHandler } from "typed-rest-client/Handlers";
-import { HttpClient, HttpCodes } from "typed-rest-client/HttpClient";
-import { IHttpClientResponse } from "typed-rest-client/Interfaces";
+import { BearerCredentialHandler } from "@actions/http-client/auth";
+import { HttpClient, HttpCodes, ITypedResponse } from "@actions/http-client";
```
dakale: @bryanmacfarlane A few comments around http-client 1.0.4 1

1: We have other interfaces exported in `@actions/http-client/interfaces` but the new `ITypedResponse` comes directly from `@actions/http-client`

src/cacheHttpClient.ts
```diff
@@ -118,17 +128,19 @@ export async function downloadCache(
...
 
     const reserveCacheRequest: ReserveCacheRequest = {
         key
     };
-    const response = await restClient.create<ReserveCacheResponse>(
-        "caches",
+    const additionalHeaders = {
+        "Content-Type": "application/json"
+    };
+    const response = await httpClient.postJson<any>(
```
dakale: @bryanmacfarlane re http-client

2: Having postJson<T> take T and return T means we cant have a separate request/response body Type, I left this common the the PR as well, and a suggestion of how to update it, but heres the example of the use-case

dakale: 3: postJson should maybe automatically set `Content-Type": "application/json`. I am having to set it here manually, I think the typed-rest-client did that automatically, because this port was failing until I sent these headers

src/cacheHttpClient.ts
```diff
@@ -52,37 +56,43 @@ function createAcceptHeader(type: string, apiVersion: string): string {
...
     keys: string[]
 ): Promise<ArtifactCacheEntry | null> {
-    const restClient = createRestClient();
+    const httpClient = createHttpClient();
     const resource = `cache?keys=${encodeURIComponent(keys.join(","))}`;
 
-    const response = await restClient.get<ArtifactCacheEntry>(
-        resource,
-        getRequestOptions()
+    const response = await httpClient.getJson<ArtifactCacheEntry>(
```
dakale: 4: getJson<T> is actually very useful for this type of operation :)

src/cacheHttpClient.ts
```diff
@@ -253,33 +264,35 @@ async function uploadFile(
...
-): Promise<IRestResponse<void>> {
-    const requestOptions = getRequestOptions();
+): Promise<ITypedResponse<any>> {
     const commitCacheRequest: CommitCacheRequest = { size: filesize };
-    return await restClient.create(
-        `caches/${cacheId.toString()}`,
+    const additionalHeaders = {
+        "Content-Type": "application/json"
+    };
+    return await httpClient.postJson<any>(
```
dakale: @bryanmacfarlane 5/5 (last comment for now): This is the other place where postJson<T> is used, but could be more useful if it only *returned* T instead of taking T as input

dakale:
A few comments that about the http-client update that we can either address now or later, but I added a couple extra changes (mostly just sending content-type headers) and all tests and canary workflows seem to be fine now

src/cacheHttpClient.ts
bryanmacfarlane: Yeah, I did that intentionally.  The other response interface is off the http client directly.  https://github.com/actions/http-client/blob/master/index.ts#L53

bryanmacfarlane: oops - no, you're right.  that other one was a class not an interface

bryanmacfarlane:
I think we should address now.  thinking about compat ...

src/cacheHttpClient.ts
```diff
@@ -118,17 +128,19 @@ export async function downloadCache(
...
-    const restClient = createRestClient();
+    const httpClient = createHttpClient();
 
     const reserveCacheRequest: ReserveCacheRequest = {
         key
     };
-    const response = await restClient.create<ReserveCacheResponse>(
-        "caches",
+    const additionalHeaders = {
+        "Content-Type": "application/json"
```
joshmgross: I think we should push a change to the http-client so that `postJson` and `getJson` have these headers automatically.

src/cacheHttpClient.ts
```diff
@@ -253,33 +264,35 @@ async function uploadFile(
...
-): Promise<IRestResponse<void>> {
-    const requestOptions = getRequestOptions();
+): Promise<ITypedResponse<null>> {
     const commitCacheRequest: CommitCacheRequest = { size: filesize };
-    return await restClient.create(
-        `caches/${cacheId.toString()}`,
+    const additionalHeaders = {
+        "Content-Type": "application/json"
+    };
+    return await httpClient.postJson<null>(
```
joshmgross: Should this be `void` instead of `null`?

dakale: I wasnt totally sure. the `ITypedResponse.result` is type `T | null` so using `void` still had the result type as `void | null` which didnt seem more useful than just null

src/cacheHttpClient.ts
dakale: Id like to, but Im not sure how much work it would be (probably not much). Setting the headers for that is easy but we would also need to have code to merge existing headers, making sure not to overwrite any existing content-type header. Ill at least investigate, it shouldnt be a lot of code + a few tests

bryanmacfarlane: merge headers exists: https://github.com/actions/http-client/blob/master/index.ts#L405

bryanmacfarlane: So, let's make the change.  I agree with @joshmgross - it's the reason we have helper methods like these.  Should be a 5 min change.  I'm available to release

dakale:
Should be finished for now. Open for more review @joshmgross or merge (Im not a contributor on this repo)

I used a fork of your canary to validate: https://github.com/dakale/cache-canary

joshmgross:
@dakale @bryanmacfarlane Do we have a good way to test this with a proxy?

dakale:
We've been using a service container as a proxy, usually squid, and then setting http(s)_proxy to the localhost address of that container. You can verify traffic is going through either by blocking all other traffic in various ways, or checking the proxy's logs. Im working on finishing setting that and contributing it to the canary build

PR #168, opened 2020-02-01
Increase cache limit to 5 GBs

dhadka:
Increases the individual file upload limit to 5 GBs to match the new server-side limit.

### DIFF
--- a/src/save.ts
+++ b/src/save.ts
@@ -56,7 +56,7 @@ async function run(): Promise<void> {
 
         await createTar(archivePath, cachePath);
 
-        const fileSizeLimit = 5 * 1024 * 1024 * 1024; // 2GB per repo limit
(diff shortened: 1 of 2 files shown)


src/save.ts
```diff
@@ -56,14 +56,14 @@ async function run(): Promise<void> {
 
         await createTar(archivePath, cachePath);
 
-        const fileSizeLimit = 2 * 1024 * 1024 * 1024; // 2GB per repo limit
+        const fileSizeLimit = 5 * 1024 * 1024 * 1024; // 2GB per repo limit
```
joshmgross:
```suggestion
        const fileSizeLimit = 5 * 1024 * 1024 * 1024; // 5GB per repo limit
```

bryanmacfarlane:
Why doesn't the server send the max size from a registry entry down in the job message?  That would allows us to change without rolling out changes to the action and would also allow you to set per org if there was an exception

PR #170, opened 2020-02-02
[Ruby] Fix the cache gems example

JuanitoFatas:
and remove trailing spaces.

From my experiments, the following to cache gems (Ruby/Bundler) works

```
    - name: Cache gems
      uses: actions/cache@v1
      with:
        path: vendor/bundle
        key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
        restore-keys: |
          ${{ runner.os }}-gem-
```

while the following copied from [the Ruby example](https://github.com/actions/cache/blob/f9c9166ecbcf2d4d7d9d99905a26ae6623cb282c/examples.md#ruby---bundler) does not work 😅    

```
    - name: Cache gems
      uses: actions/cache@v1
      with:
        path: vendor/bundle
        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
        restore-keys: |
          ${{ runner.os }}-gems-
```

The difference is

```diff
- -gems-         <-- does not work
+ -gem-          <-- working
```

Also curious about why? 🤔

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -140,7 +140,7 @@ For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` o
 
 ### Using multiple systems and `npm config`
 
-```yaml  
+```yaml
 - name: Get npm cache directory
   id: npm-cache
   run: |
@@ -171,7 +171,7 @@ The yarn cache directory will depend on your operating system and version of `ya
 
 ## PHP - Composer
 
-```yaml  
+```yaml
 - name: Get Composer Cache Directory
   id: composer-cache
   run: |
@@ -305,9 +305,9 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
 - uses: actions/cache@v1
   with:
     path: vendor/bundle
-    key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+    key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
     restore-keys: |
-      ${{ runner.os }}-gems-
+      ${{ runner.os }}-gem-
 ```
 When dependencies are installed later in the workflow, we must specify the same path for the bundler.
 


liamnichols:
@JuanitoFatas: When you say that it doesn't work, what sort of error do you get? 

I tried both, but I get the following in the post cleanup job:

> /usr/bin/tar -cz -f /Users/runner/runners/2.164.0/work/_temp/1f21752b-9b58-4912-b451-1d9e1147440f/cache.tgz -C /Users/runner/runners/2.164.0/work/project-name/project-name/vendor/bundle .
> tar: could not chdir to '/Users/runner/runners/2.164.0/work/project-name/project-name/vendor/bundle'

Not sure if it relates to your fix/issues though

liamnichols:
Oh, actually this turns out that I was just trying to cache a non-existent directory, any value works here gem, or gems.

The first issue I had, was `vendor/bundle` vs `vendor/bundler` and then the second issue was that it turns out bundler takes the path relative to the `Gemfile`, and that was actually nested... would be good maybe if we could better expose errors like this maybe (“file/directory not found”)

joshmgross:
The key can be whatever you want it to be. @JuanitoFatas please file an issue if you're hitting errors.

>would be good maybe if we could better expose errors like this maybe (“file/directory not found”)
@liamnichols Yep, good point. Right now we just take the input directory and attempt to tar it. It would be good to check that the directory exists first. Could you file an issue?

JuanitoFatas:
It said cache not found if I used `-gems-`, but `-gem-` worked.
There is no error.

スクリーンショット 2020-02-16 13 12 19
スクリーンショット 2020-02-16 13 13 00

joshmgross:
@JuanitoFatas That's an expected outcome. The first time using the cache action there will be no existing cache to restore, so it will be created at the end of your run.

JuanitoFatas:
> @JuanitoFatas That's an expected outcome. The first time using the cache action there will be no existing cache to restore, so it will be created at the end of your run.

I have tried few times with `-gems` but it cannot find the directory to restore, then I switched to `-gem` and it worked.

liamnichols:
> I have tried few times with -gems but it cannot find the directory to restore, then I switched to -gem and it worked.

I found that if the initial store to the cache failed with only a warning (i.e my warning from tar) and not an error, then it would have resulted somehow in nothing being cached but it thinking that something was cached. 

Then, when I resolved the `tar` issue, the cache wasn't being updated since the Gemfile.lock sha never changed so the non-existent cache remained cached.

I had to change `-gems-` also to something else (anything else) since there is no way to clean the cache (#2) from this invalid state other than to wait for it to be purged automatically (after 7 days iirc).

I'll write an issue to describe this also

liamnichols:
Tickets raised (#191 and #192), thanks!

PR #179, opened 2020-02-06
Add test-proxy workflow job, check in dist/, and test local action

dakale:
Check in dist/ per @bryanmacfarlane  & @joshmgross and test the local action instead of a versioned one, also testing with an http_proxy

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -35,7 +35,7 @@ jobs:
       run: |
         echo "::set-output name=dir::$(npm config get cache)"
 
-    - uses: actions/cache@v1
+    - uses: ./
       with:
         path: ${{ steps.npm-cache.outputs.dir }}
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -52,3 +52,45 @@ jobs:
 
     - name: Build & Test
       run: npm run test
+
+  test-proxy:
+    name: Test on ubuntu-latest
+    runs-on: ubuntu-latest
+    services:
+      squid:
+        image: dakale/squid
+        ports: ['3128:3128']
+        options: '--health-cmd "exit 0" --health-interval 3s'
+    env:
+      http_proxy: http://localhost:3128
+      https_proxy: http://localhost:3128
+    steps:
+    - uses: actions/checkout@v1
+
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+
+    - name: Get npm cache directory
+      id: npm-cache
+      run: |
+        echo "::set-output name=dir::$(npm config get cache)"
+
+    - uses: ./
+      with:
+        path: ${{ steps.npm-cache.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install
+      run: npm ci
+
+    - name: Prettier Format Check
+      run: npm run format-check
+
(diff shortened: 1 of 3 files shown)


.github/workflows/workflow.yml
```diff
@@ -52,3 +52,45 @@ jobs:
 
     - name: Build & Test
       run: npm run test
+
+  test-proxy:
```
bryanmacfarlane: the test job should also use the local action.  that gives you confidence in a PR when the checks pass that your code changes.  I think you also want 3 jobs.  build-test (it builds and units pass), test (e2e test using ./ local actions) and test-proxy (same as previous but with a proxy).  See the setup-node workflow which did a good job at this.

bryanmacfarlane:
good that dist is checked in.  you need it to validate changes in a PR.  Note that on a PR validation the resultant merge into master is what is being tested.

bryanmacfarlane:
Adding @ericsciple for review.  I want this and other setup actions to follow the same patterns if at all possible.

.github/workflows/workflow.yml
dakale: I wanted to keep the scope down to just adding proxy support rather than also rewriting how the action is tested. I dont have much of an opinion yet on what an effective test for this action looks like so I just ported the existing one to also run in a proxied environment

bryanmacfarlane: Ok for the extra job.  But making the existing job do a dot slash so it's consistent to what you're adding is needed.  That at least makes the existing and your new one consistent.

dakale: I agree since this PR is already checking in the dist/ which was the blocker for not running the local action (I think). That change was made

Ultimately, it may make sense to split it into 3, first which tests the action code (npm install & test), then the other two test the action itself, which doesnt necessarily need to have anything to do with npm. The existing cache canary i think just writes a text file through the cache. I wonder if it makes more sense to more that type of canary to be local to the action repo?

.github/workflows/workflow.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       run: |
         echo "::set-output name=dir::$(npm config get cache)"
 
-    - uses: actions/cache@v1
+    - uses: ./
```
bryanmacfarlane: awesome.  PRs will now test what's changing with the resultant merge into master :)

.github/workflows/workflow.yml
```diff
@@ -52,3 +52,45 @@ jobs:
...
+      with:
+        path: ${{ steps.npm-cache.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install
+      run: npm ci
+
+    - name: Prettier Format Check
```
bryanmacfarlane: these next three steps shouldn't be here - the job above is about making sure build, unit, lint is covered.  This job is about validating the cache action works through a proxy and the steps above do that.  Notice setup-node pattern for this.

dakale: Im not really sure the npm ci step after the cache restore is super useful either. All that would really be doing is making sure the build can still run after attempting to restore the npm cache from the cache service, which it may or may not, I dont really know the mechanism there. At least with installing and rerunning the tests you get a bit of insight into whether or not the cache action somehow corrupted the actual contents of the packages its restoring. 

I removed them anyways because I think a longer term solution wont use them but I think some future work would need to be planned to anyways to meet that outcome

ericsciple:
since David is on call this week, i opened a PR here https://github.com/actions/cache/pull/185 to validate expected E2E behavior and also through a proxy

joshmgross:
Merged https://github.com/actions/cache/pull/185 which built off this PR. Thanks all!

PR #181, opened 2020-02-09
Run tar with sudo [fixes #133]

### LINKED ISSUE #133: Tracking: Permission Errors when restoring and creating cache
### Issue
Files not created by the local user may fail when restored from the cache as tar preserves file permissions. Permission errors can also occur when the current user does not have permissions to read files when creating the cache. This is likely just a Linux and/or macOS issue (all the reports are Linux)

### Reports
- https://github.com/actions/cache/issues/11
- https://github.com/actions/cache/issues/117
- https://github.com/actions/cache/issues/130
- https://github.com/actions/cache/issues/131

### Possible solutions
1. Run `tar` with `sudo`
   - What are the downsides of this approach?

jwt27:
This works for me. I don't see any downsides to using `sudo` here.

Try it out with `uses: jwt27/cache@dist`

### DIFF
--- a/__tests__/tar.test.ts
+++ b/__tests__/tar.test.ts
@@ -25,14 +25,15 @@ test("extract tar", async () => {
     const tarPath = IS_WINDOWS
         ? `${process.env["windir"]}\\System32\\tar.exe`
         : "tar";
+    let tarParams = ["-xz", "-f", archivePath, "-C", targetDirectory];
+    let tarExec = `${tarPath}`;
+    if (!IS_WINDOWS) {
+        tarExec = "sudo";
+        tarParams = [`${tarPath}`, ...tarParams];
+    }
+
     expect(execMock).toHaveBeenCalledTimes(1);
-    expect(execMock).toHaveBeenCalledWith(`"${tarPath}"`, [
-        "-xz",
-        "-f",
-        archivePath,
-        "-C",
-        targetDirectory
-    ]);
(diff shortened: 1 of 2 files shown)


liamnichols:
Not sure if this is related to the original issue, but I tried your fork out since I was already hitting issues and I got the following:

```
Post job cleanup.
sudo /usr/bin/tar -cz -f /Users/runner/runners/2.164.0/work/_temp/727f0d62-646e-46f2-a73a-dea2fd1ceebe/cache.tgz -C /Users/runner/runners/2.164.0/work/global-ios/global-ios/vendor/bundle .
tar: could not chdir to '/Users/runner/runners/2.164.0/work/global-ios/global-ios/vendor/bundle'
[warning]Tar failed with error: The process 'sudo' failed with exit code 1
```

I think I was getting the `tar: could not chdir to ...` error before though so maybe this isn't related to using `sudo`? The strange thing is that I also have a `vendor/mint` dir that I'm able to cache without an issue

joshmgross:
👋 Hey @jwt27, thanks for your contribution!

Since we can't guarantee that a self-hosted runner is using a passwordless `sudo`, this would break the action for those users. The rust-specific issues in #133 have now been resolved, if you're using a different ecosystem then please file a new issue with your use-case.

evandrocoan:
We could use an option to this action called `use_sudo` and if this option is provided on the `checks.yml` using this action, then it will use sudo. This way we do not break the action for `self-hosted` runners not using passwordless `sudo`.

PR #199, opened 2020-02-26
Adding examples for OCaml/esy

giltho:


joshmgross:
Could you add a link to the README as well? https://github.com/actions/cache#implementation-examples

examples.md
```diff
@@ -170,6 +171,33 @@ The yarn cache directory will depend on your operating system and version of `ya
       ${{ runner.os }}-yarn-
 ```
 
+## OCaml - esy
+Esy allows you to export built dependencies and import pre-built dependencies.
+```yaml
+    - name: Restore Cache
+      id: restore-cache
+      uses: actions/cache@v1.1.2
```
```suggestion
      uses: actions/cache@v1
```
The `@v1` tag is recommended as it tracks the latest v1 version specified

giltho: Good to know, fixed it :)

giltho:
I'm also replacing "OCaml" with "OCaml/Reason" actually, for Reason newcomers who don't know that it is actually the same thing

examples.md
```diff
@@ -170,6 +180,33 @@ The yarn cache directory will depend on your operating system and version of `ya
...
+        path: _export
+        key:  ${{ runner.os }}-esy-${{ hashFiles('esy.lock/index.json') }}
+        restore-keys: |
+          ${{ runner.os }}-esy-
+    - name: Esy install
+      run: 'esy install'
+    - name: Import Cache
+      run: |
+        esy import-dependencies _export
+        rm -rf _export
```
joshmgross: What's the benefit of deleting `_export`?

examples.md
```diff
@@ -170,6 +180,33 @@ The yarn cache directory will depend on your operating system and version of `ya
...
+```yaml
+    - name: Restore Cache
+      id: restore-cache
+      uses: actions/cache@v1
+      with:
+        path: _export
+        key:  ${{ runner.os }}-esy-${{ hashFiles('esy.lock/index.json') }}
+        restore-keys: |
+          ${{ runner.os }}-esy-
+    - name: Esy install
```
joshmgross: Not familiar with this ecosystem, why do you import the cache *after* installing?

examples.md
giltho: Two reasons:
- When you're reexporting, if a dependency has changed, if you didn't erase the folder before, you'll end up with both archives. This is not an "issue" because it won't create a failure, but you will have unwanted elements that can be quite heavy in your cache.
- It cleans your folder, if ever your project creates another `_export` folder it might clash (although the chances someone does this are low

That being said, it is not mandatory and can be removed. It could also be put right before "export-depencencies" which therefore will be executed just in case of cache miss. But this `rm -rf _export` is not really expensive in terms of computation.

examples.md
giltho: `esy install` will mostly resolve the dependency graph and prepare the terrain, this usually takes about ~20s. It's the `esy build` phase that is expensive and that will take between 30seconds and up to an hour or even more on some projects. (For example, building Z3 takes ~20mn, building Coq takes ~30mn). From a clean start, it will also build OCaml from scratch (~10mn).

The `import-dependencies` command looks into the `_export` folders and gets the built packages that *it needs*. In order to know what it needs, it needs to resolve the dependencies first and that is done by `esy install`, so it has to be done before importing the pre-built dependencies.

giltho:
Oh wait, my markdown editor added a whole lot of elements to the table of content in the examples.md, hadn't seen that, let me fix it before merging 😅

Issue #184, opened 2020-02-11
Use zstd instead of gzip


imbsky:

imbsky:
We need to wait for this issue to be closed. https://github.com/actions/virtual-environments/issues/89

imbsky:
#6 has some useful information from @krcroft and @hugovk.

These benchmarks are really useful to consider this: https://github.com/actions/cache/issues/6#issuecomment-551012192 https://github.com/actions/cache/issues/6#issuecomment-551101574

imbsky:
https://github.com/actions/virtual-environments/pull/414

imbsky:
PR has been merged upstream.

imbsky:
@joshmgross Now this is not blocked by anything. Do you have the bandwidth to do this?

joshmgross:
Thanks @imbsky for following up and getting https://github.com/actions/virtual-environments/pull/414 merged.

We'll need to wait for that to get added to the `macOS` image and then for the latest image updates to roll out with `zstd` installed. Typically that's a monthly update, but I'll follow up with the team to verify.

Once https://github.com/actions/virtual-environments/issues/89 is closed I'll take a look at implementing this.
Hopefully should be a small fix, but need to handle back-compat and any self-hosted runners that may not have `zstd` installed.

imbsky:
Yes, indeed, it's a little troublesome to handle self-hosted runners.

imbsky:
I wrote code that ignores backward compatibility. I can do the work for you, or you can refer to this and add changes such as backwards compatibility.
https://github.com/imbsky/cache/commit/e8341fc6fe7b2fa38c1b517c2bc57c5fd987b831

imbsky:
zstd has also been added to macOS. It seems to be deployed in a week or two.

imbsky:
Hmm... My code doesn't seem to work in some environments. When using zstd from tar, the behavior is unstable, so you may want to break it up into separate operations.

joshmgross:
@imbsky Some environments will use BSD tar and some will use GNU tar, so make sure the arguments you specify ones that are recognized by both.

For archive creation, you'd want to replace the `-xz` with `-x` and add the `--zstd` option.
I think for archive extraction you can just remove the `z` from `-cz`, as both GNU and BSD tar say they will automatically decompress based on the file.

See https://www.freebsd.org/cgi/man.cgi?tar(1) and https://www.gnu.org/software/tar/manual/tar.html#SEC135 for more info.

[...]

Issue #139, opened 2020-01-06
Cache restore slow compared to Semaphore CI


davetapley:
[Migrated from here](https://github.community/t5/GitHub-Actions/actions-cache-v1-is-slow-to-restore/m-p/41877#M4782).
 
[Semaphore CI have a cache](https://docs.semaphoreci.com/guided-tour/caching-dependencies/) similar to `actions/cache@v1` but theirs only takes **six seconds** to restore our  760MB `node_modules` Semaphore CI, compared to **twenty eight seconds** on GitHub Actions.

This was on Semaphore's [lowest machine type](https://docs.semaphoreci.com/ci-cd-environment/machine-types/#linux-machine-types) which is comparable to the specifications of the [GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources).

This behavior was observed on multiple builds across several days.

---

I don't have a repro because this was observed on a private repo, but it's easy enough to replicate. I'll create one if it's deemed necessary?

dhadka:
Hi @dukedave, are your caches actually being created?  There is currently a 400 MB limit on the cache size (we will be increasing this limit very soon).  I suspect the restore action is failing as a result and you're having to fetch the dependencies from their source.  We also released a perf improvement in mid December that should also help cut download times.

dhadka:
The cache limit increase is now released.  Please try again and let us know if you're still seeing slow restore times.

imbsky:
I suspect this is a problem with the compression algorithm used.

smeijer:
This is also my experience. 35 sec for ~ 161 mb.

![image](https://user-images.githubusercontent.com/1196524/75584811-ee5b0180-5a70-11ea-9adc-5bcb7926fd7f.png)

imbsky:
We're already working on making it faster, and in my environment it's about three times faster. The next major version will definitely improve.

imbsky:
See this PR: https://github.com/actions/cache/pull/202

imbsky:
Note that this PR will take some time to merge.

joshmgross:
👋  Cache Version 2 has multiple performance improvements, including using `zstd`for compression instead of `gzip`. See https://github.com/actions/cache#whats-new for more details.

I don't think there's anything that's specifically actionable here, if you have any other performance suggestions please open an issue.

mbbernstein:
Hi,

I've the same issue here:
[...]

PR #202, opened 2020-02-28
Use zstd instead of gzip

imbsky:
Closes: #184 #139 
Useful benchmark and explanation: https://github.com/actions/cache/issues/6#issuecomment-551193422

I would like to thank the following people for their help in making this change (in no particular order): @joshmgross @kcgen @hugovk @miketimofeev

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -16,9 +16,9 @@ jobs:
   # Build and unit test
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-      fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout
@@ -27,7 +27,7 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: '12.x'
-    - name: Determine npm cache directory
+    - name: Get npm cache directory
       id: npm-cache
       run: |
         echo "::set-output name=dir::$(npm config get cache)"
@@ -85,7 +85,7 @@ jobs:
   test-proxy-save:
     runs-on: ubuntu-latest
     container:
-      image: ubuntu:latest
+      image: imbsky/zstd:latest
       options: --dns 127.0.0.1
     services:
       squid-proxy:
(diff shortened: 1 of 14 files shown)


imbsky:
I use my minimal zstd docker image for testing, but we can copy and manage it under the Actions organization, or I can invite you as collaborators. (If you want.)

imbsky:
For now, this seems to be working for us. Thank you so much to the many people who worked so hard for this change.

imbsky:
Oh, maybe running a zstd installation within Actions VM would have been enough?

imbsky:
@joshmgross @dhadka This is ready for review. Please review this when you have time. and I will leave it to you when to merge this.

imbsky:
At first, I thought that there might be some problems for the users using the master branch, but that seems to be no problem.

boredland:
this cuts our caching speed in half, even though its still enormous (400mb node_modules from a yarn mongorepo. 33-50s to download/extract, 40s to copy the node_modules dirs into the right places)

imbsky:
@boredland By the way, this is a bit off-topic, but caching `node_modules` is a bad idea.

imbsky:
and this is a change for the purpose of speed up cache action using zstd, so it's not a topic to discuss here even if the result is still slow.

joshmgross:
Just merged #212, so there are a bunch of merge conflicts.

Additionally, now that we're including `version` in the cache that we create, I think we should support both `tar` and `zstd` and include that information in the version. Self-hosted runners might not have `zstd` installed.

Let me know if you need any help or want me to take over this PR

imbsky:
@joshmgross Well, I wasn't used to the test with Jest and I couldn't write very clean. I don't think there will be much change, so can you push some commits to this branch?

boredland:
Hi, are there any updates? I really like the speed with this.

joshmgross:
@boredland We're still planning on making this change, just need to ensure that we have a good scenario for self-hosted runners that might not have `zstd` installed

imbsky:
This PR is probably never merged. Because @aiqiaoy took over the rest of this PR work from me. Please check the new PR for additional information.

imbsky:
I will keep this PR as a draft until the new PR is merged, but I will close it when it's merged.

imbsky:
This was merged into the master in another PR!

PR #206, opened 2020-03-06
Add C++ Boost library example

erenon:
While Linux and Windows have Boost preinstalled, on macOS, it is not. As downloading and building boost takes a lot of time, caching the resulting install dir can result in considerable resource saving.
The shown example was tested here: https://github.com/Morgan-Stanley/binlog/actions/runs/50704053

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -1,6 +1,7 @@
 # Examples
 
 - [Examples](#examples)
+  - [C++ - Boost](#c---boost)
   - [C# - NuGet](#c---nuget)
   - [Elixir - Mix](#elixir---mix)
   - [Go - Modules](#go---modules)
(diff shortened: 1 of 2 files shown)


examples.md
```diff
@@ -28,6 +29,40 @@
...
+
+Downloading, building and installing some boost libraries:
+
+```yaml
+- name: Cache Boost
+  id: cache-boost
+  uses: actions/cache@v1
+  with:
+    path: boost_1_72_0/installdir
+    key: ${{ runner.os }}-boost_1_72_0-fs-system-test
```
dhadka: I'm wondering if we can move the list of libraries to a variable:

```
env:
    boost_version: boost_1_27_0
    boost_libs: filesystem,system,test
```

So that when either changes, the `key` is in sync.  Otherwise, if either changes but the key isn't updated, this would break.

One note though is the key has a maximum length of 512 characters, so if the list of libraries grows to large this will fail.  We could hash the list of libraries to workaround the limit.

PR #195, opened 2020-02-21
Small message change

guerrerocarlos:
Remove dot that generates confusion in wether that's part of the key or not

PR #209, opened 2020-03-10
Attempt to delete the archive after extraction

henrymercer:
This PR modifies the `restore` job to attempt to delete the downloaded archive once it has been extracted.  This reduces storage space used on the Actions runner once the cache Action has finished executing.

This change is particularly useful for users using a large proportion of the 5GB of currently available cache space, as keeping both the archives and the unzipped content could use close to 10GB of the available [14GB](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) of storage on GitHub-hosted Actions runners.

### DIFF
--- a/dist/restore/index.js
+++ b/dist/restore/index.js
@@ -1651,6 +1651,7 @@ const io = __importStar(__webpack_require__(1));
 const fs = __importStar(__webpack_require__(747));
 const os = __importStar(__webpack_require__(87));
 const path = __importStar(__webpack_require__(622));
+const util = __importStar(__webpack_require__(669));
 const uuidV4 = __importStar(__webpack_require__(826));
 const constants_1 = __webpack_require__(694);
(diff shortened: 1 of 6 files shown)


dist/restore/index.js
```diff
@@ -2831,18 +2836,29 @@ function run() {
             try {
                 const cacheEntry = yield cacheHttpClient.getCacheEntry(keys);
                 if (!((_a = cacheEntry) === null || _a === void 0 ? void 0 : _a.archiveLocation)) {
-                    core.info(`Cache not found for input keys: ${keys.join(", ")}.`);
+                    core.info(`Cache not found for input keys: ${keys.join(", ")}`);
```
henrymercer: Note that this change was originally made in #195 but the Action wasn't recompiled.

PR #211, opened 2020-03-11
Documenting using cache for Kotlin DSL of gradle

raj-saxena:
Kotlin DSL files end `.kts`. Added comment for Kotlin based projects.


joshmgross:
Fixed with https://github.com/actions/cache/pull/216 which allows for both`.gradle` and `.gradle.kts` files.

Thanks!

Issue #215, opened 2020-03-16
Debug logs for hashFiles function


JeremyLoy:
The `hashFiles` function accepts a glob (which I think is parsed by the actions/toolkit glob package).

Its extremely hard at the moment to determine if your glob (and therefore cache-key) is correct as it doesn't output the names of the hashed files.

I propose the filename inputs to `hashFiles` are logged to provide a simpler, smoother, and more accurate experience using actions/cache

mcotillo:
#212

mcotillo:
> @gudmundur  @``

joshmgross:
Thanks for the suggestion @JeremyLoy!

Since `hashFiles` is supported by the runner, going to transfer that issue to that repo

PR #216, opened 2020-03-16
Include Kotlinscript Gradle files

JeremyLoy:
Tested this with my own repo which uses a mix of `build.gradle` and `build.gradle.kts` files and this glob seems to be working correctly.

As an aside, please checkout #215 as it would make the process of verifying these globs easier!


joshmgross:
Thanks!

PR #226, opened 2020-03-23
Complete Rust documentation for Windows and Linux

evandrocoan:


examples.md
```diff
@@ -376,7 +376,15 @@ When dependencies are installed later in the workflow, we must specify the same
 
 ## Rust - Cargo
 
+The following example is for Linux. For Windows, rust replace `~/.cargo` by `C:\Rust\.cargo`.
+Also, do forget to change `**/Cargo.lock` to `**/Cargo.toml` if your project uses the former.
+See [Cargo.toml vs Cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) for more information.
+Note, on Windows, you should also use forward slash ``**/Cargo.lock`` instead of backslash ``**\Cargo.lock``.
```
joshmgross: I would expect both `**/Cargo.lock` and `**\Cargo.lock` to work, is that not the case?

evandrocoan: I just did a quick test and it seems to be working with backslashes: https://github.com/evandrocoan/TestGithubActions/runs/540711928?check_suite_focus=true#step:3:4

So, I rephrased that from should to can.

dhadka:
I don't think this PR is needed anymore so I'm closing.  But please re-open or comment if I'm mistaken!

https://github.com/actions/cache/issues/133#issuecomment-599102035 is merged so the permissions issue should be fixed.

I also checked the path on Windows and found that `~/.cargo` is a symlink to `C:\Rust\.cargo`:

```
l----           5/19/2021  8:21 PM                .cargo -> C:\Rust\.cargo
```

PR #237, opened 2020-03-31
Fix cache version in lerna example

desbo:
As far as I know, there is no `cache@v2`.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -189,7 +189,7 @@ The yarn cache directory will depend on your operating system and version of `ya
 
 ```yaml
 - name: restore lerna
-  uses: actions/cache@v2
+  uses: actions/cache@v1
    with:
      path: |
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -189,7 +189,7 @@ The yarn cache directory will depend on your operating system and version of `ya
 
 ```yaml
 - name: restore lerna
-  uses: actions/cache@v2
+  uses: actions/cache@v1
```
```suggestion
  uses: actions/cache@master
```

We're actually close to releasing a v2, but it isn't available yet. This example uses multiple paths in a single `path` which isn't going to be released in v1. For now I think it's safe to use `@master` until `v2` is released.

joshmgross:
Thanks for your contributions! Fixed with https://github.com/actions/cache/pull/242

PR #242, opened 2020-04-02
Fix Lerna Example

joshmgross:
This updates the Lerna example to reference the `master` branch since `v2` is still a work-in-progress and the multi-path feature (https://github.com/actions/cache/pull/212) is not yet released.

Also updating the `hashFiles` to reference every `yarn.lock` file
>Lerna uses yarn.lock at root but also in every package.
>https://github.com/actions/cache/pull/212#issuecomment-607772401

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -11,8 +11,8 @@
     - [macOS and Ubuntu](#macos-and-ubuntu)
     - [Windows](#windows)
     - [Using multiple systems and `npm config`](#using-multiple-systems-and-npm-config)
-  - [Node - Yarn](#node---yarn)
   - [Node - Lerna](#node---lerna)
+  - [Node - Yarn](#node---yarn)
   - [OCaml/Reason - esy](#ocamlreason---esy)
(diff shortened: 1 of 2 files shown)


examples.md
```diff
@@ -168,6 +168,19 @@ For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` o
       ${{ runner.os }}-node-
 ```
 
+## Node - Lerna
+
+>Note this example uses the new multi-paths feature and is only available at `master`
+```yaml
+- name: restore lerna
```
gcourtemanche, line 175: Spacing is off by 1 space for `with`

```
- name: restore lerna
  uses: actions/cache@master
  with:
    path: |
      node_modules
      */*/node_modules
    key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```

joshmgross: Thanks! Fixed

YraIssa:
[Ty](Ty u) #351

Issue #249, opened 2020-04-08
Extracting multipath tar needs to respect relative paths


dhadka:
Using for example:

```
    - uses: actions/cache@master
      with:
        path: ~/cache-folder
        key: ${{ runner.os }}-cache
```

`~` points to `/home/runner`, so the cache is creating a tar of `/home/runner/cache-folder`.

However, we are extracting relative to the `pwd` or `GITHUB_WORKSPACE` (see https://github.com/actions/cache/blob/master/src/tar.ts#L35), which is `/home/runner/work/<repo-name>/<repo-name>`.  The cache contents is extracted to `/home/runner/work/<repo-name>/<repo-name>/cache-folder`.

**Expected:** The cache folder is extracted to `~/cache-folder` or `/home/runner/cache-folder`.

dhadka:
Works after adding the `-P` option when creating the tar.

dhadka:
Here's a test for and repro of the issue - https://github.com/actions/cache/runs/569639238?check_suite_focus=true

dhadka:
@ethanis Can you please check this?  Should createTar be using the -P option? extractTar already uses it.

ethanis:
This seems odd - I'm able to create a tarball with relative paths without using the `-P` option.

Steps for me to repro:

```bash
in Desktop/tmp_restore/repo
  cat ~/Desktop/tmp_cache/repo/manifest.txt
../up.txt
one.txt
two.txt
sub/three.txt

in Desktop/tmp_restore/repo
  tar -cz -f "cache.tgz" -C /Users/ethand/Desktop/tmp_cache/repo --files-from /Users/ethand/Desktop/tmp_cache/repo/manifest.txt

in Desktop/tmp_restore/repo
  tar -tvf cache.tgz
-rw-r--r--  0 ethand staff       3 Apr  8 11:54 ../up.txt
-rw-r--r--  0 ethand staff       2 Apr  8 11:53 one.txt
-rw-r--r--  0 ethand staff       2 Apr  8 11:53 two.txt
-rw-r--r--  0 ethand staff       2 Apr  8 11:53 sub/three.txt
```

PR #250, opened 2020-04-08
Fix caching directories outside of the working directory (relative paths)

dhadka:
Adds tests for including relative paths that are outside the working directory.  For example, it's common to cache `~/.nuget`, so we need to ensure these are restored to the correct location.

Fixes relative paths by adding the `-P` flag to tar creation to prevent `tar` from stripping leading paths.

Example without the `-P` flag:
```/bin/tar: Removing leading `../../../' from member names```

Fixes https://github.com/actions/cache/issues/249

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -47,7 +47,7 @@ jobs:
       run: npm run test
 
   # End to end save and restore
-  test-save:
+  test-save-relative:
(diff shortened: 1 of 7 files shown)


joshmgross:
I turned on debug logging and re-ran the tests, but forgot that re-runs on PRs are broken 😞

Issue #153, opened 2020-01-15
Cache fails when using actions/cache@v1.1.0 on Windows-2016


wangsongc:
Cache fails when using actions/cache@v1.1.0 on Windows-2016, and using actions/cache@v1.0.3 is work
https://github.com/wangsongc/Hellow-World/runs/390647603
```
Post job cleanup.
"C:\Program Files\Git\usr\bin\tar.exe" -cz -f D:\a\_temp\3e1a7faf-0d91-4c9d-9dff-dd5cb2f54509\cache.tgz -C D:\a\Hellow-World\Hellow-World .
tar (child): Cannot connect to D: resolve failed
/usr/bin/tar: Child returned status 128
/usr/bin/tar: Error is not recoverable: exiting now
[warning]Tar failed with error: The process 'C:\Program Files\Git\usr\bin\tar.exe' failed with exit code 2. Ensure BSD tar is installed and on the PATH.
```

joshmgross:
@wangsongc Thank's for reporting this, this was caused https://github.com/actions/cache/pull/126

For a workaround, I'd recommend using v1.0.3.

Note: Windows 2016 will be deprecated soon (see https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#windows-server-2016),

qaisjp:
i get this as well.

**step**

```yaml
      - uses: actions/cache@v1
        id: cache-dxfiles
        with:
          path: utils/DXFiles
          key: dxfiles-0001
```

**full workflow**  
https://github.com/multitheftauto/mtasa-blue/blob/275eae917eb0f39abd69f7e0133ffed02f318335/.github/workflows/build.yaml#L15-L19

**error message in Post actions/cache@v1**  
```
Post job cleanup.
"C:\Program Files\Git\usr\bin\tar.exe" -cz -f d:\a\_temp\8f05a114-911c-4bbb-90e7-367cad3a2601\cache.tgz -C d:\a\mtasa-blue\mtasa-blue\utils\DXFiles .
tar (child): Cannot connect to d: resolve failed
/usr/bin/tar: d\:\\a\\_temp\\8f05a114-911c-4bbb-90e7-367cad3a2601\\cache.tgz: Cannot write: Broken pipe
/usr/bin/tar: Child returned status 128
/usr/bin/tar: Error is not recoverable: exiting now
[...]

PR #252, opened 2020-04-09
Fallback to GNU tar if BSD tar is unavailable

aiqiaoy:
Test added for the fallback scenario. 

Fixes https://github.com/actions/cache/issues/153 and https://github.com/github/c2c-actions-service/issues/854

### DIFF
--- a/src/tar.ts
+++ b/src/tar.ts
@@ -1,3 +1,4 @@
+import * as core from "@actions/core";
 import { exec } from "@actions/exec";
 import * as io from "@actions/io";
 import { existsSync, writeFileSync } from "fs";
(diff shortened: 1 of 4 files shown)


src/tar.ts
```diff
@@ -39,7 +53,14 @@ export async function extractTar(archivePath: string): Promise<void> {
     // Create directory to extract tar into
     const workingDirectory = getWorkingDirectory();
     await io.mkdirP(workingDirectory);
-    const args = ["-xz", "-f", archivePath, "-P", "-C", workingDirectory];
+    const args = [
+        "-xz",
+        "-f",
+        archivePath?.replace(/\\/g, "/"),
```
joshmgross, line 59: `\` is a valid file character on UNIX systems, so we shouldn't replace it with a path separator if not on Windows.

We can either add a check for Windows here, or use `path.sep` to replace:
```
archivePath?.replace(path.sep, "/")
```
See https://nodejs.org/api/path.html#path_path_sep

src/tar.ts
```diff
@@ -58,10 +79,10 @@ export async function createTar(
     const args = [
         "-cz",
         "-f",
-        CacheFilename,
+        CacheFilename?.replace(/\\/g, "/"),
```
joshmgross, line 82: Why are these file names optional?
If they're `undefined` we should throw a warning, but at this point in the code, I would expect both of them to be defined since we've already validated our inputs.

src/tar.ts
aiqiaoy: `replace(path.sep, "/")` only replaces the first instance of the separator. I did some googling and changed to `replace(new RegExp('\\' + path.sep, 'g'), "/")` instead.

src/tar.ts
aiqiaoy: Removed. 

I kept those in tests because otherwise the testes would fail with `workspace is possibly undefined` since we have strict null check enabled.

PR #228, opened 2020-03-23
Bump prettier from 1.19.1 to 2.0.1

imbsky:

PR #254, opened 2020-04-10
Prettier 2.0.4

singingwolfboy:
This upgrades us to the latest version of [Prettier](https://prettier.io/). See [the blog post introducing version 2.0](https://prettier.io/blog/2020/03/21/2.0.0.html).

I also switched to pinning an exact version of Prettier, [as suggested in the install docs](https://prettier.io/docs/en/install.html).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7737,9 +7737,9 @@
       "dev": true
     },
     "prettier": {
-      "version": "1.19.1",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.19.1.tgz",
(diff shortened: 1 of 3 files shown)


joshmgross:
:wave: Hey @singingwolfboy, thanks for making this change! 

Similar to my response in https://github.com/actions/cache/pull/228#issuecomment-624288021, it's generally easier to have a core-contributor make these sort of changes as they can be difficult to review. While this specific change is small, I believe the changes will be larger once the Windows build is fixed, see https://github.com/actions/cache/pull/228/files

Could you file an issue for us to make this upgrade to prettier?

🙇 Thank you again for your support.

Issue #208, opened 2020-03-06
Cache fails for Pull Request re-runs with "No scopes with read permission were found on the request."


jcornaz:
In pull requests we often get:

```
[warning]No scopes with read permission were found on the request.
```

And the build runs without cache.

Please note that I'm talking about *pull request* build triggered when *pushing* a commit (or opening the PR). So this issue is *not* explained by #129.

joshmgross:
Can you provide an example of when this occurs?
Do you see it on the first PR run, or when re-running a PR?

jcornaz:
> Can you provide an example of when this occurs?

Not really sorry. I am not allowed to share that sources. I'll do my best to create a simple reproducer.

> Do you see it on the first PR run, or when re-running a PR?

If I look at the latest occurrence of the problem, it was when pushing a new commit to a PR already built.

jcornaz:
In the hope it helps, here is more information:

*We configured `action/checkout` to build the `HEAD` of the PR branch (not the merge commit).
```yaml
      - uses: actions/checkout@v2
        with:
          ref: ${{ github.event.pull_request.head.sha }}
```

* Here is how we configure the workflow to run on build requests:
```
on:
  pull_request:
    branches:
      - master
    paths-ignore:
      - '**/*.md'
```

* The workflow is composed of multiple jobs, all using the same cache key.
  * One *initial* job. runs alone
  * The next jobs run all in parallel and use `matrix` strategies. That causes ~ 20 concurrent fetches of the same cache.

bxq2011hust:
I have the same problem.
Sometimes a pr runs without cache and GitHub Action output.

```bash
[warning]No scopes with read permission were found on the request.
```

Today I meet the problem after I cancel and rerun. Here is the link
https://github.com/FISCO-BCOS/FISCO-BCOS/pull/1402/checks?check_run_id=493039648

```bash
 Run actions/cache@v1 0s
[warning]No scopes with read permission were found on the request.
Run actions/cache@v1
[warning]No scopes with read permission were found on the request.
  Run actions/cache@v1 1s
[warning]No scopes with read permission were found on the request.
Run actions/cache@v1
[warning]No scopes with read permission were found on the request.
```

joshmgross:
[...]

PR #263, opened 2020-04-17
Allow all events to access cache 

aiqiaoy:
This PR removes the check on the cache actions side and allows all events that are tied to a ref to access cache. It cannot be merged until after the server side change are fully rolled out. 

Fixes: https://github.com/github/c2c-actions-service/issues/369

### DIFF
--- a/src/utils/actionUtils.ts
+++ b/src/utils/actionUtils.ts
@@ -7,13 +7,17 @@ import * as path from "path";
 import * as util from "util";
 import * as uuidV4 from "uuid/v4";
 
+<<<<<<< HEAD
 import {
     CacheFilename,
     CompressionMethod,
-    Events,
     Outputs,
+    RefKey,
     State
 } from "../constants";
+=======
+import { Outputs, RefKey, State } from "../constants";
+>>>>>>> Allow all events to access cache
 import { ArtifactCacheEntry } from "../contracts";
 
 // From https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L23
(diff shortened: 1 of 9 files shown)


src/utils/actionUtils.ts
```diff
@@ -101,16 +101,10 @@ export async function resolvePaths(patterns: string[]): Promise<string[]> {
...
-}
-
-// Currently the cache token is only authorized for push and pull_request events
-// All other events will fail when reading and saving the cache
+// Cache token authorized for all events that are tied to a ref
 // See GitHub Context https://help.github.com/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context
 export function isValidEvent(): boolean {
-    const githubEvent = process.env[Events.Key] || "";
-    return getSupportedEvents().includes(githubEvent);
+    return RefKey in process.env;
```
joshmgross, line 107: Does this validate that it's non-empty?

aiqiaoy: Nope. I'll add it.

aiqiaoy: Added

manuyavuz-pointr:
Hello,

Can I ask if this is related to scoping problem that causes cache to be inaccessible for rerunning workflows from GitHub UI?

aiqiaoy:
@manuyavuz-pointr No this is not related to the rerun scope issue. 

https://github.com/actions/cache/issues/208 is the issue tracking the problem and the awesome @joshmgross wrote an explanation here https://github.com/actions/cache/issues/208#issuecomment-607877125.

signalnerve:
sorry for drive-by comment but just wanted to say i'm super excited for this 🎉

imbsky:
Looks good!

localheinz:
Thanks a ton!

❤️ it!

localheinz:
@aiqiaoy 

Just gave this a try, but it appears that the `deployment` event is still not supported:

Screen Shot 2020-05-16 at 14 17 02

What will it take to make `actions/cache` work with the `deployment` event?

A reference could be passed in the payload, for example.

localheinz:
Perhaps https://github.com/actions/cache/pull/317 will help?

PR #269, opened 2020-04-22
Adds socket timeout and validate file size

dhadka:
Adds socket timeout of 5 seconds and validates file size.

Recent network issues between the Actions runner and cache storage have caused flakiness in download speeds.  While debugging, we observed periods where no traffic was received over the socket.  In some cases, no traffic is received but the socket remains open for many minutes.  Furthermore, if the socket closes before all content is received, the cache action incorrectly proceeds to untar the file, causing an "unexpected end of file" error.

This PR sets a socket timeout of 5 seconds.  As a result, if there is a period were no traffic is received for 5 seconds, the download will abort and will be treated as a cache miss.  Additionally, the downloaded file size is compared against the expected content length, if provided.

### DIFF
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -9,7 +9,7 @@ import {
 import * as crypto from "crypto";
 import * as fs from "fs";
 
-import { Inputs } from "./constants";
+import { Inputs, SocketTimeout } from "./constants";
 import {
     ArtifactCacheEntry,
     CommitCacheRequest,
@@ -144,7 +144,33 @@ export async function downloadCache(
     const stream = fs.createWriteStream(archivePath);
     const httpClient = new HttpClient("actions/cache");
     const downloadResponse = await httpClient.get(archiveLocation);
+
+    // Abort download if no traffic received over the socket.
+    downloadResponse.message.socket.setTimeout(SocketTimeout, () => {
+        downloadResponse.message.destroy();
+        core.debug(
+            `Aborting download, socket timed out after ${SocketTimeout} ms`
+        );
+    });
+
     await pipeResponseToStream(downloadResponse, stream);
+
+    // Validate download size.
+    var contentLengthHeader =
+        downloadResponse.message.headers["content-length"];
+
+    if (contentLengthHeader) {
+        const expectedLength = parseInt(contentLengthHeader);
+        const actualLength = utils.getArchiveFileSize(archivePath);
+
+        if (actualLength != expectedLength) {
+            throw new Error(
+                `Incomplete download. Expected file size: ${expectedLength}, actual file size: ${actualLength}`
+            );
+        }
+    } else {
+        core.debug("Unable to validate download, no Content-Length header");
+    }
 }
 
 // Reserve Cache
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -156,7 +156,7 @@ export async function downloadCache(
     await pipeResponseToStream(downloadResponse, stream);
 
     // Validate download size.
-    var contentLengthHeader =
+    const contentLengthHeader =
         downloadResponse.message.headers["content-length"];
 
     if (contentLengthHeader) {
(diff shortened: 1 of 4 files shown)


src/cacheHttpClient.ts
```diff
@@ -144,7 +144,33 @@ export async function downloadCache(
...
     const httpClient = new HttpClient("actions/cache");
     const downloadResponse = await httpClient.get(archiveLocation);
+
+    // Abort download if no traffic received over the socket.
+    downloadResponse.message.socket.setTimeout(SocketTimeout, () => {
+        downloadResponse.message.destroy();
+        core.debug(
+            `Aborting download, socket timed out after ${SocketTimeout} ms`
+        );
+    });
```
joshmgross, line 154: What happens when this occurs? Is this when we would hit the mismatched file length below and throw the error?

src/cacheHttpClient.ts
```diff
@@ -144,7 +144,33 @@ export async function downloadCache(
...
+        const expectedLength = parseInt(contentLengthHeader);
+        const actualLength = utils.getArchiveFileSize(archivePath);
+
+        if (actualLength != expectedLength) {
+            throw new Error(
+                `Incomplete download. Expected file size: ${expectedLength}, actual file size: ${actualLength}`
+            );
+        }
+    } else {
+        core.debug("Unable to validate download, no Content-Length header");
```
joshmgross, line 172: Do we expect this to occur?

src/constants.ts
```diff
@@ -20,3 +20,5 @@ export enum Events {
 }
 
 export const CacheFilename = "cache.tgz";
+
+export const SocketTimeout = 5000;
```
joshmgross, line 24: How did we decide on this value? Are there any recommendations or guidelines from Azure Blog Storage?

src/cacheHttpClient.ts
dhadka: Yes, this closes the socket and interrupts the download, which is then caught by the file length check.

src/cacheHttpClient.ts
dhadka: No, I wouldn't expect this to occur.

src/constants.ts
dhadka: I don't think there are any guidelines.  This value has worked well in practice in the [cache-canary](https://github.com/dhadka/cache-canary/actions) repo, as I haven't seen it abort any downloads that were still working...it's only aborted downloads where traffic over the socket stopped.

munrocket:
Nice work! Hope that this will be merged.

src/constants.ts
aiqiaoy, line 24: Nit: Add SocketTimeoutInMs?

src/cacheHttpClient.ts
```diff
@@ -144,7 +144,33 @@ export async function downloadCache(
     const stream = fs.createWriteStream(archivePath);
     const httpClient = new HttpClient("actions/cache");
     const downloadResponse = await httpClient.get(archiveLocation);
+
+    // Abort download if no traffic received over the socket.
+    downloadResponse.message.socket.setTimeout(SocketTimeout, () => {
+        downloadResponse.message.destroy();
```
aiqiaoy, line 150: Nodejs newbie here but do we need to manually destroy the socket?

dhadka: Yes, from the [docs](https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback), this only invokes the timeout callback.  `end()` or `destroy()` must be called manually.  During testing, I found `end()` wasn't sufficient to kill the connection.  From the docs it says `Half-closes the socket. i.e., it sends a FIN packet. It is possible the server will still send some data.`, so if the server is unresponsive it might not close the socket.

src/constants.ts
dhadka: Added comment with details.

munrocket:
Ty!

munrocket:
Validation is work #271. So to finally fix this problem we need to wrap `warning: exit code 2` and try to download cache again, as it was done in https://github.com/actions/setup-node/issues/116

PR #270, opened 2020-04-23
Prefer zstd over gzip

### LINKED ISSUE #184: Use zstd instead of gzip

aiqiaoy:
Fixes #184 

A lot of the work was completed by @imbsky in PR https://github.com/actions/cache/pull/202. Additionally, this PR adds:
* `zstd` in the cache version
* A gzip fallback path for self-hosted runners that do not have `zstd` installed
* Jest tests


imbsky:
Awesome @aiqiaoy!! I will review this PR soon 🙂 Thanks for your work!

imbsky:
Perhaps you should pick this commit (https://github.com/actions/cache/pull/202/commits/c7743bc415800f75b178a76307ced0808804414b) change to make the test work correctly.

src/cacheHttpClient.ts
```diff
@@ -82,12 +83,11 @@ function createHttpClient(): HttpClient {
     );
 }
 
-export function getCacheVersion(): string {
+export function getCacheVersion(useZstd?: boolean): string {
```
thomaseizinger, line 86: This function is a ["boolean-trap"](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap). It might be worth considering using an enum here, like:

```ts
enum CompressionTool {
	Zstd,
	GnuTar
}
```

.github/workflows/workflow.yml
```diff
@@ -1,5 +1,8 @@
 name: Tests
 
+env:
+  ACTIONS_STEP_DEBUG: true
```
joshmgross, line 4: I think you'll need to enable this via a Secret. See https://help.github.com/en/actions/configuring-and-managing-workflows/managing-a-workflow-run#enabling-debug-logging

src/cacheHttpClient.ts
imbsky: Well, in my opinion, this is not a big deal. The enum has mistakes, depending on how you look at it, and should not be actively recommended for use. If you want to avoid use boolean, you should use string literal types combine with union types. But it's too overkill for this purpose. So, there is no problem with the current code.

imbsky:
@aiqiaoy This should work.

```typescript
const args = [
  ...(useZstd ? ["--use-compress-program", "zstd -T0 --long=31"] : ["-z"]),
  "-xf",
  archivePath.replace(new RegExp("\\" + path.sep, "g"), "/"),
  "-P",
  "-C",
  workingDirectory.replace(new RegExp("\\" + path.sep, "g"), "/"),
];
```

imbsky:
This isn't a tar problem, it's a strange specification of Node.js.

src/cacheHttpClient.ts
thomaseizinger: Never said that it is a big deal :)

I was more coming from a readability-PoV because the first change I saw in the PR diff was:

```ts
test("getCacheVersion with zstd compression returns version", async () => {
    testUtils.setInput(Inputs.Path, "node_modules");

    const result = getCacheVersion(true);

    expect(result).toEqual(
        "273877e14fd65d270b87a198edbfa2db5a43de567c9a548d2a2505b408befe24"
    );
});
```

and I didn't understand what `getCacheVersion(true)` means, hence the comment about boolean-traps :)

imbsky: Sorry. This kind of expression is really common in Japanese, So I said. (English is still difficult for me...) It certainly seems to have a readability problem, but I think it's basically within the acceptable range.

imbsky: If this needs to be changed, I suggest a major redesign, but it requires too much work and this action probably won't be a large codebase that it needs.

aiqiaoy:
> @aiqiaoy This should work.
> 
> ```ts
> const args = [
>   ...(useZstd ? ["--use-compress-program", "zstd -T0 --long=31"] : ["-z"]),
>   "-xf",
>   archivePath.replace(new RegExp("\\" + path.sep, "g"), "/"),
>   "-P",
>   "-C",
>   workingDirectory.replace(new RegExp("\\" + path.sep, "g"), "/"),
> ];
> ```

@imbsky Thanks! Nodejs newbie here and you definitely saved me a lot of headaches lol

imbsky:
My pleasure. I had the same experience before when implementing this hehe...

imbsky:
The format check seems to have failed.

imbsky:
@aiqiaoy Try adding the following `.gitattributes` file.

```gitattributes
* text=auto eol=lf
```

src/cacheHttpClient.ts
aiqiaoy: PR updated to use enum instead.

imbsky:
Oh, all tests have passed now. I will review this soon. Perhaps you are also interested in doing that. @joshmgross

src/cacheHttpClient.ts
```diff
@@ -82,12 +83,13 @@ function createHttpClient(): HttpClient {
...
     // Add salt to cache version to support breaking changes in cache entry
-    const components = [
-        core.getInput(Inputs.Path, { required: true }),
-        versionSalt
-    ];
+    const components = [core.getInput(Inputs.Path, { required: true })].concat(
+        compressionMethod == CompressionMethod.Zstd
+            ? [compressionMethod, versionSalt]
+            : versionSalt
+    );
```
joshmgross, line 92:
```suggestion
    const components = [core.getInput(Inputs.Path, { required: true }), versionSalt].concat(
        compressionMethod == CompressionMethod.Zstd
            ? [compressionMethod]
            : []
    );
```


This might be more clear if we move `versionSalt` out of the `concat` since it's not dependent on the compression method. (Note this suggested change may need to be `prettier`)

src/tar.ts
```diff
@@ -39,7 +22,7 @@ async function getTarPath(args: string[]): Promise<string> {
 
 async function execTar(args: string[], cwd?: string): Promise<void> {
     try {
-        await exec(`"${await getTarPath(args)}"`, args, { cwd: cwd });
+        await exec(`${await getTarPath(args)}`, args, { cwd: cwd });
```
joshmgross, line 25: Is there a reason you're removing these quotes around the tar path?

On Windows, the tar path might have spaces in it and the quotes are needed make sure it's a valid path to exec

src/tar.ts
```diff
@@ -49,13 +32,18 @@ function getWorkingDirectory(): string {
...
+    compressionMethod: CompressionMethod
+): Promise<void> {
     // Create directory to extract tar into
     const workingDirectory = getWorkingDirectory();
     await io.mkdirP(workingDirectory);
     const args = [
-        "-xz",
-        "-f",
+        ...(compressionMethod == CompressionMethod.Zstd
+            ? ["--use-compress-program", "zstd -d"]
```
joshmgross, line 44: What does the `-d` flag do for zstd?

src/utils/actionUtils.ts
```diff
@@ -116,3 +123,44 @@ export function isValidEvent(): boolean {
 export function unlinkFile(path: fs.PathLike): Promise<void> {
     return util.promisify(fs.unlink)(path);
 }
+
+async function checkVersion(app: string): Promise<string> {
```
joshmgross, line 127: nit: Should this be `getVersion`? We're not doing any checks in this function, those are done in `getCompressionMethod` and `useGnuTar`

src/tar.ts
imbsky: `-d` means decompression. and we should add the long flag. It has the best performance against GitHub Actions on our previous benchmarks and needs this.

imbsky:
All of the following flags are needed for compression,

```
zstd -T0 --long=31
```

and all of the following flags are needed for a decompression.

```
zstd --long=31 -d
```

CC: @kcgen

imbsky:
This explanation by @kcgen is really impressive.
https://github.com/actions/cache/issues/184#issuecomment-591617222

aiqiaoy:
> All of the following flags are needed for compression,
> 
> ```
> zstd -T0 --long=31
> ```
> 
> and all of the following flags are needed for a decompression.
> 
> ```
> zstd --long=31 -d
> ```
> 
> CC: @nickvanw 

The doc says the maximum for `--long` is 30 (1 GiB) on 32-bit platforms and 31 (2 GiB) on 64-bit platforms.

@joshmgross Do you know if we support 32 bits OS for our self-hosted runners?

joshmgross:
>@joshmgross Do you know if we support 32 bits OS for our self-hosted runners?

Yep we do. We at least have an ARM32 Runner that's in Pre-Release

src/tar.ts
aiqiaoy: Forgot about Windows. Reverted.

kcgen:
@aiqiaoy, great catch! There will be little lost in knocking down the window size to the universally-compatible `30` value; it's still a huge improvement (over existing compressors) for those scenarios where people are caching data with long running redundancy.

aiqiaoy:
PR updated. 

@joshmgross @imbsky It would be great if you can take a look again when you have time :)

src/tar.ts
```diff
@@ -49,13 +32,27 @@ function getWorkingDirectory(): string {
...
     await io.mkdirP(workingDirectory);
+    // --d: Decompress.
+    // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
     const args = [
-        "-xz",
-        "-f",
+        ...(compressionMethod == CompressionMethod.Zstd
+            ? [
+                  "--use-compress-program",
+                  isOS64() ? "zstd -d --long=31" : "zstd -d --long=30"
```
joshmgross, line 52: Does this have any impact on compatibility between 32bit and 64bit OS's using the same cache?

I.e. Can a cache created on a 64-bit runner be extracted by a 32-bit runner?

If so, we may want to include the OS arch in the cache version hash

aiqiaoy: You are right it would. 

I think I'm gonna change it to use the universal 30. It wouldn't be the fastest but having too many version of essentially the same cache is also not ideal.

imbsky:
This looks good to me! I really appreciate your work!

src/tar.ts
imbsky: Are you going to make any changes related to long distance matching in this PR?

aiqiaoy: Yes. I'm changing it to 30 in the next commit unless someone strongly oppose (and with good reasoning of course)

imbsky: Okay, I think that's fine.

kcgen: Another option is to drop down to 30 across the board.

This would eliminate code complexity (by dropping the architecture bit-size checks and branches), and would avoid wasting storage caused by cache users exercising mixed-bit-size platforms.

The trade off is giving up the the compression benefits of finding matching across 2GB spans of data versus 1GB spans; which for some highly-redundant data (say, like ) can make a big difference, but for binary / code / etc.. it's not going to be a big loss.



-  forfeit a very marginal amount of compression

aiqiaoy:
Thanks everyone for the review and feedback. I'll check this PR in early Monday just to be on the safer side.

imbsky:
Thanks @aiqiaoy for doing the rest of the work.
Thanks @miketimofeev for helping me make changes to the infrastructure.
and kudos to @kcgen for helping me most with this change.
Thank you all!

shunkakinoki:
It is way faster, thank you!

imbsky:
@joshmgross Is there anything left to do for the v2 or next v1 series release?

imbsky:
If I remember correctly, the multi path stuff changes and this were the only remaining tasks for v2.

joshmgross:
> @joshmgross Is there anything left to do for the v2 or next v1 series release?

We're tracking https://github.com/actions/cache/pull/263 for v2, but that might be back-ported to v1 as well.

imbsky:
Oh, I see. Will the next version probably be released after the PR is merged?

imbsky:
Well, I just want to check, so there is no rush to do anything.

PR #284, opened 2020-04-29
Better error handling during download

dhadka:
Improves error handling by using the same fix described in https://github.com/actions/toolkit/pull/369.

For example, in the case the download stream is interrupted before all bytes are downloaded, this will raise the error `Premature close`.  This should eliminate the need to check the content length header, but it also doesn't hurt to leave the content length check in place.


ericsciple:
the promisified pipeline simplifies things quite a bit :) - fyi docs here: https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_stream_pipeline_streams_callback

PR #285, opened 2020-04-29
Add Python example using 'pip cache dir' to get cache location

hugovk:
:tada: I'm happy to say `pip cache dir` has been released in this week's pip 20.1:

* https://github.com/pypa/pip/issues/7350
* https://github.com/pypa/pip/pull/8095
* https://pip.pypa.io/en/stable/news/

This means we can simplify the cache config for multiple operating systems:

```yaml
- name: Get pip cache dir
  id: pip-cache
  run: |
    python -m pip install -U "pip>=20.1"
    echo "::set-output name=dir::$(pip cache dir)"

- name: pip cache
  uses: actions/cache@v1
  with:
    path: ${{ steps.pip-cache.outputs.dir }}
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

If/when the https://github.com/actions/virtual-environments/ ship with pip 20.1 or later, `python -m pip install -U "pip>=20.1"` could be removed.

This PR adds this to examples.md, and fixes an indentation problem in an existing example.

---

I didn't remove any of the other examples:

* "Multiple OS's in a workflow" could still be useful, if someone just wanted, say, the Windows config

* "Using a script to get cache location": perhaps this should be removed? It uses an internal pip API, although could still be useful if someone had to use pip<20.1 for some reason.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -294,10 +294,10 @@ Replace `~/.cache/pip` with the correct `path` if not using Ubuntu.
 
 > Note: This uses an internal pip API and may not always work
 ```yaml
-- name: Get pip cache
-   id: pip-cache
-   run: |
-     python -c "from pip._internal.locations import USER_CACHE_DIR; print('::set-output name=dir::' + USER_CACHE_DIR)"
+- name: Get pip cache dir
+ id: pip-cache
+ run: |
+   python -c "from pip._internal.locations import USER_CACHE_DIR; print('::set-output name=dir::' + USER_CACHE_DIR)"
 
 - uses: actions/cache@v1
   with:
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -290,14 +291,33 @@ Replace `~/.cache/pip` with the correct `path` if not using Ubuntu.
...
 ```
 
+### Using pip to get cache location
+
+> Note: This requires pip 20.1+
+```yaml
+- name: Get pip cache dir
+  id: pip-cache
+  run: |
+    python -m pip install -U "pip>=20.1"
```
joshmgross, line 301: Is it safe to recommend that users install pip just to get a cache directory? We could just keep the "Note: This requires pip 20.1+"  and leave out the install step.

hugovk: It would be safer to pin to `pip==20.1`, but then that would get out of date, so yeah, let's just leave out the command and let users make a conscious decision how to upgrade.

Updated!

hugovk:
You're welcome, thank you for GitHub Actions!

PR #290, opened 2020-05-02
Recommend --deployment flag for Bundler

sonicdoe:
Bundler’s `--deployment` flag activates [deployment mode](https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE) which is specifically optimized for deployment and continuous integration. As these use cases arguably cover the majority of workflows, I think it should be recommended over manually setting the path using `bundle config`.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -385,7 +385,13 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
     restore-keys: |
       ${{ runner.os }}-gems-
 ```
-When dependencies are installed later in the workflow, we must specify the same path for the bundler.
+When dependencies are installed later in the workflow, we must use the `--deployment` flag to install gems to `vendor/bundle`.
+
+```yaml
+- name: Bundle install
+  run: bundle install --deployment --jobs 4 --retry 3
+```
+If you don’t want to run Bundler in deployment mode, you can instead use `bundle config`.
 
 ```yaml
 - name: Bundle install


sonicdoe:
I’ve just noticed that https://github.com/actions/cache/pull/289 proposes a similar change.

examples.md
```diff
@@ -385,7 +385,13 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
...
       ${{ runner.os }}-gems-
 ```
-When dependencies are installed later in the workflow, we must specify the same path for the bundler.
+When dependencies are installed later in the workflow, we must use the `--deployment` flag to install gems to `vendor/bundle`.
+
+```yaml
+- name: Bundle install
+  run: bundle install --deployment --jobs 4 --retry 3
+```
+If you don’t want to run Bundler in deployment mode, you can instead use `bundle config`.
```
joshmgross, line 394: Could we make it clear that you don't need both of these Bundle install steps?

sonicdoe: What do you think about either completely dropping this sentence or replacing it with the suggestion below?

> Note: Bundler’s [deployment mode](https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE) requires an up-to-date `Gemfile.lock`.

I think having a `Gemfile.lock` is a good practice and the majority of projects do have one. In my opinion, we don’t necessarily need to complicate this example by directly providing an alternative. (As for out-of-date lockfiles, `--deployment` is arguably a helpful safeguard.)

sonicdoe:
As @scottjacobsen already said in https://github.com/actions/cache/pull/289#issuecomment-624309222, it might be good to wait for the outcome of https://github.com/ruby/setup-ruby/pull/53 and then simply adopt that.

aparna-ravindra:
I see there is another PR that closes this.

sonicdoe:
Do you have a link to that other pull request? I only know of https://github.com/actions/cache/pull/289, which was closed a few months ago.

aparna-ravindra:
> Do you have a link to that other pull request? I only know of #289, which was closed a few months ago.

https://github.com/actions/cache/pull/396

PR #291, opened 2020-05-02
Use Prettier for Markdown

sonicdoe:
Configures [Prettier](https://prettier.io) to run on Markdown files (`*.md`). Analogous to the existing use of Prettier for TypeScript, this would help with having a single consistent style of writing Markdown documents.


.prettierrc.json
```diff
@@ -7,5 +7,12 @@
     "trailingComma": "none",
     "bracketSpacing": true,
     "arrowParens": "avoid",
-    "parser": "typescript"
```
sonicdoe, line 10: [From Prettier’s docs](https://prettier.io/docs/en/options.html#parser):

> Prettier automatically infers the parser from the input file path, so you shouldn't have to change this setting.

Therefore, this option can be safely removed.

PR #293, opened 2020-05-04
Allow importing save and restore without side effects

jirfag:
Hi,
I need to import save and restore function from my custom action. If I understand it properly - currently such importing will run functions in background. Therefore I've moved running in `*_main.ts` files.

Also, `dist/*.js` files changed, It wasn't intended. I'll try to find out why.


joshmgross:
We've created an `@actions/cache` npm package for this use-case: https://github.com/actions/toolkit/tree/master/packages/cache

jirfag:
thank you!

Issue #267, opened 2020-04-20
Networking issues impacting cache save/restore


dhadka:
Since we've received a number of issues related to network issues impacting save/restore, I'm creating a "catch all" issue to track the investigation and mitigation.

### Related Issues
https://github.com/actions/cache/issues/265 https://github.com/actions/cache/issues/259 https://github.com/actions/cache/issues/207 https://github.com/actions/cache/issues/200 https://github.com/actions/cache/issues/189

### Diagnostics 
The figure below shows the **max** server latency from our cloud storage provider for one of the storage accounts (the cache service uses 16 total, but all show similar trends).  There's a clear increase in high latency requests the past two weeks or so.

![image](https://user-images.githubusercontent.com/2496211/79810164-64c50300-8340-11ea-9948-0828bb0d974e.png)

I've escalated with our provider and they have started an investigation.  I'll update here as we know more.

pjoe:
Following here (from #265). Would be great if at least network failure was detected probably (during restore) and either the action failing or at least leaving disk empty, instead of a half restored and corrupted state on disk, that for us breaks following steps.

pjoe:
Is there a way to set a timeout for cache restore?

In current situation we are burning build minutes just waiting for cache restore that eventually times out and leaves corrupted cache on disk :(

everops-alex:
We're seeing this as well, restoring ~300mb of ruby gems with unpredictable delays.

I've seen as fast as 9s (seconds), average of 15-20s, and hundreds of highs of 40s, 50s, 3m20s, 5m, using cache@v1.1.2 (and other vanilla cache tags) we have ran ~10,000 cache pulls with t

I've tested switching to zstd compression, and while the compression and unpackaging is faster, we are still seeing slow retrieval swings, as high as 50 second retrieval in the ~400 tests I ran with zstd.

cc @chrispat

dhadka:
@pjoe I've yet to try this, but the docs say you can set a [step timeout](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) that will kill the process if the step exceeds that limit.  Note, however, that *this will cause the step to fail*.  What we really need is a cache-specific timeout that will abort the restore but allow the workflow to continue, treating it like a cache miss.

[...]

PR #306, opened 2020-05-11
Add retries to all API calls

dhadka:
Adds one retry to every API call if a retryable http response code is returned (e.g., 503) or an Error is thrown (e.g., socket-level timeout).

Helps with https://github.com/actions/cache/issues/267

### DIFF
--- a/src/cacheHttpClient.ts
+++ b/src/cacheHttpClient.ts
@@ -30,6 +30,13 @@ function isSuccessStatusCode(statusCode?: number): boolean {
     return statusCode >= 200 && statusCode < 300;
 }
 
+function isServerErrorStatusCode(statusCode?: number): boolean {
+    if (!statusCode) {
+        return true;
+    }
+    return statusCode >= 500;
+}
+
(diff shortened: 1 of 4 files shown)


src/cacheHttpClient.ts
```diff
@@ -109,8 +185,8 @@ export async function getCacheEntry(
         keys.join(",")
     )}&version=${version}`;
 
-    const response = await httpClient.getJson<ArtifactCacheEntry>(
-        getCacheApiUrl(resource)
+    const response = await retryTypedResponse("getCacheEntry", () =>
+        httpClient.getJson<ArtifactCacheEntry>(getCacheApiUrl(resource))
```
joshmgross, line 189: The http-client has built-in retries for read operations that we could use for this call and `downloadCache`. It's on the `IRequestOptions` and is ignored for write operations.

dhadka: I'd have to check again, but I think one issue with http-client is that it only retries on certain http status codes.  If there's a socket-level error, it wouldn't retry.  But as you say, this is certainly something that could be put into http-client.

PR #313, opened 2020-05-15
Switch cache action to use the cache node package

aiqiaoy:
Biggest change is in `setCacheState`. We are now only saving cache key not the entire artifact cache object since cacheKey what we get back from `saveCache`.


src/restore.ts
```diff
@@ -25,89 +23,42 @@ async function run(): Promise<void> {
...
 
-            core.info(
-                `Cache restored from key: ${cacheEntry && cacheEntry.cacheKey}`
-            );
+            core.info(`Cache restored from key: ${cacheKey}`);
         } catch (error) {
-            utils.logWarning(error.message);
-            utils.setCacheHitOutput(false);
+            if (error.name === cache.ValidationError.name) {
+                throw error;
```
joshmgross, line 57: Nice!

package.json
```diff
@@ -25,16 +25,13 @@
   "dependencies": {
     "@actions/core": "^1.2.0",
     "@actions/exec": "^1.0.1",
-    "@actions/glob": "^0.1.0",
-    "@actions/http-client": "^1.0.8",
     "@actions/io": "^1.0.1",
-    "uuid": "^3.3.3"
+    "@actions/cache": "^0.1.0"
```
joshmgross, line 29: Let's make sure we publish a `v1` of `@actions/cache` before we publish this `v2` action. Would be odd if we had a major action version consume a preview package

src/utils/actionUtils.ts
```diff
@@ -1,86 +1,35 @@
...
+        cacheKey.localeCompare(key, undefined, {
             sensitivity: "accent"
         }) === 0
     );
 }
 
-export function setCacheState(state: ArtifactCacheEntry): void {
-    core.saveState(State.CacheResult, JSON.stringify(state));
+export function setCacheState(state: string): void {
+    core.saveState(State.CacheResult, state);
```
joshmgross, line 15: Let's change the name of this `CacheResult` constant since we're only storing the key now in the state

src/utils/actionUtils.ts
```diff
@@ -1,86 +1,35 @@
...
 }
 
-export function setOutputAndState(
-    key: string,
-    cacheResult?: ArtifactCacheEntry
-): void {
-    setCacheHitOutput(isExactKeyMatch(key, cacheResult));
+export function setOutputAndState(key: string, cacheKey?: string): void {
+    setCacheHitOutput(isExactKeyMatch(key, cacheKey));
     // Store the cache result if it exists
```
joshmgross, line 24:
```suggestion
    // Store the cache key if it exists
```

joshmgross:
Let's add a note in the README pointing to the package in the toolkit

imbsky:
Well, will Zstd support be removed?

aiqiaoy:
@imbsky See https://github.com/actions/toolkit/pull/469 for the zstd fixes.

imbsky:
(I just checked the PR 🙂)

PR #317, opened 2020-05-16
Enhancement: Allow usage when GITHUB_REF or ACTIONS_CACHE_REF are defined

localheinz:
This PR

* [x] allows use when `GITHUB_REF` environment variable is not defined, but `ACTIONS_CACHE_REF`

Follows https://github.com/actions/cache/pull/263#issuecomment-629636936.

💁‍♂️ I would like to be able to use `actions/cache` for `deployment` events.

Here's my use case:

- I create a `deployment` in workflows, e.g., when pushing to a branch that should be deployed to a staging environment, or when pushing to a branch that should be deployed to a production environment
- I process `deployment` events in a separate workflow, where an application is deployed to an environment depending on the `payload` of the `deployment` event, and here I would like to be able to use `actions/cache` to speed up builds. For `deployment` events, `GITHUB_REF` is not defined.

❓I'm wondering, the previously available `RefKey` is only used to verify whether the environment variable `GITHUB_REF` is defined, but as far as I could see, the actual value of `GITHUB_REF` is never used. Why is that the case? Why do we care in the first place?

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -4,7 +4,7 @@ import { promises as fs } from "fs";
 import * as os from "os";
 import * as path from "path";
 
-import { Events, Outputs, RefKey, State } from "../src/constants";
+import { Events, Outputs, RefKeys, State } from "../src/constants";
 import { ArtifactCacheEntry } from "../src/contracts";
 import * as actionUtils from "../src/utils/actionUtils";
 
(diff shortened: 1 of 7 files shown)


joshmgross:
👋 Thanks for your support @localheinz, please see my response on your other PR: https://github.com/actions/cache/pull/318#issuecomment-629836750 as to why this won't work and consider filing an issue for your use-case.

localheinz:
Thank you, @joshmgross!

PR #318, opened 2020-05-17
Fix: Remove event validation

localheinz:
This PR

* [x] removes the event validation entirely

💁‍♂️ This is an alternative to #317 - I have tried using this action with the `deployment` event, and as far as I understand, the event validation currently serves no real purpose. The action currently verifies that the `GITHUB_REF` environment variable is set, but never uses its value.

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -4,7 +4,7 @@ import { promises as fs } from "fs";
 import * as os from "os";
 import * as path from "path";
 
-import { Events, Outputs, RefKey, State } from "../src/constants";
+import { Events, Outputs, State } from "../src/constants";
 import { ArtifactCacheEntry } from "../src/contracts";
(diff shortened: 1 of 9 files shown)


joshmgross:
👋 Hey @localheinz, thanks for your contribution and feedback.

>I have tried using this action with the deployment event, and as far as I understand, the event validation currently serves no real purpose.

The event validation mirrors the validation done on the server when creating and restoring caches. Every cache is associated with a `scope`, which is equivalent to the branch it is created on. These scopes are stored on the JWT token used to authenticate with the backing cache service. The reason we do this validation in the cache action is to provide a better error message to users and save time from creating a cache that will not be saved by the server. See https://github.com/actions/cache/pull/68

We'd need to do some more work on the backend cache service to support caching for events without a scope/ref attached to them. Please file an issue in this repository explaining your use-case and workflow.

localheinz:
Thank you, @joshmgross!

PR #325, opened 2020-05-24
Improved and expanded Rust examples

mzabaluev:
- Avoid caching redundant data.
- Demonstrate how to share cached dependency downloads across build jobs in a multi-OS matrix.
- A recipe for testing libraries that do not have `Cargo.lock` checked in.


mzabaluev:
The new examples at work:
https://github.com/mzabaluev/github-actions-playground/tree/cargo-bin-example
https://github.com/mzabaluev/github-actions-playground/tree/cargo-lib-example

mzabaluev:
I have updated the examples to `actions/cache@v2`.

matu3ba:
Examples are intended to be minimal as you can see from other languages. Please make another repo to include them there, because github will likely not maintain big examples.

Issue #323, opened 2020-05-22
Document and Release V2 of actions/cache


joshmgross:
With the recent improvements made to the cache action, we'll be releasing a V2 version of this action.

### What's New

* Added support for multiple paths (https://github.com/actions/cache/issues/16), glob patterns, and single file caches (https://github.com/actions/cache/issues/33)
* Increased performance and improved cache sizes using `zstd` for compression (https://github.com/actions/cache/issues/184). (Off for windows-latest hosted runner)
* Allowed caching for all events with a ref (https://github.com/actions/cache/issues/63)
    * See [Events that trigger workflows](https://help.github.com/en/actions/reference/events-that-trigger-workflows) for info on which events do not have a `GITHUB_REF`
* Created the [`@actions/cache`](https://github.com/actions/toolkit/tree/master/packages/cache) npm package to allow other actions to utilize caching (https://github.com/actions/cache/issues/55)
* Added a best-effort step to delete the archive after extraction to reduces storage space (https://github.com/actions/cache/pull/209). 

### To Do

- [x] Update README to include a "What's New" section and info about upgrading to `v2`
- [x] Update [examples](https://github.com/actions/cache/blob/master/examples.md) to reference `v2` and use multiple paths if possible
- [x] Update the help docs ([Caching dependencies to speed up workflows](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) to reference `v2` (Tracked via an internal issue
- [x] Update the action.yml description for `path` to mention glob, multipath, and single file support. This should link to [`@actions/glob`](https://github.com/actions/toolkit/tree/master/packages/glob) for glob pattern documentation
- [x] Update the toolkit cache package README to include a link to the cache action as example usage. 
- [x] Change log https://github.com/github/releases/issues/835

aiqiaoy:
[v2.0.0](https://github.com/actions/cache/releases/tag/v2.0.0) is now released!

Issue #301, opened 2020-05-08
zstd compression hanging on Windows with large files


dhadka:
The caching action always hangs when using `actions/cache@master` on Windows in my test runs using a random 150 MB file.

To dig in more, I created a separate repo that runs the compression commands.  On a 1 MB file, it works fine:

```
2020-05-08T19:01:50.9215932Z bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.5.f-ipp
2020-05-08T19:01:51.7976258Z Create random file
2020-05-08T19:01:51.8560894Z 1+0 records in
2020-05-08T19:01:51.8563149Z 1+0 records out
2020-05-08T19:01:51.8568671Z 1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.011073 s, 94.7 MB/s
2020-05-08T19:01:51.8578782Z Running zstd standalone
2020-05-08T19:01:52.3317293Z 
2020-05-08T19:01:52.3325676Z 
2020-05-08T19:01:52.3331145Z     Directory: D:\a\test-zstd\test-zstd
2020-05-08T19:01:52.3376822Z 
2020-05-08T19:01:52.3392726Z Mode                 LastWriteTime         Length Name
2020-05-08T19:01:52.3395053Z ----                 -------------         ------ ----
2020-05-08T19:01:52.3412014Z -a---            5/8/2020  7:01 PM        1048613 file.zstd
2020-05-08T19:01:52.3412805Z Running tar standalone
2020-05-08T19:01:52.3608654Z -a---            5/8/2020  7:01 PM        1050112 file.tar
2020-05-08T19:01:52.3622919Z Running tar + zstd in two steps
2020-05-08T19:01:52.4062645Z -a---            5/8/2020  7:01 PM        1050624 file_combined.tzst
2020-05-08T19:01:52.4064236Z Running tar + zstd using --use-compress-program
2020-05-08T19:01:52.4668289Z -a---            5/8/2020  7:01 PM        1049135 file.tzst
2020-05-08T19:01:52.4670493Z Done
```

But on a 150 MB file, it hangs until it I cancel the action:

```
2020-05-08T19:03:52.6081814Z bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.5.f-ipp
2020-05-08T19:03:53.4979676Z Create random file
2020-05-08T19:03:53.6897822Z 150+0 records in
2020-05-08T19:03:53.6909594Z 150+0 records out
[...]

PR #329, opened 2020-05-26
Update readme and examples to use v2

aiqiaoy:
https://github.com/actions/cache/issues/323

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,57 +8,27 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
-## What's New in V2
+## What's New
 
-* Added support for caching multiple paths,
+* Added support for multiple paths, [glob patterns](https://github.com/actions/toolkit/tree/master/packages/glob), and single file caches. 
 
 ```yaml
-- name: Cache multiple relative paths
-  uses: actions/cache@v2
-  with:
-    path: |
-      node_modules
-      dist
-    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
-```
-```yaml
-- name: Cache multiple absolute paths
+- name: Cache multiple paths
   uses: actions/cache@v2
   with:
     path: |
       ~/cache
-      /path/to/dependencies
-    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
-```
-
-[glob patterns](https://github.com/actions/toolkit/tree/master/packages/glob), 
-
-```yaml
-- name: Cache using glob patterns
-  uses: actions/cache@v2
-  with:
-    path: |
-      **/*.ts
+      !~/cache/exclude
       **/node_modules
     key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
 ```
 
-and single file caches. 
-
-```yaml
-- name: Cache single file
-  uses: actions/cache@v2
-  with:
-    path: cache.tar
-    key: ${{ runner.os }}
-```
-
-* Increased perfomance and improved cache sizes using `zstd` compression
-> Note this feature is off for Windows runner that are using `bsdtar` (e.g., windows-latest hosted runner) due to a bug in ziping large random files with `bsdtar`
-* Allowed caching for all events with a ref
-> See [events that trigger workflow](https://help.github.com/en/actions/reference/events-that-trigger-workflows) for info on which events do not have a `GITHUB_REF`
+* Increased performance and improved cache sizes using `zstd` compression for Linux and macOS runners
+* Allowed caching for all events with a ref. See [events that trigger workflow](https://help.github.com/en/actions/reference/events-that-trigger-workflows) for info on which events do not have a `GITHUB_REF`
 * Released the [`@actions/cache`](https://github.com/actions/toolkit/tree/master/packages/cache) npm package to allow other actions to utilize caching
 * Added a best-effort cleanup step to delete the archive after extraction to reduce storage space
+
+Refer [here](https://github.com/actions/cache/blob/v1/README.md) for previous versions
 
 ## Usage
 
@@ -67,8 +37,7 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 
 ### Inputs
 
-* `path` - Directories to store and save the cache. Supports pattern matching, multipath and single file cache
-> See [`@actions/glob`](https://github.com/actions/toolkit/tree/master/packages/glob) for supported patterns. 
+* `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/master/packages/glob) for supported patterns. 
 * `key` - An explicit key for restoring and saving the cache
 * `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key
 
--- a/README.md
+++ b/README.md
@@ -8,6 +8,58 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
+## What's New in V2
+
+* Added support for caching multiple paths,
+
+```yaml
+- name: Cache multiple relative paths
+  uses: actions/cache@v2
+  with:
+    path: |
+      node_modules
+      dist
+    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
...
+```yaml
+- name: Cache single file
+  uses: actions/cache@v2
+  with:
+    path: cache.tar
+    key: ${{ runner.os }}
+```
+
+* Increased perfomance and improved cache sizes using `zstd` compression
+> Note this feature is off for Windows runner that are using `bsdtar` (e.g., windows-latest hosted runner) due to a bug in ziping large random files with `bsdtar`
```
hashtagchris, line 57:
```suggestion
> Note this feature is off for Windows runner that are using `bsdtar` (e.g., windows-latest hosted runner) due to a bug in zipping large random files with `bsdtar`
```

Should we link to the bsd tar issue?

README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
+## What's New in V2
+
+* Added support for caching multiple paths,
+
+```yaml
```
joshmgross, line 15: For the sake of keeping this "What's New" section short, let's combine these into one example for multiple paths

README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
...
+- name: Cache single file
+  uses: actions/cache@v2
+  with:
+    path: cache.tar
+    key: ${{ runner.os }}
+```
+
+* Increased perfomance and improved cache sizes using `zstd` compression
```
joshmgross, line 56:
```suggestion
* Increased performance and improved cache sizes using `zstd` compression
```

README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
+## What's New in V2
```
konradpabjan, line 11: This can simply be `What's new` (without the v2)

Similar to the artifact actions, checkout, and setup-python. Checkout started this pattern, you can see it [here](https://github.com/actions/checkout#whats-new)

You can also add a link to the README of the `v1` for the previous version.

README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
...
+  uses: actions/cache@v2
+  with:
+    path: cache.tar
+    key: ${{ runner.os }}
+```
+
+* Increased perfomance and improved cache sizes using `zstd` compression
+> Note this feature is off for Windows runner that are using `bsdtar` (e.g., windows-latest hosted runner) due to a bug in ziping large random files with `bsdtar`
+* Allowed caching for all events with a ref
+> See [events that trigger workflow](https://help.github.com/en/actions/reference/events-that-trigger-workflows) for info on which events do not have a `GITHUB_REF`
```
joshmgross, line 59:
```suggestion
* Allowed caching for all events with a ref, see [events that trigger workflow](https://help.github.com/en/actions/reference/events-that-trigger-workflows) for info on which events do not have a `GITHUB_REF`
```

It might be easier to read if we combine this into one bullet point, the `>` note blocks break up the list.

README.md
```diff
@@ -8,14 +8,67 @@ This action allows caching dependencies and build outputs to improve workflow ex
...
+
 ## Usage
 
 ### Pre-requisites
 Create a workflow `.yml` file in your repositories `.github/workflows` directory. An [example workflow](#example-workflow) is available below. For more information, reference the GitHub Help Documentation for [Creating a workflow file](https://help.github.com/en/articles/configuring-a-workflow#creating-a-workflow-file).
 
 ### Inputs
 
-* `path` - A directory to store and save the cache
+* `path` - Directories to store and save the cache. Supports pattern matching, multipath and single file cache
```
konradpabjan, line 70: Are there plans to change this to `paths` since multiple paths are now supported or will this be kept as is?

joshmgross, line 70:
```suggestion
* `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/master/packages/glob) for supported patterns. 
```

We should update this description in the `action.yml` as well, since that will show up in the Workflow editor when users look up the inputs for this action.

examples.md
```diff
@@ -50,20 +50,32 @@ If you do not want to include them, consider to move the cache folder like below
...
-  - uses: actions/cache@v1
+  - uses: actions/cache@v2
     with:
       path: ${{ github.workspace }}/.nuget/packages
       key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
       restore-keys: |
         ${{ runner.os }}-nuget-
 ```
 
+With `actions/cache@v2` you can now exclude unwanted packages with [exclude pattern](https://github.com/actions/toolkit/tree/master/packages/glob#exclude-patterns)
```
joshmgross, line 61: >"If you do not want to include them, consider to move the cache folder like below."

Should we remove the above recommendation since the wild card exclude pattern is a better approach?

examples.md
```diff
@@ -108,27 +120,20 @@ steps:
 We cache the elements of the Cabal store separately, as the entirety of `~/.cabal` can grow very large for projects with many dependencies.
```
joshmgross, line 120: Let's update this description, since we're no longer caching these separately.

examples.md
```diff
@@ -422,42 +426,32 @@ When dependencies are installed later in the workflow, we must specify the same
...
 
 ```yaml
-- name: Cache cargo registry
-  uses: actions/cache@v1
-  with:
-    path: ~/.cargo/registry
-    key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
-- name: Cache cargo index
-  uses: actions/cache@v1
+- name: Cache cargo
```
joshmgross, line 429: We could optionally just remove this `name` from the example, it was only needed to differentiate the multiple cache steps and most examples don't specify a name.

README.md
joshmgross: On this note, I'm not sure if we need to specify this caveat about Windows. The "What's New" doesn't need to be an exhaustive description of every feature. We could just say 
>Increased performance and improved cache sizes using `zstd` compression for Linux and macOS runners.

There's a very small portion of Windows runners that will benefit as well, but a majority will not.

>Should we link to the bsd tar issue?

The issue is https://github.com/actions/cache/issues/301. Linking the issue here in the What's New section would be good if we keep the note about this feature being disabled for most Windows runners.

README.md
joshmgross: This will be kept as-is to make it easier to upgrade to `v2`.

The three options are:
1. Keep `path` as-is and make it easy to upgrade
2. Rename to `paths`, break any users who try to change `v1` to `v2` and not know about this input name change.
3. Support both `path` and `paths`, this might get confusing to users and we'd have to figure out what to do if a user specified both.

imbsky:
I've opened an additional PR, which isn't very rush but may also be worth including in the initial release of v2. https://github.com/actions/toolkit/pull/475

examples.md
aiqiaoy: My understanding is that we are still caching them separately, just in one step now. I think we should keep it so that users don't accidentally cache the entire `~/.cabal` directory and use up all the cache storage.

examples.md
aiqiaoy: I guess it depends on how many packages the user wants to exclude. Maybe move the wild card example up to show that it's a better approach?

action.yml
```diff
@@ -3,7 +3,7 @@ description: 'Cache artifacts like dependencies and build outputs to improve wor
 author: 'GitHub'
 inputs:
   path:
-    description: 'A directory to store and save the cache'
+    description: 'Directories to store and save the cache. Supports pattern matching, multipath and single file cache'
```
joshmgross, line 6: Can we use the description from the README? We're not restricted to only directories

README.md
```diff
@@ -8,14 +8,36 @@ This action allows caching dependencies and build outputs to improve workflow ex
...
+  uses: actions/cache@v2
+  with:
+    path: |
+      ~/cache
+      !~/cache/exclude
+      **/node_modules
+    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
+
+* Increased performance and improved cache sizes using `zstd` compression for Linux and macOs runners
```
joshmgross, line 26:
```suggestion
* Increased performance and improved cache sizes using `zstd` compression for Linux and macOS runners
```

mkevre (commented):
aiyan/v2-release-doc

PR #338, opened 2020-05-30
Update Scala - SBT example to include Coursier cache

daniel-shuy:
Previously I didn't include an example for SBT 1.3+ in #134 since the cache size limit was too small for Coursier (SBT 1.3 uses Coursier instead of Ivy, see https://github.com/actions/cache/issues/59#issuecomment-552846834). Since the cache size limit has been increased in [v1.1.1](https://github.com/actions/cache/releases/tag/v1.1.1), this PR adds an example for SBT 1.3.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -443,12 +443,16 @@ SBT 1.2
   uses: actions/cache@v2
   with:
     path: ~/.ivy2/cache
-    key: ${{ runner.os }}-sbt-ivy-cache-${{ hashFiles('**/build.sbt') }}
+    key: ${{ runner.os }}-sbt-ivy-cache-${{ github.run_id }}
+    restore-keys: |
+      ${{ runner.os }}-sbt-ivy-cache-
 - name: Cache SBT
   uses: actions/cache@v2
   with:
     path: ~/.sbt
-    key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
+    key: ${{ runner.os }}-sbt-${{ github.run_id }}
+    restore-keys: |
+      ${{ runner.os }}-sbt-
 ```
 
 SBT 1.3.0-1.3.10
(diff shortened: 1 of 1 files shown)


daniel-shuy:
Any update on this?

examples.md
```diff
@@ -437,13 +437,68 @@ When dependencies are installed later in the workflow, we must specify the same
 
 ## Scala - SBT
 
+SBT 1.2
 ```yaml
+- name: Cache SBT ivy cache
+  uses: actions/cache@v2
+  with:
+    path: ~/.ivy2/cache
+    key: ${{ runner.os }}-sbt-ivy-cache-${{ hashFiles('**/build.sbt') }}
```
x7c1, line 446: I guess we cannot use the hash of build.sbt, since it's not always changed when dependencies are changed.

No need to see a specific file (like Gemfile.lock, package-lock.json, etc) to know this, downloaded modules are shared globally on a machine (as you can see on the above `~/.ivy2/cache`).

I suppose `restore-keys` can solve this instead, which can found existing cache (on every run if dependecies are changed or not):
 
```yaml
        key: ${{ runner.os }}-sbt-ivy-cache-${{ github.run_id }}
        restore-keys: |
          ${{ runner.os }}-sbt-ivy-cache-
```

(but not sure if this restore-keys can restore last key actually.)

daniel-shuy: Oh right, I forgot you can define dependencies outside `build.sbt`.

Maybe we should always load from the cache regardless if the dependencies have changed?
```yml
  uses: actions/cache@v2
  with:
    path: ~/.ivy2/cache
    key: ${{ runner.os }}-sbt-ivy-cache
```

x7c1: I'm afraid it doesn't work.
An existing cache, according to the documents, should not be updated:

https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

> Once you create a cache, you cannot change the contents of an existing cache but you can create a new cache with a new key.

and I've found the sentense:

> When the action finds a partial match, the most recent cache is restored to the path directory.

Looks good to use `restore-keys`.

daniel-shuy: Thanks, I've updated the PR accordingly

OndrejSpanel (approved):
I think this really should be merged now, as most active projects use SBT 1.5 or new and current example does not work for them.

vsvipul:
Hi @daniel-shuy . Apologies for looking at this so late. But I think we should only keep the example for the latest version in the docs to avoid making it too long. 
Also, if you see this, can you please resolve the merge conflicts. Thank you.

vsvipul:
Closing this for now. Feel free to reopen if you come back.

Issue #330, opened 2020-05-26
Cache misses due to path-based versioning and its normalization


mzabaluev:
[This run](https://github.com/mzabaluev/github-actions-playground/actions/runs/116171611) using `actions/cache@v2` shows how a cache created by the `update-deps` job is not hit by the jobs in the `test` matrix that depend on `update-deps` and use the same key. This works in `actions/cache@v1`, so it looks like a regression to me.

What's more peculiar, each of the three `test` jobs reported to have successfully stored its cache state under this same key.
In the [next run](https://github.com/mzabaluev/github-actions-playground/actions/runs/116174830), all jobs hit the cache, but which state did they get?

dhadka:
Hey @mzabaluev, thanks for spotting this.  Can you please enable [step debugging](https://help.github.com/en/actions/configuring-and-managing-workflows/managing-a-workflow-run#enabling-debug-logging), change the key, and re-run the workflow?  This will provide some more output so we can see what's going on.

mzabaluev:
@dhadka [Here you are](https://github.com/mzabaluev/github-actions-playground/actions/runs/116368502)

I see that the 'Post Cache cargo registry index' step in the producer job gets to write its cache first:
```
Resource Url: https://artifactcache.actions.githubusercontent.com/SwZ48ggH2fmptrBol7q4o6DwD74REZ2KvEjuDiE8Jb9dgsMc8M/_apis/artifactcache/caches/58
```

Then the 'Post Cache cargo registry' step allocates at the same resource:

```
Resource Url: https://artifactcache.actions.githubusercontent.com/SwZ48ggH2fmptrBol7q4o6DwD74REZ2KvEjuDiE8Jb9dgsMc8M/_apis/artifactcache/caches/59
```

There is also intra-action state `CACHE_KEY` which is set by the caching steps prior to the above, but in the opposite order:

```
2020-05-27T02:11:21.4628234Z ##[debug]Save intra-action state CACHE_KEY = cargo-registry-debug1-39f2b8d68761c35ef5f7317c495a2e84f0aed5d34a5d724b2144d58bf6ed66b3
...
2020-05-27T02:11:22.0139963Z ##[debug]Save intra-action state CACHE_KEY = cargo-index-debug1-5d9c7f87b512b22e0543a050c81e91b650b44aff
```

A consumer job queries and fails with:

```
[...]

PR #343, opened 2020-06-02
Improve string split to handle whitespace

dhadka:
Currently, the following two cache definitions result in two separate caches:

```
- uses: actions/cache@v2
  with:
    path: |
      ~/cache
      !~/cache/exclude
```

```
- uses: actions/cache@v2
  with:
    path: |
      ~/cache 
      !~/cache/exclude
```

Don't see any difference? There's an extra whitespace in one.  This results in different version hashes, which ends up creating two different caches.

This PR calls `trim()` on each entry in the `path` and `restore-keys` to remove any leading and trailing whitespace.  Additionally, this moves the split logic to a separate method and adds unit test coverage.

### DIFF
--- a/dist/save/index.js
+++ b/dist/save/index.js
@@ -5306,7 +5306,7 @@ var __importStar = (this && this.__importStar) || function (mod) {
     return result;
 };
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.isValidEvent = exports.logWarning = exports.getCacheState = exports.setOutputAndState = exports.setCacheHitOutput = exports.setCacheState = exports.isExactKeyMatch = void 0;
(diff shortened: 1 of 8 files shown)


dist/save/index.js
```diff
@@ -5350,6 +5350,14 @@ function isValidEvent() {
     return constants_1.RefKey in process.env && Boolean(process.env[constants_1.RefKey]);
 }
 exports.isValidEvent = isValidEvent;
+function getInputAsArray(name, options) {
+    return core
+        .getInput(name, options)
+        .split("\n")
+        .map(s => s.trim())
```
mzabaluev, line 5357: I suppose this gets rid of the "\r" on Windows as well, which I believe was the issue observed in https://github.com/actions/cache/issues/330#issuecomment-634505960

dhadka: Yes, I also added a test for this case.

dhadka:
@mzabaluev Those are both good suggestions.  I've filed an enhancement here - https://github.com/actions/cache/issues/345.

PR #351, opened 2020-06-15
Add note about using setup-node before cache

joshmgross:
When using both `actions/cache` and `actions/setup-node`, it's recommended to use `actions/setup-node` first if `npm config` is used to retrieve the cache directory. Outside of this scenario, the order should not matter.


YraIssa:
joshmgross/setup-node-docs

PR #353, opened 2020-06-18
Allow updating cache with 'update' key

### LINKED ISSUE #342: Feature request: option to update cache
**Problem Description**
Currently. the cache action either restores an existing cache on cache-hit, or generates a missing cache on cache-miss. It does not update the cache on cache-hit.

This works well for caching static dependencies, but not for caching build artefacts.

**Proposed Solution**
Add an option allowing the user to enable cache updates.
This should be `false` by default to retain backwards-compatibility.
 
```yml
uses: actions/cache@v2
with:
  path: ccache
  key: ${{ matrix.CONFIG }}-${{ matrix.CXX }}-${{ matrix.TYPE }}
  update: true   # <~~ explicitly request an update
```

**Motivation**
Some programming languages benefit greatly from build caching. C++ in conjunction with ccache is the prime example. Using caching commonly decreases compilation times by at least 70%. Medium-sized projects easily take 20 minutes to compile.
ccache also manages the cache size itself and automatically removes obsolete entries, thus the cache won't explode with continuous updates.

It also saves time and money for both user and provider. The environment will be happy too.

davidsbond:
Closes #342 

This change adds an `update` option that allows the cache to be updated if the key matches.

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -194,3 +194,13 @@ test("getInputAsArray handles empty lines correctly", () => {
     testUtils.setInput("foo", "\n\nbar\n\nbaz\n\n");
     expect(actionUtils.getInputAsArray("foo")).toEqual(["bar", "baz"]);
 });
+
+test("getInputAsBoolean returns true if the value is set to 'true'", () => {
+    testUtils.setInput("foo", "true");
+    expect(actionUtils.getInputAsBoolean("foo")).toEqual(true);
+});
+
+test("getInputAsBoolean returns false if the value is set to anything else", () => {
+    testUtils.setInput("foo", "false");
+    expect(actionUtils.getInputAsBoolean("foo")).toEqual(false);
+});
--- a/__tests__/save.test.ts
+++ b/__tests__/save.test.ts
@@ -118,6 +118,41 @@ test("save with exact match returns early", async () => {
     expect(failedMock).toHaveBeenCalledTimes(0);
 });
 
+test("save with exact match updates when configured", async () => {
+    const failedMock = jest.spyOn(core, "setFailed");
+
+    const primaryKey = "Linux-node-bb828da54c148048dd17899ba9fda624811cfb43";
+    const savedCacheKey = primaryKey;
+
+    jest.spyOn(core, "getState")
+        // Cache Entry State
+        .mockImplementationOnce(() => {
+            return savedCacheKey;
+        })
+        // Cache Key State
+        .mockImplementationOnce(() => {
+            return primaryKey;
+        });
+
+    const inputPath = "node_modules";
+    testUtils.setInput(Inputs.Path, inputPath);
+    testUtils.setInput(Inputs.Update, "true");
+
+    const cacheId = 4;
+    const saveCacheMock = jest
+        .spyOn(cache, "saveCache")
+        .mockImplementationOnce(() => {
+            return Promise.resolve(cacheId);
+        });
+
+    await run();
+
+    expect(saveCacheMock).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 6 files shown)


davidsbond:
Having a bit of trouble getting the tests to work, I'm not super familiar with TS/JS, so if anyone could help me out getting tests written for the main logic change that'd be great.

elizabrock:
@davidsbond I just sent you a PR on your PR.  With that, the tests pass and your updates to the action can be run.

However, the underlying actions library will need to be updated to accommodate your changes.

You can see this here: 

Screen Shot 2020-06-25 at 9 58 13 AM

I think that with some creativity, we could get around this but I haven't figured that part out.

davidsbond:
Hey @elizabrock thanks for the PR PR, I'll merge your PR into my PR :)

davidsbond:
@elizabrock somewhere in https://github.com/actions/toolkit appears to be the place we'd need to make further changes

davidsbond:
Having an initial look it seems like the constraint could be server side, so perhaps this is not possible

aparcar:
I'm currently using the patch and it seem to work great. Thanks for the work!

eyal0:
The problem with this is the the `update` flag needs to be specified before the cache is downloaded.  It would be nicer if we could specify it *afterwards*.  Imagine that the cache is caching the latest version of some downloaded source code.  If the source code on the web matches the source code in the cache, use the cache.  Otherwise, download the new source code and re-build it and cache that.

So the steps in my CI would be:

1. Download the cache.
2. Compare the version of the utility in the cache to the version on the web.
3. If the compare matches, set `update` to `false`.  Otherwise `true`.
4. Also, if the compare doesn't match, re-download and rebuild that utility.
4. Do the rest of the CI.
5. In the Post-CI action, obey the update flag as needed.

There's no way to do this with the update flag as explained above because the need to update isn't known until after the cache is downloaded.

paresy:
Unfortunately this does not work for me (anymore?). It fails with the following error: `Unable to reserve cache with key build-ubuntu-ccache, another job may be creating this cache.` I think the problem is already documented here: https://github.com/actions/toolkit/issues/505

eyal0:
I'm seeing that error, too.  Seems like a new error?  Maybe something is broken with the underlying cache?

eyal0:
Did it used to work?  I thought that it worked at one point...

eyal0:
> Having an initial look it seems like the constraint could be server side, so perhaps this is not possible

I agree with this.  When you want to upload to the server, you need a `cacheId` which the server gives you.  The server returns a cacheId of -1 when you try to reserve a cacheId that already exists.  I'll keep looking into this.

davidsbond:
@eyal0, yeah. I'm surprised by reports from people that this works. As from my comment last year it looks like the have a constraint server-side on cache keys. None of the maintainers have done anything with my PR and issues, so I think it's unlikely for this feature to get picked up. We'll have to wait and see.

eyal0:
Maybe it was a fluke and it used to work and it stopped working?  Or maybe it would only work on the first time so it *appeared* to work.  Also, I think that it will work if the cache gets cleared out in some other way, like maybe the cached entry expires.

Anyway, if the server won't co-operate then this will be difficult to do!  It might be at the server has some hidden command to delete a cache entry.  We could delete it before trying to reserve a cacheId for a new one.  But how would we ever discover this secret command?

I guess that another solution would be to add a random suffix to key.  That would allow it to upload.  The trick would then be to **disable** uploading if the cache was found, because you don't want to upload every run and waste time and space.

eyal0:
@davidsbond Here's my workaround: https://github.com/actions/cache/pull/498#issuecomment-807300432

Kind of a hack but I think that it's working.

VakninAmit (approved):

lcswillems:
Whats the reason it has been closed?

Issue #366, opened 2020-07-11
Github Actions / Service has permissions issues with cache files on windows


RevoluPowered:
**On non self hosted runners I'm getting:**

```
../../../../../../.scons_cache/08/088da2b137de218e3019b1cb196248ee: Can't create '\\\\?\\.scons_cache\\08\\088da2b137de218e3019b1cb196248ee'
../../../../../../.scons_cache/07/0711ff940b39b6e4d8a6d9093fdfbefa: Can't create '\\\\?\\.scons_cache\\07\\0711ff940b39b6e4d8a6d9093fdfbefa'
../../../../../../.scons_cache/07/074bf38f0ada07b13e2afa79146ada1c: Can't create '\\\\?\\.scons_cache\\07\\074bf38f0ada07b13e2afa79146ada1c'
../../../../../../.scons_cache/06/06e0868992c3187a7d95a3fc9bf288a0: Can't create '\\\\?\\.scons_cache\\06\\06e0868992c3187a7d95a3fc9bf288a0'
../../../../../../.scons_cache/05/05ce05c3b99bcddcc6c5a96ac0700a55: Can't create '\\\\?\\.scons_cache\\05\\05ce05c3b99bcddcc6c5a96ac0700a55'
../../../../../../.scons_cache/05/05f17c7fb7ad134042bc8293eee959c0: Can't create '\\\\?\\.scons_cache\\05\\05f17c7fb7ad134042bc8293eee959c0'
../../../../../../.scons_cache/04/0409c7d9c5987d35edcd10581a34daab: Can't create '\\\\?\\.scons_cache\\04\\0409c7d9c5987d35edcd10581a34daab'
../../../../../../.scons_cache/03/034ef1fcd439b0cc6f14aaca365c0a03: Can't create '\\\\?\\.scons_cache\\03\\034ef1fcd439b0cc6f14aaca365c0a03'
../../../../../../.scons_cache/02/02a30a25ca6b73691e4759c2c93433d3: Can't create '\\\\?\\.scons_cache\\02\\02a30a25ca6b73691e4759c2c93433d3'
../../../../../../.scons_cache/02/02ffc7c73449955dd78ed7d9ec738557: Can't create '\\\\?\\.scons_cache\\02\\02ffc7c73449955dd78ed7d9ec738557'
[...]

Issue #361, opened 2020-07-05
Caching outside of GITHUB_WORKSPACE problematic


mcg1969:
Here's an example form of the command that `actions/cache@v2` uses on a mac to create a cache:
```
/usr/bin/tar --use-compress-program zstd -T0 -cf cache.tzst -P -C /Users/runner/runners/2.169.1/work/conda-pack/conda-pack --files-from manifest.txt
```
In this case, the name of the repository is `conda-pack`, and `.../work/conda-pack/conda-pack` is the root directory of the checked-out git repository.

When I first began building an action for this repository, I wanted to cache something in the `$HOME` directory. This seemed to work fine on Linux, but on Mac, I was getting these errors:
```
/usr/bin/tar --use-compress-program zstd -d -xf /Users/runner/work/_temp/a2e257c1-bc6e-422a-a46f-5f0b9f8b21e7/cache.tzst -P -C /Users/runner/work/conda-pack/conda-pack
../../../../../conda/: Can't create '../../../../../conda'
../../../../../conda/man/: Failed to create dir '../../../../../conda'
../../../../../conda/conda-meta/: Failed to create dir '../../../../../conda'
../../../../../conda/condabin/: Failed to create dir '../../../../../conda'
../../../../../conda/ssl/: Failed to create dir '../../../../../conda'
../../../../../conda/bin/: Failed to create dir '../../../../../conda'
```
It makes sense that a tar file cannot unpack to relative paths that fall outside of the root directory of the extraction. And to be clear, I've determined that for technical reasons I don't want to be in `$HOME` in Windows.

However, your instructions include numerous examples of caching paths outside of the `GITHUB_WORKSPACE` directory. It's unclear to me why my approach did not work. I would definitely like to be able to cache things outside of `GITHUB_WORKSPACE`—even if it is just one directory up. I don't want to have to "corrupt" my git repository clone.

Below are the relevant logs and workflow spec, which might have bugs for other reasons, mind you.

https://github.com/conda/conda-pack/runs/837672740?check_suite_focus=true
https://github.com/conda/conda-pack/actions/runs/157856418/workflow

mcg1969:
Actually I think the real problem may that the location of `GITHUB_WORKSPACE` is inconsistent! It's at different depths, which throws off any attempt to use these relative paths outside of the root, even if `tar` were to support them.

[...]

PR #367, opened 2020-07-11
Fix caching not restoring on windows hosts

RevoluPowered:
TAR is in windows-latest by default, no local is not the correct approach for the Windows 2019 / Latest image and it will break the download of artifacts. 

Godot needs this to work now since we use this feature to replace AppVeyor for our CI.

I think the confusion is Win32 TAR vs GNU TAR


RevoluPowered:
Fixes #366 #361

vsvipul:
Adding @pdotl  to reply to this PR and see if this will be helpful.

pdotl:
@vsvipul the changes here are stale and we are working on adding GNU tar as the default tar. Closing this PR since it is outdated. @RevoluPowered please open an issue if you still think this is needed.

Issue #370, opened 2020-07-14
Potential for hung steps due to implicit process exit


jclem:
**tl;dr** When this action's [`run`](https://github.com/actions/cache/blob/eed9cfe64d00dd64cbb36bec915d41dd6fab9f6c/src/restore.ts#L63) function completes, there is a potential for jobs to [hang](https://github.com/android-password-store/Android-Password-Store/runs/868857729?check_suite_focus=true) since we aren't explicitly exiting the process (due to other bugs leaving active event listeners around).

The Node.js runtime keeps track of _active event listeners_. When there are no more active event listeners, the [program exits](https://bytearcher.com/articles/why-does-hello-world-exit-but-http-server-doesnt/). This is why a process that just runs `console.log("Hello, world")` exits when the statement executes, but a process that starts an HTTP server doesn't exit after the server has started.

In order to prevent this action from hanging in the event that there are active event listeners somehow still around once `run()` completes, we should explicitly exit the process:

```javascript
run().then(() => process.exit())
```

PR #371, opened 2020-07-14
Exit after run to prevent hanging if there are active listeners

dhadka:
Closes https://github.com/actions/cache/issues/370

If there are active listeners after `run()` finishes, NodeJS will not exit.  Generally, leaving an active listener is a bug that should be fixed, but to prevent this from causing the step to hang we can explicitly exit.


src/restore.ts
```diff
@@ -60,6 +60,6 @@ async function run(): Promise<void> {
     }
 }
 
-run();
+run().then(() => process.exit(0));
```
joshmgross, line 63: I think this is the wrong approach here. If we're confident that these hanging events are only from error cases, then we should have a corresponding `process.exit` in the `catch` blocks. For save, we'd want a success status code and for restore we'd want a failure status code.

The downside of this current implementation is that we're exiting with a success status code when the step may have failed.

joshmgross:
Could you double-check the generated dist files? I wouldn't expect 100k lines to be changed for those.

jclem:
Whoops, this should be `process.exit()` so that we honor any `process.exitCode` previously set.

src/restore.ts
joshmgross: Sounds like `process.exit()` is what we want to honor the exit code set by `core.setFailed`
https://github.com/actions/cache/pull/371#issuecomment-658296834

jclem:
Ah, as @ericsciple pointed out, we need to flush stdout/stderr, so just calling `process.exit()` isn't safe.

jclem:
Another option would be to set a timeout for writing stdout/stderr and then force exit:

```typescript
const OUTPUT_TIMEOUT = 60_000

function runSafeMain(fn: (...args: any[]) => Promise<unknown>) {
  fn().finally(waitForOutput)
}

function waitForOutput() {
  const timeout = setTimeout(() => {
    console.error('Output timeout exceeded')
    process.exit()
  }, OUTPUT_TIMEOUT)

  timeout.unref()
}
```

dhadka:
@joshmgross Fixed the `dist/` files having tons of changes. Had it pointing to a local copy of `@actions/toolkit`.  (The test that looks for uncommitted changes also detected this issue).

I'll wait until I hear back from `@jclem` and `@ericsciple` on how to handle flushing before doing anything with this.

aiqiaoy:
```
[warning]downloadCache failed: Request timeout: /758370072ebc48a1bae6dd969702511b/59eed5cca5c5ea119b0600155dafe08a?sv=2018-03-28&sr=b&sig=4O4ae0FEDWtyXHalSOBL7FMQNrPJdJSajOxwzi2WJsY%3D&se=2020-07-14T11%3A18%3A50Z&sp=r&rscl=x-e2eid-ca91503f-2fdf483a-86d18754-8bf30f0c
8
##[error]The operation was canceled.
```

It seems from the log the request got stuck after the timeout error. I suspect it's because of a bug in the http-client package we use (code [here](https://github.com/actions/http-client/blob/5f61f72a73224ee5541c6961f864e09e9f4cf6d7/index.ts#L498)) but I'm not proficient enough in nodejs runtime to figure out where.. @jclem Any idea?

aiqiaoy:
@dhadka I'm changing default branch to `main` for cache so I rebased your PR to on top of `main`

bishal-pdMSFT:
@vsvipul are you reactivating this PR?

vsvipul:
@bishal-pdMSFT  I'm not sure if this is still required. I'll check and follow up accordingly.

vsvipul:
Closing this for now, as we aren't facing any issues related to this.

PR #372, opened 2020-07-15
Rename master to main

aiqiaoy:


.github/workflows/workflow.yml
```diff
@@ -3,12 +3,14 @@ name: Tests
 on:
   pull_request:
     branches:
+      - main
       - master
```
konradpabjan, line 7: `master` can go away. I did that for the artifact actions and setup-python.

Long term any PRs or new changes won't go into master so there is no point in keeping it around

PR #375, opened 2020-07-16
Bump cache module to version 1.0.0

dhadka:
Use version 1.0.0 of the `@actions/cache` [module](https://www.npmjs.com/package/@actions/cache).  Cache files hosted on Azure will now use the Azure Storage SDK for NodeJS to download the files, replacing the [http-client](https://github.com/actions/http-client).

Also bumps the cache action version to 2.1.0.


.github/workflows/workflow.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       shell: bash
       run: |
         npm run build
-        if [ "$(git status --porcelain | wc -l)" -gt "0" ]; then
+        if [ "$(git diff --ignore-space-at-eol | wc -l)" -gt "0" ]; then
```
dhadka, line 55: `tslib.js` in `@azure/core-auth` uses CRLF line endings which gets flagged by this step.  Ignoring EOL characters.

dhadka:
This PR had two CodeQL warnings coming from the new Azure libraries.  Marked as false positive.  We were reading the AZURE_LOG_LEVEL env var and printing the value to the console.  As that env var is not a secret, it is not a security concern.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cache",
-  "version": "1.1.2",
+  "version": "2.1.0",
```
aiqiaoy, line 3: Why are we bumping the cache action version?

dhadka: Sorry if I'm not understanding, but shouldn't this match the tag / release version?  I'm planning to release this change as `@v2.1`

PR #384, opened 2020-07-27
Update examples.md for rust cargo

jensim:
Update the example for rust so that the cache key actually works as intended

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -456,7 +456,9 @@ When dependencies are installed later in the workflow, we must specify the same 
       ~/.cargo/registry
       ~/.cargo/git
       target
-    key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    key: cargo-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -456,7 +456,9 @@ When dependencies are installed later in the workflow, we must specify the same
       ~/.cargo/registry
       ~/.cargo/git
       target
-    key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    key: cargo-${{ hashFiles('**/Cargo.lock') }}
+    restore-keys: |
+      cargo-
```
mzabaluev, line 461: This has a disadvantage: without pruning, the crates cached in the `registry` and `git` subdirectories will accumulate by being handed down from earlier cache states to the fresh one.

mzabaluev:
See #325 for some more elaborate examples that avoid caching redundant data and limit content for inexact restoration. Basically, you only need `~/.cargo/registry/index` to be treated this way.

Issue #326, opened 2020-05-25
Preserve exact timestamps


OlegYch:
Some tools (like https://www.scala-sbt.org/) rely on files timestamps with millisecond precision.
Currently cache preserves timestamps only down to second precision.
The fix would be to simply use posix format when tarring files.
https://github.com/actions/cache/pull/233 was my attempt at fixing it but it no longer works with current master

jcayzac:
I really need this too!
@OlegYch meanwhile, please do not delete your `posix` branch as my workflow depends on it :)

jcayzac:
Actually, after spending two days trying to get incremental builds to work, I just remembered git doesn't support timestamps and I gave up.

jcayzac:
I guess I need an action that takes the timestamp of the base ref, applies it to all files, then takes the timestamp of the head ref and applies it to all modified/added files.

jcayzac:
> I guess I need an action that takes the timestamp of the base ref, applies it to all files, then takes the timestamp of the head ref and applies it to all modified/added files.

I ended up using [chetan/git-restore-mtime-action](https://github.com/chetan/git-restore-mtime-action) for that.

### Example steps:

```yaml
- name: 'Restore timestamps'
  uses: chetan/git-restore-mtime-action@v1.0

- name: 'Cache: Derived data'
  uses: OlegYch/cache@posix
  id: cache-derived-data
  with:
    path: .build/derived-data
    key: ${{ runner.os }}-deriveddata-${{ github.base_ref }}@${{ github.event.pull_request.base.sha }}-${{ github.head_ref }}@${{ github.sha }}
    restore-keys: |
      ${{ runner.os }}-deriveddata-${{ github.base_ref }}@${{ github.event.pull_request.base.sha }}-${{ github.head_ref }}@
      ${{ runner.os }}-deriveddata-${{ github.base_ref }}@${{ github.event.pull_request.base.sha }}-
      ${{ runner.os }}-deriveddata-${{ github.base_ref }}@
```

OlegYch:
@jcayzac posix branch does not work as it's based on broken master, posixv1 works, i'm going to keep it until this is fixed in upstream

jcayzac:
For reference, here's the commit https://github.com/OlegYch/cache/commit/a39be91a23ff474b988f47a25b952db79045befa

OlegYch:
@jcayzac yep that one works

imbsky:
[...]

PR #390, opened 2020-08-05
Update the @actions/cache package to the latest version

smorimoto:
As the title says.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@actions/core": "^1.2.0",
     "@actions/exec": "^1.0.1",
     "@actions/io": "^1.0.1",
-    "@actions/cache": "^1.0.1"
+    "@actions/cache": "^1.0.2"
   },
   "devDependencies": {
     "@types/jest": "^24.0.13",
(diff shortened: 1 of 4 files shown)


smorimoto:
The main purpose of this PR is to update the @actions/cache for https://github.com/actions/toolkit/pull/533 and https://github.com/actions/cache/issues/326. But I think we can update other dependencies with this PR.

smorimoto:
@aiqiaoy @joshmgross Could you review this PR?

package.json
```diff
@@ -23,10 +23,10 @@
...
   "license": "MIT",
   "dependencies": {
-    "@actions/core": "^1.2.0",
-    "@actions/exec": "^1.0.1",
-    "@actions/io": "^1.0.1",
-    "@actions/cache": "^1.0.1"
+    "@actions/core": "^1.2.4",
+    "@actions/exec": "^1.0.4",
+    "@actions/io": "^1.0.2",
+    "@actions/cache": "^1.0.2"
```
konradpabjan, line 29: Can we only update `@actions/cache`? We intentionally do not update all NPM packages to the latest versions unless required.

`npm update @actions/cache` and then `npm install` and `npm run release` should do the trick

smorimoto: I changed it.

PR #396, opened 2020-08-13
Update Ruby - Bundler example

sonicdoe:
Updates the “Ruby - Bundler” example to use [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby)’s examples from [Caching `bundle install` automatically](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) and [Caching `bundle install` manually](https://github.com/ruby/setup-ruby#caching-bundle-install-manually).

Closes #289.
Closes #290.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -430,21 +430,28 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
 
 ## Ruby - Bundler
 
+[`ruby/setup-ruby`](https://github.com/ruby/setup-ruby) provides a way to automatically run `bundle install` and cache the result:
+
+```yaml
+- uses: ruby/setup-ruby@v1
+  with:
+    bundler-cache: true
+```
+
+If necessary, you can also cache gems manually using `actions/cache`:
+
 ```yaml
 - uses: actions/cache@v2
   with:
     path: vendor/bundle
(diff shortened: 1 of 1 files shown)


sonicdoe:
If this pull request gets merged, it’d be good to remove the “The code above is a more complete version of the [Ruby - Bundler example](https://github.com/actions/cache/blob/master/examples.md#ruby---bundler).” sentence from [setup-ruby’s readme](https://github.com/ruby/setup-ruby/blob/master/README.md#caching-bundle-install-manually).

examples.md
```diff
@@ -409,22 +409,15 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
 
 ## Ruby - Bundler
 
+[`actions/setup-ruby`](https://github.com/ruby/setup-ruby) already provides a way to automatically run `bundle install` and cache the result:
```
aiqiaoy, line 412: This is great but I think we should keep the manual install example too. Reasons being we want this page to be the main place for people to find examples of using cache action and we don't want broken link in case the `setup-ruby` readme changes.

sonicdoe: I’ve now updated the example to include both the automatic way (using the `bundler-cache` option) as well as the manual way (using `actions/cache`).

PR #402, opened 2020-08-19
Add a reference to gradle-cache-action to Java-Gradle configuration

vlsi:
https://github.com/burrunan/gradle-cache-action is [@actions/cache](https://github.com/actions/toolkit/tree/main/packages/cache)-based action for better caching when Gradle is used.

It comes with defaults, so there's no need to configure paths.

PS. I add the exclusion of `!~/.gradle/wrapper/dists/**/gradle*.zip` to the default configuration, as the file (100-150MiB) is used one time only (see https://github.com/gradle/gradle/issues/3605), however, I include the change into the single PR to avoid unexpected merge conflicts between the two

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -133,6 +133,9 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
 
 ## Java - Gradle
 
+Consider using [gradle-cache-action](https://github.com/burrunan/gradle-cache-action) as it provides fine-grained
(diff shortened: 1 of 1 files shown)


vlsi:
Just in case, the way to use `gradle-cache-action` would as follows, however, I'm not sure it needs to be added to `actions/cache` examples.

```yaml
- uses: burrunan/gradle-cache-action@v1
  name: Build
  with:
    job-id: jdk${{ matrix.jdk }}
    arguments: --scan build
```

Issue #321, opened 2020-05-21
Add a check-only mode or action


mzabaluev:
In some workflows, there are jobs whose only purpose is to ensure that proper cache state exists for other jobs to pick up. If such a job hits an already stored state, downloading that state for this job only wastes time and network bandwidth because the job does not need to do anything. Here's an [example](https://github.com/mzabaluev/github-actions-playground/blob/6344e933320075660d9d3028d78df2f78a2e23d0/.github/workflows/test.yml#L4).

It would be nice to add a mode that only checks existence of the key and sets the `cache-hit` output.

DocX:
Have similar case.
We have workflow with unit tests parallelized using the matrix.

In order to run bundle install just once (and not in each of the parallel jobs), we have first job that builds the cache (ie installs the ruby bundle).

If the cache is already hit, we don't need to download it, because the second step in the job have `if: steps.bundle-cache.outputs.cache-hit != 'true'`

As @mzabaluev mentioned, having option to only "check" the cache hit / skip the download if cache is hit, would save bandwidth and time

MartijnHols:
I assume you would want the cache to still save. I can not think of any reason to want a check-only mode, where it doesn't save either unless you're going to define two cache steps in your job and you're going to make the second depend on the first's check result but that seems more cumbersome and slower than just a save-only-if-changed mode.

A save-only-if-changed mode also seems kind of confusing, because what if for some other reason you still don't do your build and now it saved under the key? You now have a broken cache. You can't put the saving under the if-cache hit-condition as that's a post run and uncontrollable.

I'm thinking for advanced usages perhaps it would be better to split the cache action into 3 parts where you have to manually add a save step whereas the standard cache action does that automatically. The 3 parts would be `cache-check`, that only peeks and does nothing else, `cache-restore` that only restores for a [read-only](https://github.com/actions/cache/pull/474) mode, and a `cache-save` for storing changes. Then you can do a job with the steps:

* cache-check
* build (if !cache-check.cache-hit)
* save  (if !cache-check.cache-hit)

That seems easier to follow and maintain when you want to optimize for performance beyond what the standard cache action (restore+saving) does. 

(it's also quicker in case the post-if is useless as just parsing and executing a JS post run that does nothing can take 10 seconds)

MartijnHols:
I have a `save-only` flag implemented here: https://github.com/MartijnHols/cache/tree/feature/save-only, just waiting for #474 to be merged before making a PR since that branch includes those changes.

[...]

PR #420, opened 2020-09-29
Add the ability to only check for a cache key without restoring

bowd:
### Overview

In some situations it helps to have a way to only check that a cache key exists without spending the extra time to download the cache: for example when you have a job that handles building and a separate job that handles running tests. 
The job that builds is a NOOP when the cache exists but it needs to verify first. If the build is pretty large that adds time spent downloading a payload that gets ignored and then downloaded again in the other job.

Fixes: #321 

### ❗ Until this gets merged

This can be used today by referencing `bowd/cache@v3` in the actions instead of `actions/cache@v2`
The input is `only-check-key`. If that is set to `true` it will skip the downloading but output the same `cache-hit` as the normal flow.


src/restore.ts
```diff
@@ -1,4 +1,6 @@
 import * as cache from "@actions/cache";
+import { getCacheEntry } from "@actions/cache/lib/internal/cacheHttpClient";
```
dhadka, line 2: Thanks for making this change!  I would personally prefer adding a new method to `@actions/cache` that we could call here, such as:

```
    if (onlyCheck) {
        const result: boolean = cache.containsKey(primaryKey, cachePaths)
        utils.setCacheHitOutput(result);
    }
```

rather than exposing these internals.  Not only would this simplify the logic here but also allow anyone consuming the `@actions/cache` module to take advantage of this enhancement.  Thoughts?

bowd: I agree that would be the cleanest way. The only reason I did it this way initially as a proof of concept is because I can actually use it and see it in action. Because `@actions/toolkit` is a monorepo wrapping multiple NPM packages it's also impossible to reference them pointing to a git branch, thus I couldn't build this repo, pointing to a fork of `@actions/toolkit` 

But I'm happy to link this to a PR on `@actions/toolkit` as well and keep this open as an alternative until that gets merged and published, after which we can finalise the implementation here.

bowd:
Also unrelated to this CI is failing because of this: https://github.com/actions/toolkit/pull/589

src/restore.ts
dhadka: > But I'm happy to link this to a PR on @actions/toolkit as well and keep this open as an alternative until that gets merged and published, after which we can finalise the implementation here.

Sounds good to me, thanks!

supermario:
If anyone else stumbles across this, https://github.com/marketplace/actions/cache now has `lookup-only`.

PR #421, opened 2020-09-29
Caching action should no-op on GHES

dhadka:
No-op and display a helpful message instead of displaying 4xx errors when running on GHES.


src/restore.ts
```diff
@@ -6,6 +6,12 @@ import * as utils from "./utils/actionUtils";
 
 async function run(): Promise<void> {
     try {
+        if (utils.isGhes()) {
+            core.info("Cache action is not supported on GHES");
```
joshmgross, line 10: Should this be a warning or error?

dhadka: Changed to warning

joshmgross (approved):

Issue #415, opened 2020-09-13
CACHE_UPLOAD_CHUNK_SIZE ignored?


johannesstricker:
The post cache task continues to fail for my local runner because it seems to be hitting the 60secs upload timeout. I've tried to reduce the `CACHE_UPLOAD_CHUNK_SIZE` as mentioned in https://github.com/actions/cache/issues/259#issuecomment-627448818, but from the `start` and `end` of the uploaded chunk that you can see in the screenshot below, it appears to me that the chunk size is still the default 32mb.

![image](https://user-images.githubusercontent.com/12081471/93022195-2a5a4c00-f5e8-11ea-899c-05007ad74d17.png)

Here's the code of my cache step:
```
      - name: Cache vcpkg dependencies
        uses: actions/cache@v2
        env:
          CACHE_UPLOAD_CHUNK_SIZE: 4194304
        with:
          path: ${{ env.VCPKG_ARCHIVE_DIR }}
          key: ${{ matrix.generator }}-${{ hashFiles('vcpkg.json') }}
          restore-keys: |
            ${{ matrix.generator }}-
```

dhadka:
Looks like that env var was removed when splitting the cache action into the cache module - https://github.com/actions/toolkit/commit/1413cd0e32562e29ebe9aa04d4b7176120890791#diff-f121c5607011c504c6934838327d2b8cR332

Created a PR to add this back as an input instead of an environment variable - https://github.com/actions/cache/pull/424

dhadka:
@johannesstricker `v2.1.2` of this action (which also include the `v2` tag) includes this fix.  Should work with:

```
- name: Limit upload chunk size
  uses: actions/cache@v2
  with:
    path: **/node_modules
    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
    upload-chunk-size: 4000000
```

johannesstricker:
This works now, thanks 👍

PR #424, opened 2020-10-02
Adds input for upload chunk size

dhadka:
Adds input to limit the upload chunk size.  This is useful for self-hosted runners with limited upload speeds, as an upload taking longer than the server timeout will fail.  We used to have this configurable via an environment variable, but that was removed when splitting the cache action out into the cache module.

```
- name: Limit upload chunk size
  uses: actions/cache@v2
  with:
    path: **/node_modules
    upload-chunk-size: 4000000
```

Fixes https://github.com/actions/cache/issues/415

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -12,7 +12,7 @@ inputs:
     description: 'An ordered list of keys to use for restoring the cache if no cache hit occurred for key'
     required: false
   upload-chunk-size:
-    description: 'The chunk size used when uploading large cache files'
+    description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
 outputs:
   cache-hit:
--- a/action.yml
+++ b/action.yml
@@ -11,6 +11,9 @@ inputs:
   restore-keys:
     description: 'An ordered list of keys to use for restoring the cache if no cache hit occurred for key'
     required: false
+  upload-chunk-size:
+    description: 'The chunk size used when uploading large cache files'
+    required: false
 outputs:
   cache-hit:
     description: 'A boolean value to indicate an exact match was found for the primary key'
--- a/src/save.ts
+++ b/src/save.ts
@@ -41,7 +41,9 @@ async function run(): Promise<void> {
         });
 
         try {
-            await cache.saveCache(cachePaths, primaryKey);
+            await cache.saveCache(cachePaths, primaryKey, {
+                uploadChunkSize: utils.getInputAsInt(Inputs.UploadChunkSize)
+            });
         } catch (error) {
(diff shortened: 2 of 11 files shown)


src/utils/actionUtils.ts
```diff
@@ -63,3 +63,14 @@ export function getInputAsArray(
         .map(s => s.trim())
         .filter(x => x !== "");
 }
+
+export function getInputAsInt(
```
joshmgross, line 67: Do we need to validate that the number isn't a decimal?

src/save.ts
```diff
@@ -41,7 +41,9 @@ async function run(): Promise<void> {
         });
 
         try {
-            await cache.saveCache(cachePaths, primaryKey);
+            await cache.saveCache(cachePaths, primaryKey, {
+                uploadChunkSize: utils.getInputAsInt(Inputs.UploadChunkSize)
```
joshmgross, line 45: `getInputAsInt` can return undefined, should we fail in that case or just pass in a default?

joshmgross (commented):
We'll need to document this as an optional input in the [`action.yml`](https://github.com/actions/cache/blob/main/action.yml#L4) file

src/utils/actionUtils.ts
dhadka: Yes, I switched to `parseInt` but for a different reason.  `Number('')` returns `0`.  `parseInt('')` returns `NaN`.

src/save.ts
dhadka: This would use the default (32 MBs).  This is handled by [`getUploadOptions`](https://github.com/actions/toolkit/blob/main/packages/cache/src/options.ts#L56).  If undefined it fills in the default value.

action.yml
```diff
@@ -11,6 +11,9 @@ inputs:
   restore-keys:
     description: 'An ordered list of keys to use for restoring the cache if no cache hit occurred for key'
     required: false
+  upload-chunk-size:
+    description: 'The chunk size used when uploading large cache files'
```
joshmgross, line 15: We should include the unit in either the description or input name (`upload-chunk-size-bytes`)

joshmgross (approved):

aiqiaoy (approved):
Looks good. 

Any particular reason we are adding this option as input vs. environment variable?

dhadka:
@aiqiaoy No particular reason.  If I had to give a reason, it would be for simplicity.  This puts all arguments under `with:` instead of splitting some out under `env:`.  In the end, inputs become environment variables anyway 😄

KTibow (approved):

PR #431, opened 2020-10-07
Add PowerShell to examples.md

potatoqualitee:
This PR adds PowerShell to the examples.md and includes all appropriate directories for each OS.

Working example can be seen here: https://github.com/potatoqualitee/tentools/runs/1222924184?check_suite_focus=true


examples.md
```diff
@@ -295,6 +296,30 @@ Esy allows you to export built dependencies and import pre-built dependencies.
...
+ - Windows: `~\Documents\PowerShell\Modules`
+ - macOS: `~/.local/share/powershell/Modules`
+
+```yaml
+- name: Get PowerShell Cache Directory
+  id: powershell-cache
+- uses: actions/cache@v2
+  with:
+    path: "~/.local/share/powershell/Modules"
+    key: ${{ runner.os }}-PS-SqlServer-PSScriptAnalyzer
```
joshmgross, line 314: Could we add a comment to make it clear that the cache key here should match the modules listed in the `Install-Module` step?

Phantsure: @joshmgross Is this a needed step?

Phantsure:
@potatoqualitee Can you please resolve the conflicts here?

Phantsure:
@potatoqualitee Can you kindly look at Josh's comment? Thanks

potatoqualitee:
Hello I am on vacation and will set this to unread and address when I get back next week 👍🏼👍🏼

examples.md
potatoqualitee: Back when I made this, I didn't entirely understand keys. I've since created a [GitHub Action](https://github.com/marketplace/actions/powershell-module-cache) to help others cache but I'd still like to keep this example. I believe this would potentially be a good key.

```
key: ${{ runner.os }}-${{ hashFiles("~/.cache/powershell/ModuleAnalysisCache-*") }}
```

Going to test it out and come back to update with the info

potatoqualitee: `ModuleAnalysisCache` takes at least 10 seconds to be built and doesn't exist in time to be reliability tested. I'm still searching for an ideal solution.

Phantsure: @potatoqualitee Are you still working on this?

Phantsure:
@potatoqualitee Friendly ping.

potatoqualitee:
thanks so much for the friendly ping, i will mark this notification as unread and return momentarily 🔜

potatoqualitee:
@LaurentDardenne contributed an incredible PR to my potatoqualitee/psmodulecache cacher and may have the solution for this. I'll take a look at their solution and report back.

potatoqualitee:
After some review, it seems too complex to make as an example. Anyone who wants to cache PowerShell files should use https://github.com/marketplace/actions/powershell-module-cache 😅 

Thank you all for your time and effort.

LaurentDardenne:
@joshmgross 
>>_Could we add a comment to make it clear that the cache key here should match the modules listed_

Hello,
the v5.0 documentation details the construction of the key.
The key is changed and will generate a new cache:
- if we add a new module name or remove one from the list,
- if an automatic update is requested and a new version exists.

The [current](https://github.com/potatoqualitee/psmodulecache#cache-key-construction-method) documentation.
The documentation of the [next evolution](https://github.com/LaurentDardenne/psmodulecache/blob/vNext5.0WithCredential/README.md#cache-key-construction-method).

I didn't state how 'actions/cache' works.

joshmgross:
Thanks @LaurentDardenne, I think this works better as a separate action than an example for `actions/cache`.

PR #433, opened 2020-10-13
Update Rust directories recommended for caching

aleksator:
This commit applies the suggestion from The Cargo Book:
https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -453,9 +453,11 @@ When dependencies are installed later in the workflow, we must specify the same 
 - uses: actions/cache@v2
   with:
     path: |
-      ~/.cargo/registry
-      ~/.cargo/git
-      target
+      ~/.cargo/bin/
+      ~/.cargo/registry/index/
+      ~/.cargo/registry/cache/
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -453,9 +453,11 @@ When dependencies are installed later in the workflow, we must specify the same
 - uses: actions/cache@v2
   with:
     path: |
-      ~/.cargo/registry
-      ~/.cargo/git
-      target
+      ~/.cargo/bin/
+      ~/.cargo/registry/index/
+      ~/.cargo/registry/cache/
+      ~/.cargo/git/db/
```
mzabaluev, line 459: The content of these three directories is actually OS-independent, so it's possible to cache it under a separate key that does not depend on `runner.os`.

There are some caveats, though, with the current cache implementation; see also #325 for more elaborate examples.

dhadka (approved):

PR #439, opened 2020-10-20
Document that no cache is saved on an exact match and thus file hashes should be included in keys

jwodder:
The (unexpected, to me) fact that the cache isn't updated on an exact match seems to be an important point worthy of being documented, and it helps to explain exactly *why* hash expressions should be included in keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 ### Inputs
 
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns. 
-* `key` - An explicit key for restoring and saving the cache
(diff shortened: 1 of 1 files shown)


Phantsure:
Exact match of the key means cache hit. When a cache hit occurs, cache wouldn't be updated as cache is already present there. When a cache miss occurs, cache will be updated for future use. 
It seems you are suggesting that users might not be familiar with concept of caching. Is that right?

jwodder:
@Phantsure I'm suggesting they might not be familiar with the "When a cache hit occurs, cache wouldn't be updated" aspect of caching, yes.

Phantsure:
cc @vsvipul For second par of eyes. Does this change makes sense here or should it be present at all?

Phantsure:
You are suggesting to call out two things:
- Cache is not saved on cache hit 
- Use hash of dependency files in key, so as to invalidate cache when files change.

@jwodder Is that right?

jwodder:
@Phantsure Yes.

Phantsure:
- cache not saved on hit, read below example workflow: https://github.com/actions/cache#example-workflow
- Use hashfiles: https://github.com/actions/cache#creating-a-cache-key

All seems resolved, So I am closing this PR. Thank you for your contribution @jwodder

PR #473, opened 2020-11-28
Sort inputs and remove whitespaces after exclamation

### LINKED ISSUE #345: Normalize path in version hash
As suggested by @mzabaluev in https://github.com/actions/cache/pull/343#pullrequestreview-423013155, the logic to calculate the version hash could consider other normalizations to the path, including:

1. Sorting the paths so the version hash is independent of the ordering
2. Removing additional whitespace in glob patterns, such as any space between ! and the path.

mondeja:
Fixes #345


mondeja:
I'm not sure why is prettier failing in Windows :thinking:

src/utils/actionUtils.ts
```diff
@@ -60,8 +60,9 @@ export function getInputAsArray(
     return core
         .getInput(name, options)
         .split("\n")
-        .map(s => s.trim())
-        .filter(x => x !== "");
+        .map(s => s.replace(/^\!\s+/, '!').trim())
```
vsvipul, line 63: Can you explain what does this line do and why is this required?

mondeja: All white spaces after an exclamation sign in lines starting with exclamation sign are removed. Is required for normalization of excluded paths.

vsvipul: @mondeja  Is this normalization really required? In most cases, I've seen people never use spaces after ! . A simple trim() should be good enough. @bishal-pdMSFT  thoughts?

vsvipul: Also, can you push a commit to rerun the workflows, I'm unable to see the old logs to check by windows build is failing.

mondeja: I'm not sure if it is really "required". It seems that can be declared with space but the hash will not be the same.

> A simple trim() should be good enough.

I don't follow you. This implementation is not trimming, `! foo` should be converted to `!foo` but `.trim()` does not modificates it (`'! foo'.trim() === '! foo'` is `true`).

vsvipul: I understand but what I'm saying is that people don't normally use `! foo`, they'll just use `!foo` directly, as mentioned in our docs.

vsvipul: Seems to be some prettier issue for windows. Can you run `npm run format` once to see if it produces any changes?

mondeja: `npm run format` does not produce changes in my Ubuntu, seems only failing in Windows. 

> I understand but what I'm saying is that people don't normally use ! foo, they'll just use !foo directly, as mentioned in our docs.

I haven't used the space in the past and don't have any preference about it. Just implemented here because it is possble to include the space after the exclamation sign using `@actions/glob` API directly:

```javascript
> const glob = require('@actions/glob');
> const globber = await glob.create(['node_modules/.bin/*', '! node_modules/.bin/mkdirp'].join('\n'))
> await globber.glob()
```

But is up to you to include it or not.

vsvipul:
@mondeja Unless the windows-latest test is fixed, I won't be able to merge this. Please see if you can fix it.

vsvipul:
@mondeja Yes, please regenerate dist/ as that will be required for merging. When we create a new release, users using actions/cache@v3 will be pointed to the new release with this code.

mondeja:
Regenerated, see the conflicts

vsvipul:
Hmm for dist/ files there shouldn't be conflicts. I'll take a look.

vsvipul:
@mondeja What version of npm are you using ? 
I checkout out your branch and regenerated dist/ . There were some changes -
```
@vsvipul ➜ /workspaces/cache (fix-345) $ npm run build

> cache@2.1.3 build /workspaces/cache
> tsc && ncc build -o dist/restore src/restore.ts && ncc build -o dist/save src/save.ts

ncc: Version 0.20.5
ncc: Compiling file index.js
ncc: Using typescript@3.9.10 (local user-provided)
2445kB  dist/restore/index.js
2445kB  [5775ms] - ncc 0.20.5
ncc: Version 0.20.5
ncc: Compiling file index.js
ncc: Using typescript@3.9.10 (local user-provided)
2445kB  dist/save/index.js
2445kB  [4773ms] - ncc 0.20.5
@vsvipul ➜ /workspaces/cache (fix-345 ✗) $ git diff
diff --git a/dist/restore/index.js b/dist/restore/index.js
index 1700b8e..f8f36f1 100644
--- a/dist/restore/index.js
+++ b/dist/restore/index.js
@@ -48,6 +48,25 @@ module.exports =
 
 "use strict";
 
+var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
+    if (k2 === undefined) k2 = k;
+    Object.defineProperty(o, k2, { enumerable: true, get: function() { return m[k]; } });
+}) : (function(o, m, k, k2) {
+    if (k2 === undefined) k2 = k;
+    o[k2] = m[k];
+}));
+var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (function(o, v) {
+    Object.defineProperty(o, "default", { enumerable: true, value: v });
+}) : function(o, v) {
+    o["default"] = v;
+});
```

mondeja:
Node v16 and NPM v8

vsvipul:
Hi @mondeja dist/ files should not be generated, I just tried on main with your changes. Can you try to update your main from actions/cache and then rebase this branch from that main.

mondeja:
Done. I've reverted to the change previous to regeneration and force pushed.

Trying to regenerate with `npm run build` I'm getting the next errors, so I've not regenerated this time:

<details>
<summary>Errors</summary>

```
src/restore.ts:55:25 - error TS1196: Catch clause variable cannot have a type annotation.

55         } catch (error: unknown) {
                           ~~~~~~~

src/restore.ts:64:21 - error TS1196: Catch clause variable cannot have a type annotation.

64     } catch (error: unknown) {
                       ~~~~~~~

src/save.ts:52:25 - error TS1196: Catch clause variable cannot have a type annotation.

52         } catch (error: unknown) {
                           ~~~~~~~

src/save.ts:62:21 - error TS1196: Catch clause variable cannot have a type annotation.

62     } catch (error: unknown) {
                       ~~~~~~~

src/utils/actionUtils.ts:81:16 - error TS2339: Property 'isFeatureAvailable' does not exist on type 'typeof import("/...path/to/.../cache/node_modules/@actions/cache/lib/cache")'.

81     if (!cache.isFeatureAvailable()) {
                  ~~~~~~~~~~~~~~~~~~


Found 5 errors.
```
</details>

vsvipul:
Apologies for the delay, I'll try to take a look this week and see if I can cherry pick and commit this change without the check dist/ check failing.

vsvipul:
Cherry-picked this commit and working on it in https://github.com/actions/cache/pull/903. So, closing this for now. Thank you the contribution. Will get merged soon. 🙌

Issue #334, opened 2020-05-28
Support for Read Only Cache / Configurable Cache Saving


bflad:
## Version

```
actions/cache@v2
```

## Description

We have a repository with multiple workflows that could benefit from shared caching across all of them. We would like to avoid trying to create and conditionalize one large workflow.

e.g. caching the download of all Go Modules:

```yaml
# Workflow 1
name: Code Checks

on:
  push:
    branches:
      - master
  pull_request:
    paths:
      - go.sum
      - src/**

jobs:
  go_mod_download:
    name: go mod download
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-go@v2
      with:
        go-version: "1.14"
    - uses: actions/cache@v2
      id: cache-go-pkg-mod
      with:
        path: ~/go/pkg/mod
        key: ${{ runner.os }}-go-pkg-mod-${{ hashFiles('go.sum') }}
    - if: steps.cache-go-pkg-mod.outputs.cache-hit != 'true' || steps.cache-go-pkg-mod.outcome == 'failure'
      run: go mod download
  # ... much more ...

# Workflow 2

name: CHANGELOG Checks
on:
  push:
    branches:
      - master
  pull_request:
    paths:
      - CHANGELOG.md

jobs:
  misspell:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-go@v2
        with:
          go-version: "1.14"
      - uses: actions/cache@v2
        with:
          path: ~/go/pkg/mod
          key: ${{ runner.os }}-go-pkg-mod-${{ hashFiles('go.sum') }}
      - run: go install github.com/client9/misspell/cmd/misspell
      - run: misspell -error -source text CHANGELOG.md
```

Currently there can be race conditions between these workflows, where the first one that completes will save the cache, while other jobs will see the post-action hit on the cache key and prevent overwriting it. If the latter workflow completes first, the Go Modules download would contain only the download of the Go Modules relevant to the installed package, rather than all Go Modules for the project.

As a simple workaround, we could run the `go mod download` in every workflow:

```yaml
jobs:
  misspell:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-go@v2
        with:
          go-version: "1.14"
      - uses: actions/cache@v2
        with:
          path: ~/go/pkg/mod
[...]

Issue #210, opened 2020-03-10
Make Cache Key Optional?


danrbailey:
Is it possible to make the key optional and hence make the cache action read-only? This is in the case that you know that you only want to read from one of the "restore" keys and to fail if you can't find one. 

For clarity, look for restore key patterns only: 

On cache hit - read the cache, don't attempt to write a new one.
On cache miss - job fail.

An alternative could be to recognize a specific string as the key which means _skip_ the write on job completion:

```
      uses: actions/cache@v1
      with:
        path: cache
        key: donotwrite
        restore-keys: myexpectedcache-
```

luigi311:
This would be nice. I have a use case where on job 1 I download a file off a server and then create a cache of said file with the key file-${{ hashFile(file) }}, on job 2 I need to do a restore on that exact key but the file does not exist in job 2 and I don't want to download the file every single time so I want to call a cache restore with the restore key of file- so it does a partial match and restores the lastest key. The problem with the current implementation is I need to generate a bogus unique key for every single run so I don't get a cache hit and restore a version that might not be valid but then I'm storing the same file over and over with all the bogus keys.

luigi311:
Preferred syntax would be 
```yml
- name: Restore file
  uses: actions/cache@v2
  with:
    path: file
    restore_keys: file-
```

MartijnHols:
You can achieve this using my fork https://github.com/MartijnHols/actions-cache:

```yaml
name: Build app

on: push

jobs:
  build:
    needs: [install]
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Restore "node_modules" from cache
      uses: martijnhols/actions-cache/restore@main
      with:
        path: node_modules
        # key is a required field, by providing a dummy value we pass validation while it won't match a cache entry and therefore be ignored
        key: dummykey
        restore-keys: myexpectedcache-
        # Fail when the cache could not be found
        required: true

    - name: Build app
      run: yarn build
```

danrbailey:
@MartijnHols - very cool, thank you for sharing. I'll give it a try.

PR #474, opened 2020-11-29
Add read-only feature

### LINKED ISSUE #350: Read only cache?

MartijnHols:
When `read-only` is `true`, the cache is only restored and not saved. This allows for sharing the cache with multiple steps even if these steps may change them. I tried to use `inputs.read-only` in the `post-if` since just starting the post-job takes precious seconds, but that always seems to be `null`.

Available as `martijnhols/cache@read-only`

Fixes #350, #334 and maybe #210

### DIFF
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -27,6 +27,13 @@ beforeAll(() => {
             return actualUtils.getInputAsArray(name, options);
         }
     );
+
+    jest.spyOn(actionUtils, "getInputAsBoolean").mockImplementation(
+        (name, options) => {
+            const actualUtils = jest.requireActual("../src/utils/actionUtils");
+            return actualUtils.getInputAsBoolean(name, options);
+        }
+    );
 });
 
 beforeEach(() => {
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -308,3 +308,35 @@ test("restore with cache found for restore key", async () => {
     );
     expect(failedMock).toHaveBeenCalledTimes(0);
 });
+
+test("restore with read-only with cache found for key", async () => {
+    const path = "node_modules";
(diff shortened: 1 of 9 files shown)


MartijnHols:
Does this PR have a future or is it best closed?

MartijnHols:
Seeing as my PRs to the GitHub repos have been ignored for almost a year, I closed all of them. To clean up my workflows I have decided to work on a fork instead with an (in my eyes) cleaner solution. See the repository for documentation: https://github.com/MartijnHols/actions-cache Here is what sharing cache across jobs would look like:

```yaml
name: Build app

on: push

jobs:
  install:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Restore "node_modules" from cache
      id: cache
      uses: martijnhols/actions-cache/restore@main
      with:
        path: node_modules
        key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock', 'patches') }}
        restore-keys: ${{ runner.os }}-node_modules

    - name: Install dependencies
      run: yarn install

    - name: Save "node_modules" to cache
      uses: martijnhols/actions-cache/save@main
      with:
        path: node_modules
        key: ${{ steps.cache.outputs.primary-key }}

  build:
    needs: [install]
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Restore "node_modules" from cache
      uses: martijnhols/actions-cache/restore@main
      with:
        path: node_modules
        key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock', 'patches') }}
        # Fail when the cache could not be found (this should never happen unless you have a misconfiguration)
        required: true

    - name: Build app
      run: yarn build
```

kotewar:
Hey all, 👋🏽 

PR #489, opened 2020-12-14
Allow to skip the save post-step

### LINKED ISSUE #334: Support for Read Only Cache / Configurable Cache Saving

### LINKED ISSUE #350: Read only cache?

roni-frantchi:
This is similar to #474 - but improves on that by skipping the `post` (save) step entirely, saving tens of seconds on each run.  

Fixes #350 
Fixes #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,6 @@ jobs:
       run: /generate-primes.sh -d prime-numbers
 
     - name: Use Prime Numbers
-      env:
-        CACHE_SKIP_SAVE: true
       run: /primes.sh -d prime-numbers
 ```
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 * `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key
 
 ### Environment Variables
-* `CACHE_SKIP_SAVE` - [optinal] When set on step, any modifications made to the restored cache will not be persisted back at the end of the step.  
+* `CACHE_SKIP_SAVE` - [optional] When set to `true`, any modifications made to the restored cache will not be persisted back at the end of the step.  This environment variable can be set at any time using `echo "CACHE_SKIP_SAVE=true" >> $GITHUB_ENV`
 
 ### Outputs
 
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Cache Primes
+      env:
+        CACHE_SKIP_SAVE: true
       id: cache-primes
       uses: actions/cache@v2
       with:
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 * `key` - An explicit key for restoring and saving the cache
 * `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key
 
+### Environment Variables
+* `CACHE_SKIP_SAVE` - [optinal] When set on step, any modifications made to the restored cache will not be persisted back at the end of the step.  
+
 ### Outputs
 
 * `cache-hit` - A boolean value to indicate an exact match was found for the key
@@ -78,6 +81,8 @@ jobs:
       run: /generate-primes.sh -d prime-numbers
 
     - name: Use Prime Numbers
+      env:
+        CACHE_SKIP_SAVE: true
       run: /primes.sh -d prime-numbers
 ```
 
--- a/action.yml
+++ b/action.yml
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success()'
+  post-if: 'success() && !env.CACHE_SKIP_SAVE'
 branding:
   icon: 'archive'
   color: 'gray-dark'
--- a/action.yml
+++ b/action.yml
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success() && env.CACHE_SKIP_SAVE !== "true"'
+  post-if: 'success() && env.CACHE_SKIP_SAVE != "true"'
 branding:
   icon: 'archive'
   color: 'gray-dark'
--- a/action.yml
+++ b/action.yml
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success() && env.CACHE_SKIP_SAVE != "true"'
+  post-if: success() && env.CACHE_SKIP_SAVE != 'true'
 branding:
   icon: 'archive'
   color: 'gray-dark'
--- a/action.yml
+++ b/action.yml
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success() && !env.CACHE_SKIP_SAVE'
+  post-if: 'success() && env.CACHE_SKIP_SAVE !== "true"'
 branding:
   icon: 'archive'
   color: 'gray-dark'


roni-frantchi:
Thanks @domjtalbot.  
Sure thing.  
If we get a positive response from maintainers on pushing it in, I'd be happy to make that adjustment along with any others suggested.

README.md
```diff
@@ -78,6 +81,8 @@ jobs:
       run: /generate-primes.sh -d prime-numbers
 
     - name: Use Prime Numbers
+      env:
```
joshmgross, line 84: This would need to be an `env` variable on the `cache-primes` step

action.yml
```diff
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success()'
+  post-if: 'success() && !env.CACHE_SKIP_SAVE'
```
joshmgross, line 24: Unfortunately this doesn't work well, as `env` values are always strings.

So for example both `CACHE_SKIP_SAVE: true` and `CACHE_SKIP_SAVE: false` would evaluate to a non-empty string ('true' and 'false`) which equal `true`.

See the debug logging in https://github.com/joshmgross/actions-testing/runs/1560134417?check_suite_focus=true

joshmgross (commented):
Thanks for your suggestion!

Due to the type issues with `env` always being a string, I think we'd need this to be an input rather than an environment variable so that we can properly evaluate it.

Also this does not make the cache key option, so could you update your PR description to remove "Fixes #210"?

action.yml
roni-frantchi: Right!  
Thanks.  

How about that then -https://github.com/actions/cache/pull/489/commits/f4a5fe8db9227c95e5652eec8e065832f1b4b4d9

README.md
roni-frantchi: https://github.com/actions/cache/pull/489/commits/6ceddd55d128082f23eaf7026dfd38e5f14baa41

roni-frantchi (commented):
Thanks for reviewing @joshmgross !  

> I think we'd need this to be an input rather than an environment variable so that we can properly evaluate it

`post-if` is undocumented and despite my best efforts to make it evaluate inputs I couldn't - only `env` was available to me.  

> Due to the type issues with env always being a string...

I did address this one though - so perhaps using `env` would be acceptable now?

anders-kiaer:
I see this PR has gone through a first review by @joshmgross already. Will it be merged in the near future, or should use cases needing to skip the save step use e.g. a fork of this action?

Regarding using environment variable: I think this is a good choice, if my understanding is correct that previous steps in the job then can dynamically/during runtime instruct the `cache` action save/post step through https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable.

Thanks :slightly_smiling_face:

mikeage:
This looks wonderful; one question, if you don't mind. I have a very large cache (~750 MB), and when I have multiple PRs in progress, it appears that the 5GB cache gets filled up with those, and the cache from `main` gets pushed out. If I were to use this fork and set the option to skip the saving on the PRs but not on the main branch, this would keep the main cache as a base for the all of the PRs, right?

zwass:
@joshmgross Can you please re-review/merge this? I'd like to use it and would prefer not to rely on a fork.

joshmgross:
👋  Hey @zwass, the @actions/artifacts-actions team should be able to help out here.

README.md
```diff
@@ -41,6 +41,9 @@ Create a workflow `.yml` file in your repositories `.github/workflows` directory
 * `key` - An explicit key for restoring and saving the cache
 * `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key
 
+### Environment Variables
+* `CACHE_SKIP_SAVE` - [optinal] When set on step, any modifications made to the restored cache will not be persisted back at the end of the step.
```
dhadka, line 45:
```suggestion
* `CACHE_SKIP_SAVE` - [optional] When set to `true`, any modifications made to the restored cache will not be persisted back at the end of the step.  This environment variable can be set at any time using `echo "CACHE_SKIP_SAVE=true" >> $GITHUB_ENV`
```

README.md
```diff
@@ -67,6 +70,8 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Cache Primes
+      env:
+        CACHE_SKIP_SAVE: true
```
dhadka, line 74: Given that this is the default example and skipping cache save is a special case, I wouldn't include this here.  Perhaps move this into the Examples.md file and we can add some specific examples of where this is useful.

README.md
```diff
@@ -78,6 +83,8 @@ jobs:
       run: /generate-primes.sh -d prime-numbers
 
     - name: Use Prime Numbers
+      env:
+        CACHE_SKIP_SAVE: true
```
dhadka, line 87: This should be removed now that it's set on the `cache-primes` step.

README.md
dhadka: Something we can also highlight in the example is how users can either set this on the cache step as shown here or using `echo "CACHE_SKIP_SAVE=true" >> $GITHUB_ENV`.

dhadka (changes requested):
TY for proposing this change.  Just some minor comments on the README changes, but otherwise looks good.  Please ping me when ready and I can 🚢 this.

README.md
roni-frantchi, line 87:
```suggestion
```

README.md
roni-frantchi: https://github.com/actions/cache/pull/489/commits/91185bb10f818be7d9890148bf120e8c89440cc0

README.md
roni-frantchi: I think the changes you've suggested on the readme file capture it well enough

README.md
roni-frantchi: https://github.com/actions/cache/pull/489/commits/64daede5552c68991cba51f3bc0ac2bc26945a11

ollydev:
@dhadka This is what I wanted #498 for, didn't see this PR though.

Right now cache will never be saved if the job fails.
Example: Spend 10mins building dependencies (succesfully) but a future job (such as failing a test) fails. Cache will not save!

Can we add something like:
```yml
env:  
  CACHE_FORCE_SAVE: true
```
```yml
`post-if: success() && env.CACHE_SKIP_SAVE != 'true' || failed() && env.CACHE_FORCE_SAVE!= 'true'`
```

zwass:
Perhaps @ollydev's request could be merged into this one with a single flag?

`CACHE_SAVE=always` - Always save regardless of errors in preceding steps.
`CACHE_SAVE=never` - Never save regardless of errors.
Not set - Default behavior (as before this PR).

jmoguillansky-gpsw:
failed() && env.CACHE_FORCE_SAVE == 'true': how cached content was generated successfully?  I think we need some mechanism to indicate that generating the cache content was successful, and then cache it immediately, why wait until the end of the workflow?

bgianfo:
>  Please ping me when ready and I can 🚢 this.

It looks like all the comments have been addressed?

bgianfo:
> Please ping me when ready and I can 🚢 this.

@dhadka Looks like all the comments have been addressed, can we get this merged?

dhadka:
@bgianfo There's a discussion happening over in https://github.com/actions/cache/pull/571 on a similar issue.  Essentially, there are a few open issues that are somewhat related (better control over if and when the cache is saved).  Instead of adding in a bunch of piecemeal fixes, is there a more generalizable solution?

> failed() && env.CACHE_FORCE_SAVE == 'true': how do we know the content to cache was generated successfully? I think we need some mechanism to indicate that generating the cache content was successful, and then cache it immediately, why wait until the end of the workflow?

@jmoguillansky-gpsw Yes, this is essentially a limitation of running save as a "post step", which only runs at the end of the workflow.  You're essentially looking for a standalone save step that a user can run at any point in their workflow.  See https://github.com/actions/cache/pull/571#issuecomment-839123699.

Zoramite:
How does this work when you are using multiple cache actions in one workflow job?

I have a cache action for the caching node modules and I have another cache action for a benchmark build results.

If I understand this correctly, when I `echo "CACHE_SKIP_SAVE=true" >> $GITHUB_ENV` then it would prevent both the cache for the node modules **and** the cache for the benchmark to not save after the workflow is completed. Is that correct?  I really want to target one of the cache actions to not save (ex: don't save the benchmark cache unless the branch is `main`), not just blindly affect all of the cache saves.

It seems like there should be a better way for specifically preventing a single cache action from saving without **all** cache actions being prevented from caching?

neerfri:
@dakale @roni-frantchi Seems like this change is of interest to many users.
As we are not bringing this PR to conclusion downstream users are turning to use forks.
Would probably be a good idea to bring this to conclusion

Thanks!

roni-frantchi:
@dakale I've addressed all of you comments - is there anything other request in this thread that you think must be taken care of in this PR before merging?  

I would rather add any improvements later on and get this one going.

README.md
rene-leanix: @roni-frantchi & @dhadka, would be great if this PR could be merged to reduce the workflow runtime when a cache-update isn't required. Seems like only a small adjustment would be needed?
Not sure if [this issue](https://github.com/actions/cache/issues/597) could be addressed in this PR too: skipping the post-step altogether when there was a cache-hit.

vinaysshenoy:
Is this going to get merged soon? The ability to skip the save post-step is something that will be very useful for our use case as well.

README.md
roni-frantchi: > Not sure if this issue could be addressed in this PR too

I'm not sure that it can either...  
Seeing so many are expecting it, maybe we should explore doing it on another PR?..  

> Seems like only a small adjustment would be needed?

Sorry @rene-leanix it's been a while - what adjustment would that be?..

action.yml
```diff
@@ -21,7 +21,7 @@ runs:
   using: 'node12'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success()'
+  post-if: success() && env.CACHE_SKIP_SAVE != 'true'
```
JCMais, line 24: My only suggestion would be to have a more specific environment variable here.

Something like `GH_ACTIONS_CACHE_SKIP_SAVE`

README.md
rene-leanix: @roni-frantchi, I was just referring to @dhadka's comments above in this thread: [here](https://github.com/actions/cache/pull/489/files#r616281599) and [here](https://github.com/actions/cache/pull/489/files#r616282355).

Sytten:
Any hope to get this merged

iuioiua:
@roni-frantchi, is it possible for you to update the branch? It'd remove the deprecation warnings around save-state, set-output and Node.js 12 for those using the fork.

Jimimaku:
> Dies ist ähnlich wie [# 474](https://github.com/actions/cache/pull/474) - verbessert dies jedoch, indem der (gespeicherte) Schritt vollständig übersprungen wird, wodurch bei jedem Lauf Dutzende von Sekunden gespart werden.`post`
> 
> Behebt #350 Behebt #334

kotewar:
Hey all, 👋🏽 

kotewar:
Hey everyone 👋🏽

Two new actions [actions/cache/restore](https://github.com/actions/cache/tree/main/restore) and [actions/cache/save](https://github.com/actions/cache/tree/main/save) are now available with tag `v3` to everyone for use. These can now be used to achieve granular control on the restore and save steps in the cache action. 

Do try them and let us know your feedback. We hope these new actions will take care of your use cases. 🙇🏽 

Closing this PR now as we believe the new actions will take care of the same, feel free to reopen it or create an issue if need be.

PR #495, opened 2020-12-29
Add sample cache for Erlang with Rebar3

ostera:


examples.md
```diff
@@ -104,6 +105,18 @@ steps:
       ${{ runner.os }}-mix-
 ```
 
+## Erlang - Rebar3
+```yaml
+- uses: actions/cache@v2
+  with:
+    path: |
+      ~/.cache/rebar3
+      _build
```
Phantsure, line 114: Can you provide a reference somewhere, where you see this path pattern?

ostera: Sure thing :) `_build` is the default build folder, as seen here: https://rebar3.readme.io/docs/configuration

And in the smae page you can find the default global rebar3 folder set to `~/.cache/rebar3`.

ostera:
Rebased on top of `main` 🆗

Phantsure (approved):

PR #514, opened 2021-01-28
Recommend ruby/setup-ruby's bundler-cache: true option

eregon:
Manually caching gems has many issues:
* Not working if there is no Gemfile.lock but only a Gemfile
* Not having the OS version in the key, which might cause binary incompatibility with system libraries of different ABI version.
* Not taking the Ruby version in account.
* Not taking the Ruby ABI version of development builds into accounts, which cannot be done with a key, but needs the commit hash.
* Using `restore-keys` would grow the cache over time and have extra gems in the cache.
* Those reasons are summarized in https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby#caching-bundle-install-manually

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -468,21 +468,18 @@ Replace `~/.local/share/renv` with the correct `path` if not using Ubuntu.
 
 ## Ruby - Bundler
 
-```yaml
-- uses: actions/cache@v2
-  with:
-    path: vendor/bundle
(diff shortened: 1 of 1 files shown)


konradpabjan (approved):
Agree with dropping the example and recommending `bundler-cache: true` 👍

dhadka:
👍 Awesome!

PR #525, opened 2021-02-03
Use @actions/cache version 1.0.6

konradpabjan:
Includes changes from the following two PRs:

https://github.com/actions/toolkit/pull/650
https://github.com/actions/toolkit/pull/701

I also updated all files under `.licenses/*`. They must have not been updated in a while so I'm fixing that as well. I used the `license cache` command: https://github.com/github/licensed#usage


dhadka (approved):

konradpabjan:
Merging this in right now! Will wait until tomorrow before updating any tags

poiru:
The `v2` tag does not include this, it still points to `v2.1.3`. Can we update `v2` to point to `v2.1.4`? cc @joshmgross

dhadka:
@poiru There is an issue with `v2.1.4` breaking existing Mac workflows - https://github.com/actions/cache/issues/527

PR #530, opened 2021-02-06
Make it more obvious that the cache call does double duty

axelson:
I wasted 20+ minutes because I didn't realize this and couldn't find the information in the documentation anywhere. Hopefully adding it to the Readme will help someone else avoid that pain in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ jobs:
       run: /primes.sh -d prime-numbers
 ```
 
+Note: The call to cache both restores the cache and sets up a hook to restore the cache at the end of the workflow so you need to add the cache step near the beginning of your workflow.
+
 ## Implementation Examples
 
 Every programming language and framework has its own way of caching.
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ jobs:
       run: /primes.sh -d prime-numbers
 ```
 
-Note: The call to cache both restores the cache and sets up a hook to restore the cache at the end of the workflow so you need to add the cache step near the beginning of your workflow.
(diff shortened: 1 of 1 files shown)


ludwigschubert:
Same here, but I'd suggest a slight rewording of your note:
> sets up a hook to restore the
should probably read sth like
> sets up a hook to post (upload) the new version of the cache

I'm including "post" as it seems to be the terminology Github has chosen to use for this.

Thanks for opening a PR, though; I feel this is a simple but important improvement.

vsvipul:
Our example at https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows should be good enough to make this obvious that the cache needs to be restored before the cached files are required.

Adding @lucascosti to take a look and if a change is required, we should probably add that to the above docs page.

README.md
```diff
@@ -81,6 +81,8 @@ jobs:
       run: /primes.sh -d prime-numbers
 ```
 
+Note: The call to cache both restores the cache and sets up a hook to restore the cache at the end of the workflow so you need to add the cache step near the beginning of your workflow.
```
lucascosti, line 84: @vsvipul Here's a suggestion:

```suggestion
> Note: You must use the `cache` action in your workflow before you need to use the files that might be restored from the cache. If the provided `key` doesn't match an existing cache, a new cache is automatically created if the job completes successfully.
```

vsvipul (approved):
Thank you @axelson and @lucascosti for adding this.

README.md
axelson: Thanks for the updated wording :heart:

PR #537, opened 2021-02-17
It is not recommended to cache node_modules

robstoll:
According to the docs of this repo, so the example should not show it


joshmgross (approved):

ClementParis016:
@robstoll can you point towards some reference about where it's not recommended to cache `node_modules`?

robstoll:
https://github.com/actions/cache/blob/main/examples.md#node---npm

PR #554, opened 2021-03-25
Improve restore-key documentation

albertstill:


Phantsure (approved):
This seems like a good change

PR #567, opened 2021-04-17
Fix typo of cache path in Lerna example

yuichkun:
Hello, there!

I think I found a typo in the Lerna example.

If my understanding is correct, you want to recursively search for `node_modules` in child directories, so `**` is the one to use, no 🤔  ?

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -215,7 +215,7 @@ If using `npm config` to retrieve the cache directory, ensure you run [actions/s
 - name: restore lerna
   uses: actions/cache@v2
   with:
-    path: **/*/node_modules
+    path: **/node_modules
     key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
 ```
 
--- a/examples.md
+++ b/examples.md
@@ -215,9 +215,7 @@ If using `npm config` to retrieve the cache directory, ensure you run [actions/s
 - name: restore lerna
   uses: actions/cache@v2
   with:
-    path: |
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -217,7 +217,7 @@ If using `npm config` to retrieve the cache directory, ensure you run [actions/s
   with:
     path: |
       node_modules
-      */*/node_modules
+      **/*/node_modules
```
dhadka, line 220: Can this be simplified to just `path: **/node_modules`?

If I'm interpreting this correctly, the old version would match a top-level `node_modules` and all `node_modules` nested two folders deep.  The proposal here is to match all `node_modules`.

dhadka (dismissed):

examples.md
yuichkun: Oh, you are totally right! fixed it so 👍

vsvipul:
@yuichkun Your change still has `**/*/node_modules` instead of `**/node_modules`. Is this intentional ?

yuichkun:
> @yuichkun Your change still has `**/*/node_modules` instead of `**/node_modules`. Is this intentional ?

@vsvipul 
Oh my goodness, I'd want to dig a hole to put myself in.
No, it wasn't intentional, and fixed it again.
Sorry for the mess.

Do you guys squash commits when merging or should I just fix up commits&force-push?

vsvipul (approved):
LGTM!

PR #577, opened 2021-05-04
Golang example tweak - add `go-build` path - rebuild page TOC

magnetikonline:
A few small things:

- Noted in the Golang cache example, worthwhile to include the `~/.cache/go-build` path (build cache) in addition to the module cache.
- Fixed up some whitespace while in the area.

Finally, noted the TOC was a little out of sync - so fixed it up with (warning, self plug!) my little [Markdown TOC generate](https://magnetikonline.github.io/markdown-toc-generate/) utility.


examples.md
```diff
@@ -114,7 +118,9 @@ steps:
 ```yaml
 - uses: actions/cache@v2
   with:
-    path: ~/go/pkg/mod
+    path: |
+      ~/.cache/go-build
```
dhadka, line 122: I'm not familiar with Golang, but while googling this I came across [this cache example](https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds) that shows different paths for different OS'es.  That link suggests this only works on Linux, correct?  Is there an env var or command that can be used to get the build cache path dynamically?

dhadka (changes requested):
> Finally, noted the TOC was a little out of sync - so fixed it up with (warning, self plug!) my little Markdown TOC generate utility.

Hey, that's an awesome utility!  Turn it into a GitHub Action so it can check TOCs in PRs! 😜 

Just one comment/question about the path on different OS'es.  Thanks for submitting this PR!

examples.md
magnetikonline: 🤦 

Oh @dhadka you're on the money 👍  - this is why we have +1's 😄 .

I'll update the PR now with a cross-os way of doing this 👍 . Sorry, I spend my life in Linux build agents typically.

magnetikonline:
> Hey, that's an awesome utility! Turn it into a GitHub Action so it can check TOCs before merging PRs!

Thanks 👍  - something I might have to consider doing.

magnetikonline:
Thanks @dhadka - split out the examples into Linux/macOS/Windows versions in 9c0250e.

I was thinking, the alternative is you could call `go env GOCACHE` in a prior step - make that an output variable and reference it in the cache path parameter - that would make the workflow truly cross platform - but seems more effort than it's worth.

magnetikonline:
Small bump @dhadka (seems I can't re-request a review).

dhadka (approved):
TY for making those changes!

magnetikonline:
Thanks @dhadka 👍

Issue #454, opened 2020-11-08
Access to .gradle/caches on windows-latest 


JakubMosakowski:
Hi!

Cache actions seem not to be working with Gradle on the windows machine. Tar doesn't have an access to some files.

```
Post job cleanup.
C:\windows\System32\tar.exe --posix -z -cf cache.tgz -P -C D:/a/ItemistEvolved/ItemistEvolved --files-from manifest.txt
tar.exe: Couldn't open C:/Users/runneradmin/.gradle/caches/transforms-2/transforms-2.lock: Permission denied
tar.exe: Error exit delayed from previous errors.
Warning: Tar failed with error: The process 'C:\windows\System32\tar.exe' failed with exit code 1
```

```
name: Pull Request Workflow

on:
  pull_request:
    branches: [ develop ]

jobs:  
  build:
    runs-on: windows-latest
    steps:
      - name: Checkout the code
        uses: actions/checkout@v2
      - name: Get gradle cache
        uses: actions/cache@v2
        with:
          path: |
            ~/.gradle/caches
            ~/.gradle/wrapper
          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
          restore-keys: |
            ${{ runner.os }}-gradle-

...
```

JavierSegoviaCordoba:
The problem is not only the cache doesn't work, but the cache post step also takes 4 minutes until it stops with that log...

Danielku15:
I also wanted to activate the caching of some Gradle stuff in my repo and instantly ran into this issue. The issue is caused by the Gradle Daemon which keeps running. I worked around the issue in my CI scrips via: 

1. Adding `--no-daemon` to my gradlew calls. 
2. Adding an additional `gradlew --stop` step at the end of my workflow to stop any potentially running daemon which spawned without my knowledge.

Might be worth noting this in the readme for Gradle that devs need to ensure no gradle daemon is running anymore because it might make the task fail due to locked files.

Hope with this workaround you're also able to solve your build issues 😉

dhadka:
@Danielku15 Thanks, I've merged your PR.

@JakubMosakowski If this suggestion resoles the issue, please close or let us know if there's any other repair items.

JavierSegoviaCordoba:
When is this going to be properly fixed?

PR #588, opened 2021-05-13
Extend Gradle example with note to stop any running Daemons

Danielku15:
This PR adds a small note to the Gradle example mentioning that devs should ensure that no Daemons are running anymore. Running Daemons can hold locks on the file system which makes the `tar.exe` fail to create the package. See #454 for details. 

**Daemons holding locks cause this error**
image

**Ensuring Daemons are stopped**
image


dhadka (approved):
Thank you, LGTM!

Issue #154, opened 2020-01-17
Error: EBADF: bad file descriptor, read


myyra:
When trying to cache the Xcode DerivedData directory on macOS (`~/Library/Developer/Xcode/DerivedData`), I get the this error:

```
[warning]write EPIPE
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: EBADF: bad file descriptor, read
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5) {
  errno: -9,
  code: 'EBADF',
  syscall: 'read'
}
##[error]Node run failed with exit code 1
```

Only additional info I can add is that the directory contains a lot of files (`ls **` won't run almost anywhere inside)

josharian:
I have also experienced this using the standard Go modules cache example. The error was transient (it did not occur next run), but something else seems to have been broken by it, since all attempts to cache afterwards have failed.

Pintouch:
Also experiencing the same thing, we got the same thing using it to cache `yarn` cache folder on a recent Action build, we didn't change anything in our workflow, here is the output:

```
Post job cleanup.
/bin/tar -cz -f /home/runner/work/_temp/06b9cad5-a82c-4ae4-aae8-89bdb7b69f24/cache.tgz -C /home/runner/.cache/yarn/v6 .
[warning]Cache service responded with 404 during chunk upload.
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: EBADF: bad file descriptor, read
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5) {
  errno: -9,
  code: 'EBADF',
  syscall: 'read'
}
##[error]Node run failed with exit code 1
```

dhadka:
Merged #596 to fix this issue by catching and logging warnings.  The save step should no longer fail.

Issue #467, opened 2020-11-21
Saving cache can fail, breaking build (response 429/503)


vadi2:
```
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /Users/runner/work/Mudlet/Mudlet --files-from manifest.txt
Warning: Cache service responded with 429 during upload chunk.
/Users/runner/work/_actions/actions/cache/v2/dist/save/index.js:3305
                        throw new Error(`Cache upload failed because file read failed with ${error.message}`);
                        ^

Error: Cache upload failed because file read failed with EBADF: bad file descriptor, read
    at ReadStream.<anonymous> (/Users/runner/work/_actions/actions/cache/v2/dist/save/index.js:3305:31)
    at ReadStream.emit (events.js:210:5)
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5)
```

See https://github.com/Mudlet/Mudlet/runs/1435372417?check_suite_focus=true

Cache itself is rather simple,

```
    - name: (Linux/macOS) restore ccache
      uses: actions/cache@v2
      with:
        path: ${{runner.workspace}}/ccache
        key: ccache-${{matrix.os}}-${{matrix.compiler}}-${{matrix.qt}}-${{github.sha}}
        restore-keys: ccache-${{matrix.os}}-${{matrix.compiler}}-${{matrix.qt}}-
      if: runner.os != 'Windows'
```

vadi2:
Sometimes a `503` can also be returned:

```
Post job cleanup.
/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/Mudlet/Mudlet --files-from manifest.txt
Warning: uploadChunk (start: 637534208, end: 671088639) failed: Cache service responded with 503
/home/runner/work/_actions/actions/cache/v2/dist/save/index.js:3305
                        throw new Error(`Cache upload failed because file read failed with ${error.message}`);
                        ^

Error: Cache upload failed because file read failed with EBADF: bad file descriptor, read
    at ReadStream.<anonymous> (/home/runner/work/_actions/actions/cache/v2/dist/save/index.js:3305:31)
    at ReadStream.emit (events.js:210:5)
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5)
```

pakatagoh:
[...]

Issue #441, opened 2020-10-20
Errors during `uploadChunk` leading to uncaught errors


dhadka:
Errors thrown during `uploadChunk` are leading to a variety of uncaught errors, including:

* `Error: Cache upload failed becaues the file read failed with EBADF: bad file descriptor, read`
* `Error: Cache upload failed because file read failed with ESPIPE: invalid seek, read`

These errors cause the cache step to fail, whereas we expect caching to fail gracefully and not break a workflow.

I discuss a hypothesis for why this happens in https://github.com/actions/cache/issues/399#issuecomment-675051780, namely that:

1. Some error happens during the call to `uploadChunk`, in this case it was a 503 response
2. That error causes the finally block to execute and close the file descriptor
3. Closing the file descriptor kills the other in-progress calls to uploadChunk, resulting in the `Error: Cache upload failed becaues the file read failed with EBADF: bad file descriptor, read`

dhadka:
Merged #596 to fix this issue by catching and logging warnings.  The save step should no longer fail.

PR #596, opened 2021-05-27
Catch and log unhandled exceptions stemming from closed file descriptor

dhadka:
Due to how `uploadChunk` is implemented, an error while uploading a chunk causes `Promise.all` to reject with that error.  This then closes the file descriptor inside the `finally` block.  However, if there are any async uploads still running, closing the file descriptor may cause I/O operations to fail with `bad file descriptor` or similar errors.  But most importantly, these errors are uncaught, even though this is all run inside try-catch blocks.

Fixes https://github.com/actions/cache/issues/154, https://github.com/actions/cache/issues/441, https://github.com/actions/cache/issues/467, and https://github.com/actions/toolkit/issues/612

This solution is a bit heavy handed as it just converts all unhandled exceptions to warnings.  I also thought about fixing `uploadChunk` so we don't close the file descriptor until all upload promises complete, but this may cause slowness.  Closing the file descriptor kills all uploads and lets the action terminate quickly.

### Repro

The following script can reliably reproduce this issue.  It essentially create a large file that will take some time to read, then creates two promises: one reads the file and the second causes an error (rejects the promise).  The error closes the file descriptor which interrupts the read, producing a "bad file descriptor" error.

```
import * as fs from 'fs'

process.on('uncaughtException', e => console.log(`${Date.now()} - uncaught exception: ${e}`))
process.on('unhandledRejection', e => console.log(`${Date.now()} - unhandled rejection: ${e}`))

async function run(): Promise<void> {
    fs.writeFileSync("foo.txt", "z".repeat(300000000))

    console.log(`${Date.now()} - opening fd`)
    let fd = fs.openSync("foo.txt", "r")

    try {
        await Promise.all([
            new Promise((resolve, reject) => {
                console.log(`${Date.now()} - start reading`)
                fs.readFile(fd, (e, data) => {
                    console.log(`${Date.now()} - read file callback`)
                    if (e) throw Error(`Rethrowing file error - ${e}`)
                    else resolve("success!")
                })
            }),
            new Promise((resolve, reject) => {
                console.log(`${Date.now()} - rejecting`)
                reject(new Error("something went wrong"))
            })
        ])
    } catch (error) {
        console.log(`${Date.now()} - catching ${error}`)
    } finally {
        console.log(`${Date.now()} - closing fd`)
        fs.closeSync(fd)
    }

    // Keep alive until readFile can fail...
    await new Promise((resolve, reject) => {
        setTimeout(() => resolve("Done!"), 1000)
    })
}

run().catch(e => console.log(`${Date.now()} - post run catch ${e}`))
```

It doesn't matter if we have try-catch blocks or using `run().catch(...)`, the bad file descriptor error still makes its way out and is only caught by the `unhandledException` event:

```
> tsc repro.ts && node repro.js

1622076841093 - opening fd
1622076841109 - start reading
1622076841111 - rejecting
1622076841115 - catching Error: something went wrong
1622076841115 - closing fd
1622076841119 - read file callback
1622076841120 - uncaught exception: Error: Rethrowing file error - Error: EBADF: bad file descriptor, read
```


konradpabjan (approved):

777666007 (commented):
> 

> #675

PR #604, opened 2021-06-25
Create check-dist.yml

brcrista:
This adds a workflow to check the contents of the checked-in `dist/index.js` against the expected version.

See https://github.com/actions/upload-artifact/pull/227 for background and rationale.

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -1,0 +1,50 @@
+# `dist/index.js` is a special file in Actions.
+# When you reference an action with `uses:` in a workflow,
+# `index.js` is the code that will run.
+# For our project, we generate this file through a build process
+# from other source files.
+# We need to make sure the checked-in `index.js` actually matches what we expect it to be.
+name: Check dist/
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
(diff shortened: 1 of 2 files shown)


konradpabjan (approved):

konradpabjan:
Certain first party actions have `pre` and `post` steps so there are 2 `index.js` files

This will need some tweaks to take into account `dist/restore/index.js` and `dist/save/index.js`

brcrista:
The run failed, `dist/` has a different format than I expected 😢 https://github.com/actions/cache/tree/main/dist

dhadka:
Just FYI, we already do something like this - https://github.com/actions/cache/blob/10906ba9cd642bcc07f0f38a95a57e5c1361d156/.github/workflows/workflow.yml#L50, though I'm fine switching over so we have a consistent check for both actions.

brcrista:
Whoops, sorry @dhadka I didn't see your comment before I merged.  I'm adding the same(-ish) workflow for all actions (see https://github.com/github/c2c-actions-service/issues/2382).

I actually think the existing implementation is a bit cleaner -- didn't occur to me that you _don't_ have to `mv` the file before calling `git diff`.  I think it makes sense as its own workflow, though.  I'll make a follow-up PR.

PR #609, opened 2021-07-19
Add "see more" link to GHE-not-supported warning

slaweet:
I lived for several months thinking that support for caching action on GHE is just a matter of time, because it's such an important thing to have. Only today, I discovered that originally it was not planned at all. And that people already created some workarounds. So I hope that linking the issue from the warning message will save other people from what happened to me :-)

### DIFF
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -68,7 +68,7 @@ test("restore on GHES should no-op", async () => {
     expect(setCacheHitOutputMock).toHaveBeenCalledTimes(1);
     expect(setCacheHitOutputMock).toHaveBeenCalledWith(false);
     expect(logWarningMock).toHaveBeenCalledWith(
-        "Cache action is not supported on GHES"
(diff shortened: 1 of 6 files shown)


brcrista:
Thanks @slaweet -- I triggered CI and it looks like there is a [linting issue](https://github.com/actions/cache/pull/609/checks?check_run_id=3105429969#step:7:1).  Not sure what it is, but could you give it a look?

slaweet:
@brcrista thanks for the feedback. I fixed the lining error, amended the commit, and pushed the change. Please have a look again.

brcrista:
Fixed that error, something else failed.  Run `npm run build` and commit the `dist/index.js` file to fix.

slaweet:
I run `npm run build`, which generated changes in `dist/restore/index.js` and `dist/save/index.js`. I committed and pushed those files.

I was so naive when I thought that if I'm just editing a string literal, I can do it in GitHub without checking out the repo :smile:

brcrista (approved):
Thank you for the contribution!  :shipit:

PR #639, opened 2021-09-12
docs(examples): use expression syntax to access "steps" context

ryanvanoss:


examples.md
```diff
@@ -240,7 +240,7 @@ If using `npm config` to retrieve the cache directory, ensure you run [actions/s
   run: |
     echo "::set-output name=dir::$(npm config get cache)"
 - uses: actions/cache@v2
-  id: npm-cache # use this to check for `cache-hit` ==> if: steps.npm-cache.outputs.cache-hit != 'true'
```
joshmgross, line 243: Expressions in conditionals shouldn't require `${{ }}`

>When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression.
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

ryanvanoss: Thanks for the reference, @joshmgross! I could swear that I went without the expression syntax first, but it failed for me. Let me see if I was crazy and made that up, or if I can stand up a repo to reproduce.

ryanvanoss: Welp, I must've gone mad. I setup a quick repo and it works just fine with and without the expression syntax. Thanks for taking a look! Closing this.

PR #651, opened 2021-09-26
Fix Golang windows example - updated `README.md` test badge markdown.

magnetikonline:
- Fixed the cache paths for the Windows Golang example - the current example does not work correctly.
- While in the area, chomped some markdown whitespace.

Also some recent 2022 updates(!):

- Updated the Test build badge to use the Markdown generated by the "Create status badge" UI from within GitHub.
- Updated the `examples.md` TOC once more using https://magnetikonline.github.io/markdown-toc-generate/ 👍


magnetikonline:
FWIW - tested this via a small composite Action I recently put together - noted that the current example doesn't map to a path that actually caches Golang build resources/files under Windows.

https://github.com/magnetikonline/action-golang-cache/blob/652619ab6a0bdfb5e4f50cddd1bda080327c44f8/action.yaml#L20-L37

magnetikonline:
Ping @brcrista (calling you out - as saw you on another recent PR!). 😄

magnetikonline:
Hello 👋  @brcrista.

I _finally_ got around to putting together a POC to prove this change is needed.

See here: https://github.com/magnetikonline/golang-win-cache-test

- A simple hello world Golang program.
- Two workflows on push, running both on Windows, doing a cache - one using the old (current) and another doing the new (PR proposed) method.

See action runs of working/broken.

Also see screenshots here:

working

First screen shot shows a restored cache with a sizeable archive size.

broken

Second screen shot shows a restored cache with no actual content - just a archive file container (`192` bytes) - this shows that `%LocalAppData%\go-build` is not being expanded to the correct home of `~\AppData\Local\go-build`.

Thanks!

README.md
```diff
@@ -2,15 +2,15 @@
 
 This action allows caching dependencies and build outputs to improve workflow execution time.
 
-<a href="https://github.com/actions/cache/actions?query=workflow%3ATests"><img alt="GitHub Actions status" src="https://github.com/actions/cache/workflows/Tests/badge.svg?branch=main&event=push"></a>
+[![Tests](https://github.com/actions/cache/actions/workflows/workflow.yml/badge.svg)](https://github.com/actions/cache/actions/workflows/workflow.yml)
```
vsvipul, line 5: Why does this require change?

magnetikonline: Thanks for the review @vsvipul. Since I was in the area of `README.md` - noted this workflow badge was in the "old format" so decided to bolt on this change. 

It's the Markdown badge format generated by the GitHub badge builder:

Screen Shot 2022-02-28 at 4 02 27 pm

Screen Shot 2022-02-28 at 4 02 32 pm

Also of note - the URL format changed when using the build button Markdown UI - it's now tied to the workflow filename (e.g. `workflow.yml` in this instance), rather than `?query=workflow%3ATests`. You'd have to question that with your GitHub Actions team internally, but I guess that feels like a better hook and I tend to agree.

brcrista:
I'll defer to @vsvipul on this now

magnetikonline:
Thanks @brcrista, @vsvipul - will await your feedback 👍

README.md
magnetikonline: Screen Shot 2022-03-02 at 11 58 11 am

Ah yes, @vsvipul - the other advantage to the updated build badge URL - it now correctly highlights the correct workflow item/name in the sidebar. The old URL `?query=` isn't able to do this.

magnetikonline:
Friendly bump @vsvipul

README.md
vsvipul: Got it. Thanks for explaining.

magnetikonline: No dramas @vsvipul 👍 

Hopefully we can get the rest of the PR over the line. 👍

vsvipul (approved):
Looks like %AppData% is not being correctly expanded to a path where the cache for Go is being stored. Only other place we use this is in deno. @jheysaav You might want to check the windows deno example if this is true for that as well, we can update the docs there as well.

LGTM.

magnetikonline:
> Looks like %AppData% is not being correctly expanded to a path where the cache for Go is being stored. Only other place we use this is in deno.

Thanks @vsvipul 👍 

I think you're on the money - have raised #762 cc: @jheysaav.

PR #677, opened 2021-11-14
[examples] add c++ conan example

planetmarshall:
Add example using the [Conan](https://docs.conan.io/en/latest/) package manager for C++


planetmarshall:
See https://github.com/planetmarshall/jpegls-hdf-filter/runs/4133804617?check_suite_focus=true for a working example.

PR #682, opened 2021-11-19
Bumping up @actions/cache version to 1.0.8

aparna-ravindra:
Bumping up `@actions/cache` dependency to version 1.0.8 to support cache upload upto 10GB instead of 5GB addressing [this issue](https://github.com/actions/cache/discussions/497).

Corresponding PR in actions/toolkit is linked [here](https://github.com/actions/toolkit/pull/934)

The other changes in this PR include
1. Changes from running `npm update` as suggested [here](https://github.com/github/c2c-actions/blob/6b7f02593ed641504faf091360e05b9fc68687ac/docs/teams/platform/guides/artifact-cache/update-client-action.md#updating-the-cache-action-1)
2. With the version update, the check [Licensed](https://github.com/actions/cache/actions/workflows/licensed.yml) started failing. Fixed it by running `lincesed cache`.


package.json
```diff
@@ -23,7 +23,7 @@
   "author": "GitHub",
```
bishal-pdMSFT, line 23: Does the `version` field also need to be bumped up?

aparna-ravindra: Yes! Done!

bishal-pdMSFT: The change is not reflecting here. You may have to update the PR branch.

bishal-pdMSFT (approved):

boxofyellow (commented):
LGTM: :shipit:

brcrista (approved):

Issue #687, opened 2021-11-29
Document in the Readme that this actions is not supported on GitHub Enterprise Server


tgrall:
This issue is mostly to let know GitHub Enterprise users that this action is not working on GitHub Enterprise Server (GHES).

This will probably avoid the fact that people would test/fail.

We can point them to the issue #505

PR #688, opened 2021-11-29
Document the fact that this action is not supported on GitHub Enterprise Server 

tgrall:
I have added some information in the readme to let user know that it is not supported in GitHub Enterprise Server. This would avoid seeing user trying and failing on such deployment.

see #687

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,10 @@ steps:
 
 > Note: The `id` defined in `actions/cache` must match the `id` in the `if` statement (i.e. `steps.[ID].outputs.cache-hit`)
 
+## Known limitation
+
+- The `action/cache` is not supported on GithHub Enterprise Server deployments. GitHub Enterprise Server uses self-hosted runners, dependencies are typically cached on the runner by whatever dependency management tool is being used (npm, maven, etc.), so saving these to an external cache would be less efficient.
+
 ## Contributing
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,6 +167,10 @@ steps:
 
 > Note: The `id` defined in `actions/cache` must match the `id` in the `if` statement (i.e. `steps.[ID].outputs.cache-hit`)
 
+## Known limitation
+
+- The `action/cache` is not supported on GithHub Enterprise Server deployments. GitHub Enterprise Server uses self-hosted runners, dependencies are typically cached on the runner by whatever dependency management tool is being used (npm, maven, etc.), so saving these to an external cache would be less efficient.
```
brcrista, line 172:
```suggestion
- `action/cache` is currently not supported on GitHub Enterprise Server. <https://github.com/github/roadmap/issues/273> is tracking this.

Since GitHub Enterprise Server uses self-hosted runners, dependencies are typically cached on the runner by whatever dependency management tool is being used (npm, maven, etc.), which eliminates the need for caching in some scenarios.
```

The issue is that caching needs new functionality added to work on prem.  It needs to use user-provided storage in the same way that logs, artifacts, and packages do, rather than the cloud storage it uses on github.com.

vsvipul (approved):

vsvipul:
Merged this. Thanks @tgrall  @brcrista 🚀

PR #690, opened 2021-12-03
Simplify the Node.js NPM caching example

### LINKED ISSUE #490: Node npm 'windows' & 'multiple systems' examples are identical
See https://github.com/actions/cache/blob/main/examples.md#node---npm.

The first example uses `~/.npm` for linux setup, fine.

The Windows example uses `npm config get cache` to find the cache path, and then caches that, rather than using any specific path.

The 'Using multiple systems' example also does exactly the same. In fact, it's character for character identical. It looks like #223 did this as part of fixing the Windows example, but I think it's a bit confusing to have the exact same code twice.

Imo, this would be simpler & clearer all round with just one single example for all 3 cases (i.e. the 'multiple systems' platform-agnostic one).

pimterry:
This fixes #490. There's no good reason to have 3 examples here, especially when two of them (Windows & 'multiple systems') are effectively the same (they differ only in the ids they give the steps involved). The Linux example was different, but not better for any use cases - it just does the same thing in a more fragile & platform-specific way.

This change simplifies to a single suggested approach (the 3rd existing example) which works immediately for all platforms.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -18,9 +18,6 @@
 - [Java - Gradle](#java---gradle)
 - [Java - Maven](#java---maven)
 - [Node - npm](#node---npm)
-  - [macOS and Ubuntu](#macos-and-ubuntu)
-  - [Windows](#windows-3)
-  - [Using multiple systems and `npm config`](#using-multiple-systems-and-npm-config)
 - [Node - Lerna](#node---lerna)
 - [Node - Yarn](#node---yarn)
 - [Node - Yarn 2](#node---yarn-2)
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -200,40 +197,12 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
...
 ## Node - npm
 
-For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` on Windows. See https://docs.npmjs.com/cli/cache#cache
+For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` on Windows, but it's possible to use `npm config get cache` to find the path on any platform. See https://docs.npmjs.com/cli/cache#cache for more details.
 
 If using `npm config` to retrieve the cache directory, ensure you run [actions/setup-node](https://github.com/actions/setup-node) first to ensure your `npm` version is correct.
 
 >Note: It is not recommended to cache `node_modules`, as it can break across Node versions and won't work with `npm ci`
 
-### macOS and Ubuntu
```
vsvipul, line 209: IMO, we should keep this example as well, as it is much more simple without using an extra step. You can instead combine the two headers below (Windows and multiple systems) into one, which will be a much better approach. @bishal-pdMSFT  thoughts?

examples.md
```diff
@@ -200,40 +197,12 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
 
 ## Node - npm
 
-For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` on Windows. See https://docs.npmjs.com/cli/cache#cache
+For npm, cache files are stored in `~/.npm` on Posix, or `%AppData%/npm-cache` on Windows, but it's possible to use `npm config get cache` to find the path on any platform. See https://docs.npmjs.com/cli/cache#cache for more details.
```
vsvipul, line 200: For the link use markdown. See [npmjs docs](https://docs.npmjs.com/cli/cache#cache) for more details.

vsvipul (changes requested):
Left some comments. Do check.

examples.md
pimterry: Fixed :+1:

examples.md
pimterry: It's slightly simpler, but this is just copying and pasting anyway. Personally as a user, I just want one example that works everywhere.

More importantly though, this example isn't even 100% correct on Linux & Mac, as the hardcoded path here is actually configurable (by local & global config or env var). Checking `npm config get cache` is the only way to know what the correct path is for sure on any platform.

Also every other Node package manager in this file (Lerna, Yarn, Yarn 2) already uses a single two-step example like this, instead of having a special case that hardcodes the Mac/Linux default path.

pimterry:
@vsvipul any thoughts on this?

vsvipul:
@pimterry Apologies for the delay in replying. Let's merge this. Can you please fix the merge conflicts?

pimterry:
No problem @vsvipul - rebased with merge conflicts fixed.

Makaiza:
Hello

vsvipul (approved):
LGTM. Thank you for contributing 🚀

Kaizamarvel (approved):

PR #696, opened 2021-12-10
Add actions-cache team in CODEOWNERS

bishal-pdMSFT:
Add actions-cache team (Bishal's team) to CODEOWNERS. Existing team remain CODEOWNERS as well.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-* @actions/artifacts-actions
 * @actions/actions-cache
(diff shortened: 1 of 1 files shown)


vsvipul (approved):

.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
 * @actions/artifacts-actions
```
brcrista, line 1: Can we remove this group?  This means that `@actions/artifacts-actions` will get pinged on every PR.

You can still pull them in manually if you need support, but I think we should make it clear that your team owns this now.

bishal-pdMSFT: done

PR #706, opened 2021-12-29
Improve Haskell example

### LINKED ISSUE #555: The example Haskell cache setup is insufficient
The recommendet setup for Haskell according to https://github.com/actions/cache/blob/main/examples.md#haskell---cabal is
```
- name: Cache ~/.cabal/packages, ~/.cabal/store and dist-newstyle
  uses: actions/cache@v2
  with:
    path: |
      ~/.cabal/packages
      ~/.cabal/store
      dist-newstyle
    key: ${{ runner.os }}-${{ matrix.ghc }}
```
but this is not a good setup: The key doesn’t vary when the code or dependencies vary, so once the cache has been created, it will re-use the same cache over and over again, even as dependencies change. (This may be confusing to people migrating from Travis, which would always update the cache if the cache directories have changes).

A possible fix that may work well enough in many cases is to do write
```
- name: Cache ~/.cabal/packages, ~/.cabal/store and dist-newstyle
  uses: actions/cache@v2
  with:
    path: |
      ~/.cabal/packages
      ~/.cabal/store
      dist-newstyle
    key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('**/*.cabal', '**/cabal.project', '**/cabal.project.freeze') }}
    restore-keys: cabal-${{ runner.os }}-${{ matrix.ghc }}-
```

nomeata:
as proposed in https://github.com/actions/cache/issues/555. Fixes #555.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -168,8 +168,8 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
       ~/.cabal/packages
       ~/.cabal/store
       dist-newstyle
-    key: ${{ runner.os }}-${{ matrix.ghc }}
-```
+    key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('**/*.cabal', '**/cabal.project', '**/cabal.project.freeze') }}
+    restore-keys: cabal-${{ runner.os }}-${{ matrix.ghc }}-```
 
 ## Java - Gradle
 
--- a/examples.md
+++ b/examples.md
@@ -168,8 +168,9 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
       ~/.cabal/packages
       ~/.cabal/store
       dist-newstyle
-    key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('**/*.cabal', '**/cabal.project', '**/cabal.project.freeze') }}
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -168,8 +168,8 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
       ~/.cabal/packages
       ~/.cabal/store
       dist-newstyle
-    key: ${{ runner.os }}-${{ matrix.ghc }}
-```
+    key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('**/*.cabal', '**/cabal.project', '**/cabal.project.freeze') }}
```
vsvipul, line 171: Remove `cabal-` from the beginning for key and restore-keys as it is unnecessary and making the key long. Also move 3 backticks for markdown to last line not just the end of code.

vsvipul: image
Also, you can see the rich diff at https://github.com/actions/cache/pull/706/files?short_path=cfb2db6#diff-cfb2db6f153a4484d9bbd408f3227a264dce68fadd747cecdc30bda012b4d8cd . You have added the change under java-gradle instead of haskell-cabal.

nomeata: Sorry, maybe I shoudn’t have used the Github editor

vsvipul (approved):
LGTM!

PR #708, opened 2022-01-17
Enabling actions cache for GHES

tiwarishub:
### Description
This PR is enabling actions cache on GHES by removing the ``isGhes()`` condition and also removing its related test cases.

### Release Process
This change will not be released to the marketplace as an official release, till ac on GHES is released and tested. Till that point, this change/commit will be tagged with ``enabledForGHES.v1`` and users who have GHES with AC can use these changes with ``actions/cache@enabledForGHES.v1``. The reason for not creating a release is that it will then become visible to the marketplace.

### Open question
@bishal-pdMSFT / @aparna-ravindra 
As we know this action is not bydefault packaged in GHES. Do we want this to be packaged by default in GHES releases like these https://github.com/github/ghes-actions/tree/master/config/actions  actions or we are expecting the admin to add it using the ``actions-sync`` tool.

### Linked Issue
https://github.com/github/c2c-actions/issues/3851

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -17,24 +17,6 @@ afterEach(() => {
     delete process.env[RefKey];
 });
 
-test("isGhes returns true if server url is not github.com", () => {
-    try {
-        process.env["GITHUB_SERVER_URL"] = "http://example.com";
-        expect(actionUtils.isGhes()).toBe(true);
-    } finally {
-        process.env["GITHUB_SERVER_URL"] = undefined;
-    }
-});
-
-test("isGhes returns true when server url is github.com", () => {
-    try {
-        process.env["GITHUB_SERVER_URL"] = "http://github.com";
-        expect(actionUtils.isGhes()).toBe(false);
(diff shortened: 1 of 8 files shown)


aparna-ravindra:
>  Do we want this to be packaged by default in GHES releases like these https://github.com/github/ghes-actions/tree/master/config/actions actions or we are expecting the admin to add it using the actions-sync tool.

This is to be taken care in this issue: https://github.com/github/c2c-actions/issues/3617

bishal-pdMSFT:
Looks good to me! Do we need similar change for setup-* actions as well?

bishal-pdMSFT:
Please link PR to feature issue.

vsvipul (approved):
LGTM. Readme can also be updated to remove the GHES limitation.

aparna-ravindra:
> Readme can also be updated to remove the GHES limitation.

Let's update the readme when this is merged into main branch.

tiwarishub:
>Do we need similar change for setup-* actions as well?

Issue #661, opened 2021-10-17
An output for exact cache hit


laughedelic:
I think it would be useful to indicate somehow whether cache was restored using the exact `key` or one of the fallback `restore-keys`. 

A use case for this is being able to skip some subsequent steps only in the case of an exact cache hit. For example, if a build tool uses incremental compilation, a partial cache hit is good, it saves _some_ time, but if the cache hit was exact, we can skip the compilation step altogether which might save us much more time (e.g. build tool loading the project). A concrete case of such build tool it [sbt](https://www.scala-sbt.org/): in a medium-sized project with several modules it might spend up to a minute loading the project.

So I'm suggesting to add another output value to the action which will indicate whether hit was exact or not.

OndrejSpanel:
I think @laughedelic would like to check if the hit was on a primary key, or on a restore key. The primary key hit is "exact cache hit", while the restore key hit is "partial hit".

I think this often makes sense - when a restore key is hit, it means a not up-to-date version of the files is obtained, which is usefull, but is likely to need at least some updating (like calling `sbt update`)´, while on primary key you know everything is already up to date and you can skip any such steps.

OndrejSpanel:
After reading the documentation again I think `cache-hit` is set to true only on the exact hit, not on the restore key hit. You are therefore not notified about a partial hit at all - it might speed up something or it might not, you are expected to proceed as if there is no hit.

> A boolean value to indicate an exact match was found for the key

This is confirmed by looking at sources:

```ts
            const isExactKeyMatch = utils.isExactKeyMatch(primaryKey, cacheKey);
            utils.setCacheHitOutput(isExactKeyMatch);
```

The documentation could perhaps be changed to explicitly state this, like:

_A boolean value to indicate an exact match was found for the key (not set when only a restore key matches)_

jsoref:
`cache-hit` indeed is only true if you're hitting the primary key.

Unfortunately, you can't distinguish between a `cache-miss` and a hit of a `restore-key`.

I would love for `cache-hit` to tell me the key it hit instead of just a boolean for the primary key.

bishal-pdMSFT:
Agree, two changes make sense:
1. Log the restore key for which the partial match happened.
2. Update doc to explicitly call out about `cache-hit`

Would love to see some PR contribution as it looks like a good first issue.

jsoref:
Would you want it to use a different output, or can I reuse the current one?

Offhand, it feels like changing the existing output is probably an API break.

[...]

PR #717, opened 2022-01-23
Cache key output

jsoref:
This addresses https://github.com/actions/cache/issues/661#issuecomment-1019511446


jsoref:
Fwiw, https://github.com/jsoref/actions-cache/commit/edb9b2811ea31dd82a16eef13820e78d0c8a8f1b shows tests pass for this commit

src/constants.ts
```diff
@@ -6,7 +6,8 @@ export enum Inputs {
 }
 
 export enum Outputs {
-    CacheHit = "cache-hit"
+    CacheHit = "cache-hit",
+    CacheKey = "cache-key"
```
bishal-pdMSFT, line 10: Would it be better to name it `cache-matched-key` to keep it aligned with the corresponding state being set. 
Not sure if it is allowed to have state and output with same id

__tests__/actionUtils.test.ts
```diff
@@ -102,7 +102,8 @@ test("setOutputAndState with exact match to set cache-hit output and state", ()
     actionUtils.setOutputAndState(key, cacheKey);
 
     expect(setOutputMock).toHaveBeenCalledWith(Outputs.CacheHit, "true");
-    expect(setOutputMock).toHaveBeenCalledTimes(1);
+    expect(setOutputMock).toHaveBeenCalledWith(Outputs.CacheKey, key);
```
bishal-pdMSFT, line 105: Shouldn't it match to `cacheKey`?

bishal-pdMSFT:
@vsvipul please take a look

src/constants.ts
jsoref: I will use whatever people request. But, I'd like to avoid updates unless people agree on direction.

__tests__/actionUtils.test.ts
jsoref: Shrug, `key` and `cacheKey` are the same.

Davincayson (approved):

jsoref:
I spent some more time on this, it appears that the @actions/cache method doesn't actually provide enough information.

This new flavor does yield enough information, although admittedly it's more expensive. -- For fallback cases, it will retrieve the cache item twice, and it makes n additional cache queries. I could skip the doubled cache retrieval by changing the primary call to not include fallback items.

It'd be nicer if there was some easy way to figure out from the cache which item was providing the answer.

jsoref:
@lucascosti: I don't suppose you could look into this?

I mentioned it in github/docs#18524

If someone is willing to review, I'm happy to update.

lucascosti:
Sorry, @jsoref, because these are code changes, the Actions engineers must review this (cc @aparna-ravindra)

arminrosu:
@jsoref @lucascosti any chance we can revive this PR? If there's work to be done, I'm willing to take it over, just need to know what you need to get it merged.

jsoref:
I'm happy to update, but only if someone will commit to reviewing it and @vsvipul / @bishal-pdMSFT seem to have just dropped the hot-potato.

src/restore.ts
```diff
@@ -54,7 +54,25 @@ async function run(): Promise<void> {
...
 
-            core.info(`Cache restored from key: ${cacheKey}`);
+            let foundKey: string;
+            if (isExactKeyMatch) {
+                foundKey = primaryKey;
+                core.info(`Cache restored for key: ${foundKey}`);
+            } else {
+                let i: number;
+                for (i = 0; i < restoreKeys.length - 1; i++) {
+                    const fallbackCacheKey = await cache.restoreCache(
```
andreasabel, line 64: Why are you calling `restoreCache` again?
See line 36: it is already called with the restore keys, so how can this succeed?

andreasabel (changes requested):
What is this PR trying to achieve?

PR #718, opened 2022-01-24
Update types/node and eslint-plugin-import

vsvipul:
Update types/node and eslint-plugin-import to suppress dependabot alerts.


dist/restore/index.js
```diff
@@ -37903,9 +37903,17 @@ AbortError.prototype = Object.create(Error.prototype);
 AbortError.prototype.constructor = AbortError;
 AbortError.prototype.name = 'AbortError';
 
+const URL$1 = Url.URL || whatwgUrl.URL;
```
bishal-pdMSFT, line 37906: Did node type update introduced these save and restore js file changes?

vsvipul: No @bishal-pdMSFT . I suppose the .js files in main are out of date. I've updated those as well here by building those, as it was one of the tests.

vsvipul:
Closing as this generates extra .js file changes. Don't know the root cause for this yet.

Soundpisey:
I have seen it. Thank you very much.

PR #721, opened 2022-01-28
This is a good tip if you're using Alpine

JJ:
(Which does not have a POSIX-compliant tar by default)


bishal-pdMSFT (approved):

bishal-pdMSFT:
Thanks @JJ for your contribution 🎉

Issue #494, opened 2020-12-25
Excluded sub directory not working


mustaphazorgati:
Hey there!
The README.md provides an example where an excluded sub-path is provided. Unfortunately that sub-path still ends up in the cache.
Here is a modified example using the local maven repository.
```
- name: Cache multiple paths
  uses: actions/cache@v2
  with:
    path: |
      ~/.m2
      !~/.m2/pro/taskana
    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
```

Reading out on the next (cached) build the files in `~/.m2/pro/taskana` are contained in the cache. The following adjustment didn't work either: 
```
- name: Cache multiple paths
  uses: actions/cache@v2
  with:
    path: |
      ~/.m2
      !~/.m2/pro/taskana/**
    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
```

Is this a bug or am I having the wrong expectations? Our current workaround is to delete that subfolder in a dedicated step. It works, but is kinda annoying.

mustaphazorgati:
@dhadka I've seen you answering a lof of issues recently. Sorry for asking you directly, but can you have a look at this?

dhadka:
@mustaphazorgati Thanks for reporting this (and sorry I didn't see this earlier).  I was able to [reproduce this issue](https://github.com/dhadka/cache-test-exclusion/runs/1855643206?check_suite_focus=true).  I filed a [bug over in the actions/toolkit repo](https://github.com/actions/toolkit/issues/713) that hosts the underlying logic for search patterns.

mustaphazorgati:
Thank you @dhadka for your quick response! 
Just an FYI for other readers: The provided reproduction of the issue is found [here](https://github.com/dhadka/cache-test-exclusion/runs/1855553990?check_suite_focus=true). 
@dhadka link shows a workflow where the `Post cache test` step is not executed ;)

Adnn:
Any update on this issue? (This is making the cache difficult to use in complex workflows)

robinp:
FYI guys I added some code pointers to the underlying https://github.com/actions/toolkit/issues/713#issuecomment-1030692990 - for me it seems passing an extra option to avoid matching directiories in the globber would solve this.

Phantsure:
This issue still exists?

aparna-ravindra:
[...]

PR #726, opened 2022-02-05
Call out caveats with cache exclude patterns

robinp:
The pattern shown in the current README doesn't work as-is (see https://github.com/actions/cache/issues/494). Making it explicit in the README until fixed can save time for people trying to use the feature.


robinp:
Pinging.

PR #727, opened 2022-02-07
Fix test description

ygnoh:
Fixed a test description different from the body.


ashwinsangem (approved):
Thanks for fixing @ygnoh, LGTM.

PR #738, opened 2022-02-11
[D] Fix: should use dub.selections.json for cache

kubo39:
see upgrade section in https://dub.pm/commandline

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -86,7 +86,7 @@ steps:
 - uses: actions/cache@v2
   with:
     path: ~/.dub
-    key: ${{ runner.os }}-dub-${{ hashFiles('**/dub.json') }}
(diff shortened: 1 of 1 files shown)


vsvipul:
Hi @kubo39 . A quick read suggests to me that dub.json is like npm's package.json and dub.selection.json is more like npm's package-lock.json . Can you quickly explain why dub.selection.json should be preferred over dub.json ?

kubo39:
@vsvipul Hi, and sorry for late response.

> A quick read suggests to me that dub.json is like npm's package.json and dub.selection.json is more like npm's package-lock.json.

Yes, my understanding is that this is done so that the key changes when the dependencies change.

vsvipul (approved):
LGTM

PR #742, opened 2022-02-15
Update node version, @types/node and some outdated dependencies

vsvipul:
- Update node version and @types/node to v16.
- Update outdated dependencies.


package.json
```diff
@@ -31,18 +31,18 @@
   "devDependencies": {
```
bishal-pdMSFT, line 31: should we update the `version` field if we plan to ship this change?

bishal-pdMSFT (approved):

package.json
vsvipul: we have other changes as well. so will do the version release in a separate PR.

PR #743, opened 2022-02-16
Added `cache-restored` boolean flag

mkatychev:
Description:
```yaml
  cache-restored:
    description: 'A boolean value to indicate an exact or partial match was found resulting in a restore'
```

This flag should be used similarly to `steps.cache.outputs.cache-hit` to indicate when there is a partial match for a cache. 

The aim here is the ability to avoid certain install steps if any part of the cache was restored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,9 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 ### Outputs
 
 * `cache-hit` - A boolean value to indicate an exact match was found for the key
+* `cache-restored` - A boolean value to a match was found resulting in a restored cache
 
-> See [Skipping steps based on cache-hit](#Skipping-steps-based-on-cache-hit) for info on using this output
+> See [Skipping steps based on `cache-hit`](#Skipping-steps-based-on-cache-hit) for info on using this output
 
 ### Cache scopes
 The cache is scoped to the key and branch. The default branch cache is available to other branches. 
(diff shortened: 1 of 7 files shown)


yfyau:
Support! This would be useful

anuragc617:
@mkatychev we are working on providing granular control and looking over all the issues related to it. Then we will come up with a design for having such flags.

Jimimaku:
> Description:
> 
> ```yaml
>   cache-restored:
>     description: 'A boolean value to indicate an exact or partial match was found resulting in a restore'
> ```
> 
> This flag should be used similarly to `steps.cache.outputs.cache-hit` to indicate when there is a partial match for a cache.
> 
> The aim here is the ability to avoid certain install steps if any part of the cache was restored.

Jimimaku:
> @mkatychev we are working on providing granular control and looking over all the issues related to it. Then we will come up with a design for having such flags.

Jimimaku:
> Support! This would be useful

Flamefire:
I find the description hard to understand. As of the commit:

> ```
>   cache-hit:
>     description: 'A boolean value to indicate an exact match was found for the primary key'
>   cache-restored:
>     description: 'A boolean value to indicate a primary key match was found resulting in a restore'
> ```

I.e. the difference is "primary key match" vs "exact match for the primary key". From reading this the difference is not clear to me.

And from the PR description:

> description: 'A boolean value to indicate an exact or partial match was found resulting in a restore'

This could be misinterpreted as "true=exact, false=partial match"

A use case:

```
key: cache-${{ github.run_id }}
restore-keys: cache-
```

So maybe `cache-restored` is better described as

> A boolean value to indicate a successful cache restore; either by a match for on `key` or (partial) match on any `restore-keys`

So emphasis on "Cache restored" with a short "by any means" with hints on where to look for the reason

kotewar:
We have released a new `restore` action that outputs the cache-primary-key and the cache-matched-key that we believe solves the problem this PR is trying to solve. Hence closing it.

PR #746, opened 2022-02-22
Revert "Add example for Haskell Stack"

vsvipul:
Reverts actions/cache#173


ashwinsangem (approved):

ashwinsangem:
`master` branch is not active, so retaining its old state.

PR #751, opened 2022-02-28
UPDATE Fails cache restore on errors

mlau-displayr:
Fails the caches when there is random failure like a connection timeout

### DIFF
--- a/src/restore.ts
+++ b/src/restore.ts
@@ -65,6 +65,7 @@ async function run(): Promise<void> {
         }
     } catch (error) {
         core.setFailed(error.message);
+        utils.setCacheHitOutput(false);
     }
 }
 


mlau-displayr:
image
Prevent situations where the cache passes on errors

vsvipul:
@mlau-displayr Why do you want to set the cache hit output to false? Is there any particular usecase that I'm missing here. We are already failing the cache when there is an error like connection timeout.

mitchell-demol-displayr:
@mlau-displayr is no longer working at Displayr, so his account has been removed. 
I believe this caused our builds to fail in a subsequent step because the cache step passed with this warning and didn't actually load any of the cached items. That may not be the case anymore, or maybe he was mistaken. In any case, this PR can be closed as it hasn't caused us any issues since this was reported.

vsvipul:
Thank you for the headsup @mitchell-demol-displayr . Closing this PR for now.

PR #765, opened 2022-03-13
[docs] remove star search from yarn.lock

lookfirst:
i dunno, some recent change to cache caused the `**/yarn.lock` lookup to start failing for me. I just have a standard project with a yarn.lock in the top level and this fixed it.


Phantsure:
@lookfirst Changing the path of yarn file in example may not help everyone. That path is for the `yarn.lock` file which can be at different location for different projects. 
Thanks for your contribution.

Phantsure:
@lookfirst Can you try it again with the current recommendation and latest version of cache? It seems to work fine for me

PR #771, opened 2022-03-21
Updated what's new.

ashwinsangem:
👉 [Rendered](https://github.com/actions/cache/blob/users/ashwinsangem/update_whats_new/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ See ["Caching dependencies to speed up workflows"](https://help.github.com/githu
 
 ## What's New
 
+* Updated the minimum runner version support from node 12 -> node 16.
+* Increased the cache size limit to 10 GB.
 * Added support for multiple paths, [glob patterns](https://github.com/actions/toolkit/tree/main/packages/glob), and single file caches.
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ See ["Caching dependencies to speed up workflows"](https://help.github.com/githu
 
 ## What's New
 ### v3
-
 * Updated the minimum runner version support from node 12 -> node 16.
 
 ### v2
--- a/README.md
+++ b/README.md
@@ -9,8 +9,11 @@ This action allows caching dependencies and build outputs to improve workflow ex
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
 ## What's New
+### v3
 
 * Updated the minimum runner version support from node 12 -> node 16.
+
+### v2
 * Increased the cache size limit to 10 GB.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,8 @@ See ["Caching dependencies to speed up workflows"](https://help.github.com/githu
 
 ## What's New
 
+* Updated the minimum runner version support from node 12 -> node 16.
+* Increased the cache size limit to 10 GB.
 * Added support for multiple paths, [glob patterns](https://github.com/actions/toolkit/tree/main/packages/glob), and single file caches.
```
vsvipul, line 15: Should we remove the other items? They've been here for too long.

ashwinsangem: True, that sounds reasonable. @bishal-pdMSFT, do you have any thoughts?

bishal-pdMSFT: My suggestion would be to group them by `v3` and `v2`.

README.md
```diff
@@ -10,6 +10,8 @@ See ["Caching dependencies to speed up workflows"](https://help.github.com/githu
 
 ## What's New
 
+* Updated the minimum runner version support from node 12 -> node 16.
+* Increased the cache size limit to 10 GB.
```
bishal-pdMSFT, line 14: This was already there in v2.

README.md
bishal-pdMSFT: My suggestion would be to keep both v2 and v3 here for now. And add a CHANGELOG which will start summarizing the changes from 3.0.0 version onwards. Anytime we publish a new version in future, we should keep updating the changelog.

README.md
```diff
@@ -177,6 +182,12 @@ steps:
 
 Since GitHub Enterprise Server uses self-hosted runners, dependencies are typically cached on the runner by whatever dependency management tool is being used (npm, maven, etc.).  This eliminates the need for explicit caching in some scenarios.
 
+## Changelog schedule and history
```
vsvipul, line 185: Should we make a separate file called CHANGELOG.md . Because this will make README unnecessarily long in future.

vsvipul (approved):

README.md
```diff
@@ -9,7 +9,12 @@ This action allows caching dependencies and build outputs to improve workflow ex
 See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
 
 ## What's New
+### v3
```
vsvipul, line 13: can remove this empty line

README.md
ashwinsangem: I was thinking of the same. But, that would mean that the marketplace readme would not contain the changelog. So, I went this section to the bottom so that the current sections don't lose any visibility.

vsvipul: ok looks good for now then.

vsvipul (approved):

PR #801, opened 2022-05-04
Update dependencies to suppress dependabot alerts related to minimist version <"1.2.6"

vsvipul:
- Update dependencies to suppress dependabot alerts related to minimist version <"1.2.6".
- Update some other outdated dependencies like typescript, @types/node, @typescript-eslint/eslint-plugin, @typescript-eslint/parser, eslint.
- Update @actions/core to 1.7.0
- typescript update had some breaking changes. The .ts file changes are related to that.


src/save.ts
```diff
@@ -49,7 +49,7 @@ async function run(): Promise<void> {
                 uploadChunkSize: utils.getInputAsInt(Inputs.UploadChunkSize)
             });
             core.info(`Cache saved with key: ${primaryKey}`);
-        } catch (error) {
+        } catch (error: any) {
```
vsvipul, line 52: Using `any` type should be avoided and `unknown` should be used here and everywhere else instead. That will required adding types here. Will work with @t-dedah  on doing this.

t-dedah: Done

PR #802, opened 2022-05-08
Update ts-jest version and package-lock.json

vsvipul:
- Update ts-jest version to 28.0.2
- Minimist was not updated for ts-jest in previous update in package-lock.json as there was some problem with older npm version. This updates that and removed minimist dependency from ts-jest completely.


vsvipul:
@tiwarishub  Can you review this also please?

tiwarishub (approved):

PR #804, opened 2022-05-20
Update README to call out for version of cache

Phantsure:
Add details about versions of a cache to README

Context: Some issue were created regarding different caches being saved with different runner OS or different paths with same cache key. This PR addresses those issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,9 +179,10 @@ steps:
 
 
 ## Cache Version
-Cache version is unique for a combination of compression tools(Gzip, Zstd, etc) and the path of cache, which may differ according to runner OS as well. If two caches have different versions, then they are identified as unique cache entries. 
+Cache version is unique for a combination of compression tool used for compression of cache (Gzip, Zstd, etc based on runner OS) and the path of directories being cached. If two caches have different versions, they are identified as unique cache entries. This also means that a cache created on `windows-latest` runner can't be restored on `ubuntu-latest` as cache `Version`s are different. 
 
-Example: Below example will create 3 unique caches with same keys.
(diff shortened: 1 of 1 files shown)


bishal-pdMSFT (changes requested):
Does not have description. And the issue is not linked.

README.md
```diff
@@ -177,6 +177,57 @@ steps:
 
 > Note: The `id` defined in `actions/cache` must match the `id` in the `if` statement (i.e. `steps.[ID].outputs.cache-hit`)
 
+
+## Cache Version
+Cache version is unique for a combination of compression tools(Gzip, Zstd, etc) and the path of cache, which may differ according to runner OS as well. If two caches have different versions, then they are identified as unique cache entries.
```
bishal-pdMSFT, line 182:
```suggestion
Cache version is unique for a combination of compression tool used for compression cache (Gzip, Zstd etc based on runner OS) and the path of directories being cached. If two caches have different versions, they are identified as unique cache entries. 
```

Kaizamarvel (approved):

README.md
```diff
@@ -177,6 +177,57 @@ steps:
 
 > Note: The `id` defined in `actions/cache` must match the `id` in the `if` statement (i.e. `steps.[ID].outputs.cache-hit`)
 
+
+## Cache Version
+Cache version is unique for a combination of compression tools(Gzip, Zstd, etc) and the path of cache, which may differ according to runner OS as well. If two caches have different versions, then they are identified as unique cache entries. 
+
+Example: Below example will create 3 unique caches with same keys.
```
bishal-pdMSFT, line 184:
```suggestion
Example: Below example will create 3 unique caches with same keys. Ubuntu and windows runners will use different compression technique and hence create two different caches. And `build-linux` will create two different caches as the `paths` are different.
```

Kaizamarvel (approved):

Issue #624, opened 2021-08-10
Do not save cache if files are empty


boardfish:
Even if the target files are not present or empty, a cache key is saved:

```
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/boardfish/project --files-from manifest.txt
Cache Size: ~0 MB (22 B)
Cache saved successfully
Cache saved with key: Linux-project-data-17007abbc08...
```

I'd prefer not to register a key or save a cache in this particular instance.

For now, I'll need to check whether the file exists manually. But I'd rather not have to do that and check against a cache hit instead.

```yml
- name: TEST  | Check crystalball map is available
  id: check_crystalball_map_exists
  uses: andstor/file-existence-action@v1
  with:
    files: "tmp/crystalball_data.yml"
```

paulrberg:
Yes it would be great if we could disable the saving of cache in this circumstance.

cderv:
I would still like to this possible.

kotewar:
The issue has been fixed and released in [v3.0.3](https://github.com/actions/cache/releases/tag/v3.0.3)
Cache action will now check if at least one path/file/folder exists for the given `path` in the action. If nothing is found, no cache will be saved.

> Note: We are not checking for the size as the compressed tar/zip might still be a few bytes in size. So it is not possible to tell if the cache is empty or not just by the size of the compressed file.

Kurt-von-Laven:
Now that the `lookup-only` parameter has been added, the drawback of not saving empty caches is that it forces you to create an empty file if you just want to cache whether something has happened. For example, when vendoring dependencies, such as with Yarn [Zero-installs](https://yarnpkg.com/features/caching#zero-installs), it's important to check the correctness of committed dependencies against the package registry, but this only needs to be done once per dependency bump.

PR #808, opened 2022-05-30
Upgraded @actions/cache to 2.0.5

kotewar:
As part of fix needed for #624 , [actions/toolkit](https://github.com/actions/toolkit) was modified recently to warn the user telling no cache is being saved when there are no paths matching the given input paths. 

This PR is to update the @actions/cache dependency with recently published [v2.0.5](https://www.npmjs.com/package/@actions/cache/v/2.0.5) version of the same.


tiwarishub (approved):
LGTM

RELEASES.md
```diff
@@ -6,3 +6,9 @@
 ### 3.0.1
 - Added support for caching from GHES 3.5.
 - Fixed download issue for files > 2GB during restore.
+
+### 3.0.2
+- Added support for dynamic cache size cap on GHES.
+
+### 3.0.3
```
vsvipul, line 13: why are we doing 2 releases?

kotewar: We're doing only one release i.e. 3.0.3. We had missed updating release notes for the previous release, hence added that information here just to track the same.

vsvipul: Ah nice. you might also want to add these to README.md. I'll suggest removing v2 from there now as they are quite old.

kotewar: Alright, let me do that as well. Will also update the link for 

> Refer [here](https://github.com/actions/cache/blob/v1/README.md) for previous versions 


to [v2](https://github.com/actions/cache/blob/v2/README.md) as it is currently pointing to [v1 README](https://github.com/actions/cache/blob/v1/README.md).

kotewar:
Tested the changes in my local repository. Things are working as expected. Merging the PR.

abirhasan132:
#808

abirhasan132 (commented):
8b54d92f410048d51683f5f298411de309e50500

Issue #689, opened 2021-12-02
Fails on tar exit code 1 on ubuntu-latest


cyberpwnn:
I need to share around a small but heavily processed repository to multiple jobs, 

```yml
    runs-on: ubuntu-latest
    steps: 
      - name: Cache Repository
        uses: actions/cache@v2
        with:
          path: |
            ~/
            !~/.git/
```

Fails with

```
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/xxx/xxx--files-from manifest.txt
/usr/bin/tar: ../../../work/_temp/6483fbb9-3579-4fb0-a582-3316b336d5a9/cache.tzst: file changed as we read it
Warning: Tar failed with error: The process '/usr/bin/tar' failed with exit code 1
```

vsvipul:
Hi @cyberpwnn ! Is this repo public? If yes, can you share link to the workflow run. If not, can you share the logs here?

MatthieuLeboeuf:
Hello,
I have the same error
logs
![iAMjpav](https://user-images.githubusercontent.com/59397088/147604704-caa82bf0-9801-48bf-93f5-af460b904578.png)

dopplershift:
Seeing the same issue on [this repo](https://github.com/Unidata/MetPy). [Here's](https://github.com/Unidata/MetPy/runs/5056365907?check_suite_focus=true#step:13:1) a representative action run:
```
Post job cleanup.
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/MetPy/MetPy --files-from manifest.txt
/usr/bin/tar: ../../../work/_temp/bdee9e41-e8-4f2a-9912-6eb32de36aa/cache.tzst: file changed as we read it
Warning: Tar failed with error: The process '/usr/bin/tar' failed with exit code 1
Post job cleanup.
Cache hit occurred on the primary key setup-python-Linux-pip-9225292fdf311547e1fda261cbc2734e26bc2c516bdcf7d87f5c7af6de21fb8c, not saving cache.
```

sammcj:
Can confirm this is still an issue.

zstd is installed but:

```
/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/ec2-user/actions-runner/_work/my-repo/my-repo --files-from manifest.txt
[...]

PR #812, opened 2022-06-06
Updated @actions/cache to v2.0.6

kotewar:
As part of fix needed for https://github.com/actions/cache/issues/689 , [actions/toolkit](https://github.com/actions/toolkit) was modified recently to exclude the tar file from trying to add itself to the tar resulting into error on `ubuntu-latest` image.

This PR address the fix and has update to the @actions/cache dependency with recently published [v2.0.6](https://www.npmjs.com/package/@actions/cache/v/2.0.6) version of the same.

### DIFF
--- a/.licenses/npm/@actions/cache.dep.yml
+++ b/.licenses/npm/@actions/cache.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: "@actions/cache"
-version: 2.0.5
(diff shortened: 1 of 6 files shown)


vsvipul:
can we create a new cache version and update changelog in the same pr ?

vsvipul (approved):

PR #819, opened 2022-06-12
Adapt existing Haskell Stack example for Windows

mpilgrem:
The default `STACK_ROOT` is `~/.stack` only on Unix-like operating systems. On Windows, the default is `%APPDATA%/stack` (usually `%HOME%\AppData\Roaming\stack`).

On Unix-like OSs, Stack stores GHC and other tools in a `programs` directory in the `STACK_ROOT`. On Windows, Stack stores those tools and MSYS2 in `%LOCALAPPDATA%\Programs\stack` (usually `%HOME%\AppData\Local\Programs\stack`).

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -223,11 +223,34 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
 
 ## Haskell - Stack
 
+### Linux or macOS
+
 ```yaml
 - uses: actions/cache@v3
   name: Cache ~/.stack
   with:
     path: ~/.stack
+    key: ${{ runner.os }}-stack-global-${{ hashFiles('stack.yaml') }}-${{ hashFiles('package.yaml') }}
+    restore-keys: |
+      ${{ runner.os }}-stack-global-
+- uses: actions/cache@v3
+  name: Cache .stack-work
+  with:
+    path: .stack-work
+    key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ hashFiles('package.yaml') }}-${{ hashFiles('**/*.hs') }}
+    restore-keys: |
+      ${{ runner.os }}-stack-work-
+```
+
+### Windows
+
+```yaml
+- uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


Phantsure:
@mpilgrem `%HOME%\AppData\Roaming\stack` is stack root? Their docs website seem to mention just one path for two things mentioned by you on windows. Here: https://docs.haskellstack.org/en/stable/GUIDE/#path

mpilgrem:
@Phantsure, if environment variable `STACK_ROOT` is not set, `%HOME%\AppData\Roaming\stack` is the stack root. The location of GHC or other tools is documented here: https://docs.haskellstack.org/en/stable/faq/#can-i-change-stacks-default-temporary-directory. On Unix-like operating systems, that location is a subdirectory of the stack root. On Windows, it is outside of the stack root.

examples.md
```diff
@@ -225,12 +225,23 @@ We cache the elements of the Cabal store separately, as the entirety of `~/.caba
...
 ```yaml
 - uses: actions/cache@v3
+  if: startsWith(runner.os, 'Linux') || startsWith(runner.os, 'macOS')
   name: Cache ~/.stack
   with:
     path: ~/.stack
     key: ${{ runner.os }}-stack-global-${{ hashFiles('stack.yaml') }}-${{ hashFiles('package.yaml') }}
     restore-keys: |
       ${{ runner.os }}-stack-global-
+- uses: actions/cache@v3
```
Phantsure, line 235: Can we make this separate for windows and unix systems, like done [here](https://github.com/actions/cache/pull/819/files#diff-cfb2db6f153a4484d9bbd408f3227a264dce68fadd747cecdc30bda012b4d8cdR195)?

mpilgrem: @Phantsure, can I double-check the intended URL behind the 'here'?

mpilgrem: @Phantsure, I'll revise the PR to reflect that.

Phantsure (approved):

PR #834, opened 2022-06-24
Consuming 3.0 actions/cache

tiwarishub:


src/save.ts
```diff
@@ -44,21 +44,14 @@ async function run(): Promise<void> {
...
-                throw error;
-            } else if (typedError.name === cache.ReserveCacheError.name) {
-                core.info(typedError.message);
-            } else {
-                utils.logWarning(typedError.message);
-            }
+        const cacheId = await cache.saveCache(cachePaths, primaryKey, {
+            uploadChunkSize: utils.getInputAsInt(Inputs.UploadChunkSize)
+        });
+        if (cacheId == -1) {
```
vsvipul, line 50: Should we do - 
```
if (cacheId != -1) {
    core.info(`Cache saved with key: ${primaryKey}`);
}
```

vsvipul (approved):
LGTM

Issue #840, opened 2022-06-30
Explicitly document how to have cache always updated.


krpatter-intc:
Add to Readme the best-known-method for making sure you save your cache information

It would be worth spelling out in the README the details mentioned here:

https://github.com/actions/cache/issues/342#issuecomment-673371329

for how to support the ability to always work with and then update a cache

Jamie-:
To add to this, the docs say:
> 3) If the job completes successfully, the action automatically creates a new cache with the contents of the path directory.

https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

It does not specify that the cache is *only* stored on a cache miss.
Confirmed in the code here:
https://github.com/actions/cache/blob/main/src/save.ts#L36-L41

If the code is stay as-is, the main docs need to be updated to clarify this.

bishal-pdMSFT:
Agree, we can probably update both README as well as the docs.

PR #847, opened 2022-07-04
Adding "always-save" flag to make #840 easier

martinheidegger:
The [current recommendation](https://github.com/actions/cache/issues/342#issuecomment-673371329) has been a bit difficult for me to understand _(and it seems i [am not the only one](https://github.com/actions/cache/issues/840)?)_.

In this PR I implemented a `always-save` flag that makes this easier for me to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 * `key` - An explicit key for restoring and saving the cache
 * `restore-keys` - An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note
 `cache-hit` returns false in this case.
+* `always-save` - Flag indicating that it the cache should always be written
 
 ### Outputs
 
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -293,6 +293,41 @@ test("restore with restore keys and no cache found", async () => {
     );
 });
 
+test("restore with always-save", async () => {
+    const path = "node_modules";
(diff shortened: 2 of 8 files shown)


Bariscankibarcekici:
thank you very much

jacksongoode:
Hoping this is a major priority!

Luc45:
I desperately need this feature.

My use case:

- I need to access large amounts of data that is hosted in another git repository.
- I clone the repository (if it doesn't exist) or do a git pull, if it exists, essentially keeping my cache up-to-date with the repository.
- With the current implementation, it assumes that the cache is immutable, so I have to invalidate the entire cache every time there's a change in the other repository, essentially running `git clone` on every cache invalidation (I'm using as cache key the commit SHA of HEAD in that repo). The behavior I would like is to have a persistent cache and run `git pull` in it and save the resulting cache, therefore an `always-save` option would work very well for me.

ErcinDedeoglu:
LGTM.

@bishal-pdMSFT are you going to review or give an answer?

kotewar:
We have released two new actions [actions/cache/restore](https://github.com/actions/cache/tree/main/restore) and [actions/cache/save](https://github.com/actions/cache/tree/main/save) that take care of this case as well. You can now choose to run `save` action with condition `if: always()` and it will always save the cache for you for the given key. 

We had released a beta to get user's feedback, discussion post [here](https://github.com/actions/cache/discussions/1020).
And release post [here](https://github.com/actions/cache/discussions/1028).

Closing this as the use case as been implemented.

PR #893, opened 2022-08-09
Add option to skip uploading of cache

edoardopirovano:
Hello Actions team! We're currently encountering an issue where we'd like a nightly job to prepare a build cache for PR runs to use by building a big build target so that it gets populated with as much as possible and then individual PR checks can use what they need. So, our cache key would be today's date. However, the problem is that if a PR check completes before the nightly job, it will win the race to upload a cache, and we'll end up with an under-populated cache for the rest of the day. Thus, we'd like to disable cache uploading except on the nightly job, which is currently not supported by this Action. This PR adds a simple extra boolean input so that we can do that.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -14,6 +14,10 @@ inputs:
   upload-chunk-size:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
+  skip-upload:
+    description: 'Set this to true to skip uploading the cache at the end even if we got a miss'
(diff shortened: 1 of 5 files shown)


edoardopirovano:
Closed in favour of https://github.com/actions/cache/pull/571 which does the same thing. @aparna-ravindra any chance we could get that one merged? It has been open over a year.

Issue #171, opened 2020-02-02
Support writing to the cache even on a cache hit


Lyeeedar:
I'm currently trying to cache some resource build intermediates, with the source files coming from all over the repository. As I do not know exactly what files may be source files, I am setting the cache key to be the same one each time:

- name: Restore resource cache
      uses: actions/cache@v1
      with:
        path: my/resource/caches
        key: resourceCache-version1

However the cache isn't being updated at the end of the action, because of the cache being hit.
Ideally the action would have a parameter like uploadIfCacheChanged: true, which would make the cache check the hash of the cache contents at the start vs the hash of the contents at the end, and upload the new cache if these are different.

I have a workaround, which is to hash the entire project contents for the key each time (which should never exist), and rely on the restore-key to actually fetch the last contents:

- name: Restore resource cache
      uses: actions/cache@v1
      with:
        path: my/resource/caches
        key: resourceCache-${{ hashFiles('**/*') }}"
        restore-keys: resourceCache-

However this means that I am constantly uploading 'dead' caches to the cache store, filling it up and potentially removing ones that I want to keep. It also uploads even if the cache contents hasn't changed. It'd be much better if it just overwrote the same one each time, and only if it is different.

biner:
when i hacked the "utils.isExactKeyMatch" , 

>         
       if (utils.isExactKeyMatch(primaryKey, state)) {
            core.info(
                `Cache hit occurred on the primary key ${primaryKey}, not saving cache.`
            );
            core.info(
                `But... I want always saving cache (condition is empty)`
            );
            //return;
        }

        core.debug("Reserving Cache");
        const cacheId = await cacheHttpClient.reserveCache(primaryKey);



the [reserveCache] api throw error
`[warning]Cache already exists. Scope: refs/heads/master, Key: test-Linux-43307440, Version: (null)`


maybe we need one destory api ， or support vesion update

inikolaev:
I have a similar request - sometimes I end-up with an empty cache if build fails and cache is not populated and there's no way to fix that unless I make changes to the hashed file, which is `Pipfile.lock` in my case.

christhalinger:
When using `ccache` for C or C++ code we would need the same feature as the cache always needs to be updated.

christhalinger:
Is this a duplicate of #109?

Flamefire:
[...]

PR #896, opened 2022-08-14
Add readme section for good practices

bishal-pdMSFT:
There are certain solutions which allow to use cache for some interesting use cases (e.g. #171). Adding a section to highlight such use cases and potential good way to solve them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,23 @@ jobs:
         run: ./generate-primes -d prime-numbers
 ```
 
+## Good practices
+These are some of the good practices which can be used to fulfill certain requirements and are not necessarily the only or the recommended solution.
+
+- **Update a cache** - A cache today is immutable and cannot be updated. But some use cases require the cache to be saved even though there was a "hit" during restore. To do so, use a `key` which is unique for every run and use `restore-keys` to restore the nearest cache. For example:
+  ```
+      - name: update cache on every commit
+        uses: actions/cache@v3
+        with:
+          path: prime-numbers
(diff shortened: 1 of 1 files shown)


bishal-pdMSFT:
#840

README.md
```diff
@@ -219,6 +219,23 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: prime-numbers
+          key: primes-${{ runner.os }}-${{ github.run_id }} # Can use time based key as well
+          restore-keys: |
+            primes-${{ runner.os }}
+  ```          
+  Please note that this will create a new cache on every run and hence will consume the cache [quota](#cache-limits).
+  
+- **Use cache across feature branches** - Reusing cache across feature branches is not allowed today to provide cache [isolation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). However if both feature branches are from same base branch, a good way to achieve this is to ensure that the base branch has a cache. This cache will then be consumable by both feature branches.
```
kotewar, line 237: > same base branch

Shall we specifically mention that the base branch here is only the default branch? I think this won't work if the base branch is a feature branch.

bishal-pdMSFT: @kotewar this is what docs say

>Also, if the branch feature-b has the base branch feature-a, a workflow triggered on feature-b would have access to caches created in the default branch (main), feature-a, and feature-b

kotewar (approved):
LGTM

PR #903, opened 2022-08-22
Sort inputs and remove whitespaces after exclamation

### LINKED ISSUE #345: Normalize path in version hash

vsvipul:
Fixes #345.

Opened as a replacement PR for https://github.com/actions/cache/pull/473 to check failing tests and fix those.


damwe:
make a commit

Phantsure (approved):
Looks good to merge

damwe:
👍

PR #911, opened 2022-08-25
Remove .md files skip condition from test workflow

vsvipul:
Remove .md files skip condition from test workflow

This is required as we have made the test workflow checks required. So, they'll need to executed for all PRs. Also for PRs that include .md file changes.


pdotl (approved):
It would be great if you could update the PR description for the reasons for remove .md files from the test workflow. Changes LGTM.

pdotl (approved):

PR #912, opened 2022-08-25
Update README to include zstd workaround for windows.

pdotl:
This PR contains the following changes:

1. Move `Cache segment restore timeout` section to `Know practices and workarounds`
2. Add `yaml` formatting for actions code blocks
3. Add zstd workaround for windows and zstd requirement for self-hosted windows runners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,10 +248,10 @@ A cache today is immutable and cannot be updated. But some use cases require the
 #### Use cache across feature branches
 Reusing cache across feature branches is not allowed today to provide cache [isolation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). However if both feature branches are from the default branch, a good way to achieve this is to ensure that the default branch has a cache. This cache will then be consumable by both feature branches.
 
-#### Improving cache restore performance on Windows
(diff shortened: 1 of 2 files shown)


wrexbe:
This doesn't fix it. Even in your own test, this doesn't fix it
https://github.com/pdotl/sample-cache-demo/runs/7959974880?check_suite_focus=true
The Cache Python virtualinv step on Ubuntu took 2 seconds, and on Windows it took 11 seconds.
That is more then 5 times slower on windows, even with a tiny cache.

pdotl:
Hi @wrexbe, I agree. This workaround is being used by customers vs using `gzip` which is even slower. We are simply documenting the same. We are looking into improving windows performance for cache by looking at several alternatives.

README.md
```diff
@@ -223,18 +223,18 @@ jobs:
         if: steps.cache-primes.outputs.cache-hit != 'true'
         run: ./generate-primes -d prime-numbers
 ```
-## Cache segment restore timeout
 
+## Known practices and workarounds
```
Phantsure, line 227: Would it make sense to create separate file for this?

pdotl: I think it would yes. We can move it to another file.

Phantsure (approved):
LGTM

PR #931, opened 2022-09-22
docs: Indicate that `restore-keys` uses prefix in README

ELHoussineT:
I think we should state here that `restore-keys` supports prefixes as it is mentioned [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key).


vsvipul:
I'm not sure if this is required. The line `An ordered list of keys to use for restoring stale cache if no cache hit occurred for key` already signifies what you are trying to add.

ELHoussineT:
@vsvipul "list of keys" does not disclose that the keys can be prefixes to the keys.

vsvipul:
Something like `An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key`

ELHoussineT:
@vsvipul good idea.

vsvipul (approved):
lgtm

vsvipul:
Thank you for contributing 🎉

Issue #941, opened 2022-09-30
3.0.9 release is sorting restore-keys alphabetically which dramatically changes behaviour


eastlondoner:
This commit changes the restore-keys behaviour so that the order of the keys provided in the workflow is no longer respected when deciding what key to restore.

https://github.com/actions/cache/commit/204fa1ab69e643b91bf310c47cbc10fa2843d7fd#diff-79e9c65d7bc0c52a974cc3ca32be05e5fd06d175bfa85edbcef3af6df03ffe8dR150

This is a significant change in the behaviour and I think it's a bug

vsvipul:
@eastlondoner Agreed. This slipped through and should have only been done with paths in the first place. I'll look into this.

PR #942, opened 2022-10-03
Remove sort logic from inputs

vsvipul:
Fixes https://github.com/actions/cache/issues/941.

This bug was introduced with this commit https://github.com/actions/cache/pull/903/files and also sorts restore-keys which we don't want. Initial intention was to sort paths only.

This PR removes sorting completely. We can discuss later if we want to sort paths separately in save.ts / restore.ts .

After merging this, we'll need to release a new cache version immediately to fix this bug.

### DIFF
--- a/__tests__/actionUtils.test.ts
+++ b/__tests__/actionUtils.test.ts
@@ -215,23 +215,6 @@ test("getInputAsArray handles empty lines correctly", () => {
     expect(actionUtils.getInputAsArray("foo")).toEqual(["bar", "baz"]);
 });
 
-test("getInputAsArray sorts files correctly", () => {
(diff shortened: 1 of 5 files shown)


Phantsure:
I thought the previous PR was created to sort the files to be cached. Why remove that sort?

vsvipul:
@Phantsure That can be done in the save.ts / restore.ts file separately only for the path array, but not in the getInput fxn. Should I add that change as well ?

vsvipul:
@Phantsure  Then it's good to just remove sort here and merge this today as a fix.

Phantsure (approved):
makes sense

Issue #953, opened 2022-10-12
`save-state` and `set-output` are deprecated


zacharyburnett:
Running this action results in the following warnings:
```
Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
https://github.com/spacetelescope/stenv/actions/runs/3234699238/jobs/5298298424#step:14:20

```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
https://github.com/spacetelescope/stenv/actions/runs/3234699238/jobs/5298298424#step:14:22

jelmerk:
Same

rjkerrison:
I believe this will be fixed once all code is using version 1.10.0 of `@actions/core`. [Bumping to version 1.10.0 of @actions/core](https://github.com/actions/cache/commit/b9c8bfe4426d1b96548309ee742254ebf1a4f21c#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R27) is in `main` but not yet in a release version. Unfortunately `@actions/cache` has its own dependency on `@actions/core`, which is [currently at version 1.2.6](https://github.com/actions/toolkit/blob/1bc93f3cdf638ff423193f992119be08b7bb0666/packages/cache/package.json#L40).

fingolfin:
Actually this repo also uses `set-output`, and recommends it in the README and an example:

<img width="567" alt="Screenshot 2022-10-12 at 22 13 04" src="https://user-images.githubusercontent.com/241512/195438584-905f5a49-b085-48b1-a89b-56539c798736.png">

oliversalzburg:
There are also several `set-output` usages on https://github.com/actions/cache/blob/main/examples.md

PR #954, opened 2022-10-12
replace deprecated set-output command

netanelrabinowitz:
Replace all reference to deprecated `set-output` command `::set-output name= ...` in the README, the examples file
and the`Tests` workflow. #953


bishal-pdMSFT:
The [docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) don't seem to suggest such difference. @rentziass is this difference in behaviour across Windows and Linux expected?

> Previous Test run failed because the different syntax needed for windows to read the `GITHUB_OUTPUT` so instead of introducing a new step, something like the snippet bellow, I just evaluate the npm cache directory directly in the `path` input
> 
> ```
> - name: Determine npm cache directory windows
>     if: runner.os == 'Windows'
>     run: echo "dir=$(npm config get cache)" >> $env:GITHUB_OUTPUT
> - name: Determine npm cache directory non-windows
>     if: runner.os != 'Windows'
>     run: echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
> ```

rentziass:
Hey @bishal-pdMSFT 👋  yeah it's there but you need to switch from Bash to PowerShell at the very top of the page :)

bishal-pdMSFT:
>you need to switch from Bash to PowerShell at the very top of the page

@rentziass at top of page?? Do you mean to use `pwsh` as shell?

heddn:
The composer docs work correctly.

TheNorthMemory:
Hi there, another choice via [here](https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs)

> Use `defaults` to create a `map` of default settings that will apply to all jobs in the workflow. 

Placed the following `yml` for the workflow,

```yaml
defaults:
  run:
    shell: bash
```

An full example can be found [here](https://github.com/TheNorthMemory/xml/blob/main/.github/workflows/ci.yml)

Hopes it may helpful.

examples.md
```diff
@@ -316,7 +316,7 @@ If using `npm config` to retrieve the cache directory, ensure you run [actions/s
 - name: Get npm cache directory
   id: npm-cache-dir
   run: |
-    echo "::set-output name=dir::$(npm config get cache)"
+    echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
```
voxpelli, line 319: I sadly don't think this works on Windows

.github/workflows/workflow.yml
```diff
@@ -25,14 +25,10 @@ jobs:
...
         node-version: 16.x
-    - name: Determine npm cache directory
-      id: npm-cache
-      run: |
-        echo "::set-output name=dir::$(npm config get cache)"
     - name: Restore npm cache
       uses: actions/cache@v3
       with:
-        path: ${{ steps.npm-cache.outputs.dir }}
+        path: $(npm config get cache)
```
voxpelli, line 31: This seems to give the following error in the action:

> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

Togopeva13:
> Previous Test run failed because the different syntax needed for windows to read the `GITHUB_OUTPUT` so instead of introducing a new step, something like the snippet bellow, I just evaluate the npm cache directory directly in the `path` input
> 
> ```
> - name: Determine npm cache directory windows
>     if: runner.os == 'Windows'
>     run: echo "dir=$(npm config get cache)" >> $env:GITHUB_OUTPUT
> - name: Determine npm cache directory non-windows
>     if: runner.os != 'Windows'
>     run: echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
> ```

PR #959, opened 2022-10-18
fix wrong timeout env var key in README.md

walterddr:
according to https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout the key should be `SEGMENT_DOWNLOAD_TIMEOUT_MINS` not `SEGMENT_DOWNLOAD_TIMEOUT_MIN`


aparna-ravindra (approved):

aparna-ravindra:
Thank you for your contribution! 💚

PR #961, opened 2022-10-18
Update README with clearer info about cache-hit and its value

kotewar:
This PR is being raised to clarify the output `cache-hit` on partial and full match of the key found in `key`, `restore-keys` of the workflow. This is being done as we have [PRs](https://github.com/actions/cache/pull/895) being raised to remove restore keys. 

Updated rendered [file](https://github.com/actions/cache/blob/14f2d18ea2a8de1085ccf3aedfeb86ef1f4b7bdd/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 
 * `cache-hit` - A boolean value to indicate an exact match was found for the key. 
 
-> Note: `cache-hit` will be set to `true` only when cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`. It is recommended to install the missing/updated dependencies in case of a partial key match when the key is dependent on the `hash` of the package file.
+> Note: `cache-hit` will be set to `true` only when cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`.
 
 See [Skipping steps based on cache-hit](#skipping-steps-based-on-cache-hit) for info on using this output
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,17 +34,18 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
...
 #### Environment Variables
 * `SEGMENT_DOWNLOAD_TIMEOUT_MINS` - Segment download timeout (in minutes, default `60`) to abort download of the segment if not completed in the defined number of minutes. [Read more](#cache-segment-restore-timeout)
 
 ### Outputs
 
-* `cache-hit` - A boolean value to indicate an exact match was found for the key
+* `cache-hit` - A boolean value to indicate an exact match was found for the key. 
 
-> See [Skipping steps based on cache-hit](#Skipping-steps-based-on-cache-hit) for info on using this output
+> Note: `cache-hit` will be set to `true` only when cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`. It is recommended to install the missing/updated dependencies in case of a partial key match when the key is dependent on the `hash` of the package file.
```
pdotl, line 46: IMHO, we should not be making recommendations at this particular place in the file. With that said, this is quite a specific recommendation for the cache use-case of caching dependencies while using `hash` of package file as key. So we might have to make it more generic if we are to add it here.

aparna-ravindra: +1

kotewar: I just moved this note from `Inputs` section to `Outputs` section with more detailed info 🙈 

What would be a better place to put this info then? It has to be visible to users right away so that people use actions/cache correctly and have correct expectations regarding the restore from the same.

README.md
```diff
@@ -34,17 +34,18 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
 * `key` - An explicit key for restoring and saving the cache
-* `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key. Note
-`cache-hit` returns false in this case.
```
anuragc617, line 38: Isn't the cache-hit false when cache is restored via `restore-keys`?

README.md
pdotl: I believe the 2nd line of recommendation should go in the skipping steps we are linking. As there we are already giving an example, you can easily make the above recommendation based on that example there.

kotewar: Please review now @pdotl

pdotl (approved):
LGTM.

PR #964, opened 2022-10-22
Configurable save cache via environment variable

### LINKED ISSUE #92: Configurable save cache on failure

### LINKED ISSUE #334: Support for Read Only Cache / Configurable Cache Saving

### LINKED ISSUE #350: Read only cache?

rhaschke:
Following the suggestion in https://github.com/actions/cache/issues/92#issuecomment-754108243, this implements conditional execution of the post-step to save the cache.
If the environment variable `GHA_CACHE_SAVE` is set to `always`, the `post-if` condition will evaluate to true.
If it is set to `never`, the condition will evaluate to false, thus not saving the cache.

Fixes #350
Fixes #334
Fixes #92

In contrast to #474, which used input variables and an implementation of the condition evaluation in source code, this PR implements the behavior using an appropriate `post-if` condition, thus nicely visualizing the execution status in the log UI.
Actually, an implementation using input variables would be more appealing, but input variables are currently silently ignored in `post-if` condition expressions.
Compared to #489, this PR not only allows skipping the post-step but also enforcing it.

simple test results can be found here:
- [default + success()](https://github.com/rhaschke/test/actions/runs/3303219650)
- [default + failure()](https://github.com/rhaschke/test/actions/runs/3303224615)
- [always](https://github.com/rhaschke/test/actions/runs/3303231700)
- [never](https://github.com/rhaschke/test/actions/runs/3303160629)

Unfortunately, I'm not proficient in JavaScript or TypeScript. Thus I cannot provide unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 #### Environment Variables
 * `SEGMENT_DOWNLOAD_TIMEOUT_MINS` - Segment download timeout (in minutes, default `60`) to abort download of the segment if not completed in the defined number of minutes. [Read more](#cache-segment-restore-timeout)
 
+* `GHA_CACHE_SAVE` - Controls when to save cache. By default, a new cache is created if the job completes successfully. If this variable is set to `always`, the cache is saved also on job failure. If set to `never`, the cache is never saved, i.e. realizing read-only cache.
+
 ### Outputs
 
 * `cache-hit` - A boolean value to indicate an exact match was found for the key
@@ -80,7 +82,8 @@ jobs:
       run: /primes.sh -d prime-numbers
 ```
 
-> Note: You must use the `cache` action in your workflow before you need to use the files that might be restored from the cache. If the provided `key` matches an existing cache, a new cache is not created and if the provided `key` doesn't match an existing cache, a new cache is automatically created provided the job completes successfully.
+> Note: You must use the `cache` action in your workflow _before_ you need to use the files that might be restored from the cache. If the provided `key` matches an existing cache exactly, a new cache is not created. Include `${{github.run_id}}` as the last component of your `key` to save the cache in each workflow run.
+If the provided `key` doesn't match an existing cache exactly, a new cache is automatically created provided the job completes successfully. This behavior can be changed with the environment variable `GHA_CACHE_SAVE` ([see above](#environment-variables)).
 
 ## Implementation Examples
 
--- a/action.yml
+++ b/action.yml
@@ -21,7 +21,9 @@ runs:
   using: 'node16'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: 'success()'
+  post-if: |
+    ((env.GHA_CACHE_SAVE == 'always') && (success() || failure())) ||
+    ((env.GHA_CACHE_SAVE != 'never') && success())
 branding:
   icon: 'archive'
   color: 'gray-dark'


eyal0:
All caches are using the same environment variable so you cannot use multiple caches and have them controlled separately.

This seems like it could be a problem, right,

rhaschke:
> All caches are using the same environment variable so you cannot use multiple caches and have them controlled separately.

That's wrong. You can define different environment variables for each step:
https://github.com/rhaschke/test/actions/runs/3305095368/jobs/5454935646
```
jobs:
  default:
    steps:
      - name: Cache1
        uses: rhaschke/cache@main
        with:
          path: data1
          key: ${{ runner.os }}-data1-${{ github.run_id }}
          restore-keys: ${{ runner.os }}-data1-

      - name: Cache2
        uses: rhaschke/cache@main
        with:
          path: data2
          key: ${{ runner.os }}-data2-${{ github.run_id }}
          restore-keys: ${{ runner.os }}-data2-
        env:
          GHA_CACHE_SAVE: 'always'
```

eyal0:
Ibut.the environment variable is evaluated at the end in the post action, not at the time that you declare the cache, right?

Anyway, your link didn't work for me. Not sure why.

rhaschke:
> The environment variable is evaluated at the end in the post action, not at the time that you declare the cache, right?

Yes. I think the per-step environment variables are defined for each execution of the step - on top of the global ones. [According to the docs, this even holds for `GITHUB_ENV`](https://docs.github.com/en/actions/learn-github-actions/environment-variables): _"This file is unique to the current step and changes for each step in a job."_

> Anyway, your link didn't work for me.

Thanks for the hint. The repo was private. Should work now.

eyal0:
Is it possible with this change to decide whether or not to save the cache during the execution of CI?

For example, I want to load the cache, examine the version of a library that was just loaded from the cache, and then download from the web the latest version number. Then I want to compare the two and decide based on that whether or not to make a new cache and save it.

Could I do that?

rhaschke:
As far as I understand the docs, that's not possible as the `GITHUB_ENV` files are separate for each step. But, give it a try!
I guess to enable this, you would need to use an output variable in the `post-if` condition...

To enable such flexible behavior you would need to allow a custom condition to be provided e.g. via inputs and then evaluate that expression programmatically in the `post` step.
However, I just focused on the very basic behavior many people requested for years.

eyal0:
Okay. That was the behavior that I tried to get with my PR.

It would be interesting to know what the community needs are.

I feel like probably none of it will get merged. We'll see!

rhaschke:
> That was the behavior that I tried to get with my PR.

I guess you are referring to #498? I didn't see that one before. Thanks for indirectly pointing it out. Indeed, it implements slightly different behavior.

> I feel like probably none of it will get merged. We'll see!

It's really a pity that this repo has so many long-standing PRs and issues.

anuragc617:
@rhaschke thanks for working on this PR. Currently, we are working on providing more granular controls. This PR comes under that feature, so I will block this PR until we finalise the design.

rhaschke:
> Currently, we are working on providing more granular controls. This PR comes under that feature.

Great to hear. I'm very much looking forward to this.

kotewar:
Hey all, 👋🏽 

We have created a [discussion](https://github.com/actions/cache/discussions/1020) with a proposal that we feel will solve this problem, do let us know your feedback, thanks 😊

kotewar:
We have released two new actions `restore` and `save` that we believe solves this problem. Using `if: always()` condition with the save action achieves the desired outcome, hence closing this PR.

PR #995, opened 2022-11-22
add support for read-only cache

erik-krogh:
## Description
Add support for not saving the cache.  

Fixes https://github.com/actions/cache/issues/350
Fixes https://github.com/actions/cache/issues/334
Fixes https://github.com/actions/cache/issues/92 

Similar to both https://github.com/actions/cache/pull/489 and https://github.com/actions/cache/pull/964, but here it's implemented as an input to the workflow instead of an environment variable.  

## Motivation and Context
I needed the behavior for this PR: https://github.com/github/codeql/pull/11362. 

In `github/codeql` we have some code that gets faster by using a cache.  
But that code can take [some shortcuts](https://codeql.github.com/docs/codeql-cli/manual/query-compile/#cmdoption-codeql-query-compile-n) if it doesn't have to save a new cache.
So on PRs we essentially run in a "read-only" mode, where we don't expect any cache to be populated.  
And we make sure that the cache is populated on pushes to `main`.  

Our current workaround is to use a key based on the current commit SHA, which no other run will hit.  
But that is currently filling up the cache space in the repo: https://github.com/github/codeql/actions/caches 

## How Has This Been Tested?
- I added a test-case.  
- It's been concretely tested on a workflow here: https://github.com/github/codeql/pull/11362 

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. **(I'll do that if the PR looks OK).** 

### DIFF
--- a/__tests__/save.test.ts
+++ b/__tests__/save.test.ts
@@ -35,6 +35,14 @@ beforeAll(() => {
         }
     );
 
+    jest.spyOn(actionUtils, "getInputAsBoolean").mockImplementation(
+        (name, options) => {
+            return jest
+                .requireActual("../src/utils/actionUtils")
+                .getInputAsBoolean(name, options);
+        }
+    );
+
     jest.spyOn(actionUtils, "isExactKeyMatch").mockImplementation(
         (key, cacheResult) => {
             return jest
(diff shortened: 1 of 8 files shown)


erik-krogh:
> Thanks @erik-krogh for your pull request, we are working on some similar asks (restore only, save only) based on the frequency of these asks.
> Would you mind if I close this PR as you can expect us to release similar functionalities in the coming days?

Sounds great 👍 

I'll watch releases in this repo, and change my workflows to your implementation when it arrives.

Issue #955, opened 2022-10-13
optional action failure on cache miss


deitch:
Can we have an option to fail on cache miss? The current behaviour is:

* cache hit: restore, exit successfully
* cache miss: do not restore, exit successfully

Much of the time, the above makes sense. But there can be workflows where you might want to fail if cache miss. For example, 2 jobs that depend on each other and use cache to move ephemeral artifacts around. We don't want them as formal artifacts, but a cache miss should cause the job to stop.

Right now, you can do it with a little convolution:

```yaml
      - name: update from cache
        id: cache
        uses: actions/cache@v3
        with:
          path: foo
          key: some-key
      - name: Fail if cache miss
        uses: actions/github-script@v6
        if: steps.cache.outputs.cache-hit != 'true'
        with:
          script: core.setFailed('Cache hit failed')
```

That gets unwieldy, especially if it has to be done in a few jobs. The following would be much easier:

```yaml
      - name: update from cache
        id: cache
        uses: actions/cache@v3
        with:
          path: foo
          key: some-key
          fail_on_miss: true      # this, or some similar key
```

pdotl:
Hi there @deitch, we are planning some changes for users to have granular controls over cache. More info to follow once we have the plan finalised. Thank you for creating this issue, will keep it open for tracking currently.

deitch:
That sounds interesting. I rather am looking forward to that.

kotewar:
Hi @deitch ,

This looks more like a registry (npm registry/maven repository) pattern, where if the dependency is present in the registry with a required version, the build would succeed, else it would fail saying the version doesn't exist.

Is that what you are looking for?

deitch:
Not quite @kotewar , although now that you describe it, the parallels are obvious.

In my use case, we have a long build process that creates output artifacts. These are used in multiple downstream jobs. The build is expensive, so we do it once, then cache, then pull the cache for the many downstream jobs. If they are missing, that is an issue that I want to catch and fail the job there and then, rather than waiting for some "missing artifact" error in the actual job execution.

```
build job ----> jobA
          ----> jobB
          ----> jobC
          ----> jobD
          ----> jobE
          ----> jobF
          ----> jobG
```

I looked at using [workflow artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) for this, which would work, but it is a mismatch. These are temporary outputs, useful only within the scope of the workflow. When all of the downstream jobs are done, they should go away.

[...]

Issue #852, opened 2022-07-10
New option to create `path`


szepeviktor:
Currently `mkdir -p $cache_path` should be issued to create directory specified in `path`.
`cache` does not create the `path` directory if it doesn't exist.
https://github.com/szepeviktor/Valinor/runs/7272413993?check_suite_focus=true

Please consider adding a `"create_directory": true` option to create directory specified in `path`.
Thank you!!

kotewar:
Hey @szepeviktor 👋🏽 
The directory creation should happen at build time, and should be created by the build program itself.
Can you please specify why you would want a new directory to be specifically created?

kotewar:
@szepeviktor, would you like to share the use case for this suggestion so that we can prioritise it?

szepeviktor:
@kotewar On the very first run the cache is empty and many tools do not autocreate directories, mainly sub-subdirectories.
And these tools are failing because of the non-existent directory structure.

kotewar:
Thanks for bringing this up @szepeviktor. 

Can you please share these tools as well so that we can use the same for testing the behaviour and make the necessary fixes?

szepeviktor:
Okay. I respond within a day.

szepeviktor:
https://github.com/CuyZ/Valinor/blob/master/infection.json.dist#L8
Using a **non-existent** temporary directory for `infection/infection` causes an error.

kotewar:
How about using `shell:bash` to do this job for you?

Currently we haven't received much requests in this area, but this can be achieved using custom [shell commands](https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow) to achieve this folder creation.

szepeviktor:
Thank you.
I'm into solutions, not work-arounds.

fquinner:
Vcpkg does this too with its binary cache. Will only use an existing directory.

    VCPKG_DEFAULT_BINARY_CACHE

PR #1006, opened 2022-11-30
Actions Cache Granular Control Implementation

kotewar:
## Description
To address the granular control asks from our users, we have tackled the 4 use cases in this PR.

We added two new actions
1. restore/action.yml - to only restore cache
2. save/action.yml - to only save cache
~3. Fail on cache miss~ (can be handled by workflow conditions)
~4. Save cache on failure~ (can be handled by workflow conditions using the new action)

## Motivation and Context

This will grant our users more control - especially in cases where they only want to restore or save cache explicitly.

## How Has This Been Tested?
Testing right now, will update with new workflows for the same.

NA

## Types of changes

- [x] Documentation (add or update README or docs) - will be done in subsequent PRs.

## Checklist:


restore/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  path:
+    description: 'A list of files, directories, and wildcard patterns to cache and restore'
+    required: true
+  key:
+    description: 'An explicit key for restoring and saving the cache'
+    required: true
+  restore-keys:
+    description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
+    required: false
+  exit-on-cache-miss:
```
bishal-pdMSFT, line 14: Since we already have an output for `cache-hit`, user can add a next step to check the output value and fail the workflow job. Do we really need a separate input?

restore/action.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Restore Cache'
+description: 'Restore Cache artifacts like dependencies and build outputs to improve workflow execution time'
+author: 'GitHub'
+inputs:
+  path:
+    description: 'A list of files, directories, and wildcard patterns to cache and restore'
```
bishal-pdMSFT, line 6: This is only for restore now.

restore/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    required: true
+  key:
+    description: 'An explicit key for restoring and saving the cache'
```
bishal-pdMSFT, line 9: Same here. This is only for restore.

save/action.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Save Cache'
+description: 'Save Cache artifacts like dependencies and build outputs to improve workflow execution time'
+author: 'GitHub'
+inputs:
+  path:
+    description: 'A list of files, directories, and wildcard patterns to cache and restore'
```
bishal-pdMSFT, line 6: This is only for cache and not restore.

save/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    required: true
+  key:
+    description: 'An explicit key for restoring and saving the cache'
```
bishal-pdMSFT, line 9: Same here.

src/save.ts
```diff
@@ -27,7 +27,9 @@ async function run(): Promise<void> {
         const state = utils.getCacheState();
 
         // Inputs are re-evaluted before the post action, so we want the original key used for restore
-        const primaryKey = core.getState(State.CachePrimaryKey);
+        const primaryKey =
+            core.getState(State.CachePrimaryKey) || core.getInput(Inputs.Key);
```
bishal-pdMSFT, line 31: If I am not wrong, state value will be used by cache action and input will be used by save action. If so, then this statement is can be confusing as it is mixing logic for two actions.
We should have a pattern such that such action specific logic is separated out in an action specific wrapper.

src/restore.ts
```diff
@@ -35,22 +35,46 @@ async function run(): Promise<void> {
             restoreKeys
         );
 
+        //Check if user wants to save cache despite of failure in any previous job
+        const saveCache = core.getInput(Inputs.SaveOnAnyFailure).toLowerCase();
```
bishal-pdMSFT, line 39: Now that we have two different actions, user can ensure this by calling `save` action explicitly and using a condition of `if: always()`. I think this is same as what David had mentioned in ADR comment.

bishal-pdMSFT: Essentially what I am saying is that the original `cache` action behavior remains same. User should use new `restore` and `save` action with right conditionals if they want to have better control like save on failure, or fail on cache miss etc

src/restore.ts
```diff
@@ -35,22 +35,46 @@ async function run(): Promise<void> {
...
+                `Exporting environment variable ${Variables.SaveCacheOnAnyFailure}`
+            );
+            core.exportVariable(Variables.SaveCacheOnAnyFailure, saveCache);
+            core.info(
+                `Input Variable ${Variables.SaveCacheOnAnyFailure} is set to true, the cache will be saved despite of any failure in the build.`
+            );
+        }
+
         if (!cacheKey) {
+            if (core.getInput(Inputs.FailOnCacheMiss).toLowerCase() == "true") {
```
bishal-pdMSFT, line 51: As mentioned in earlier comment, this can be achieved by running a separate step which fails based on the `cache-hit` output.
It would be better to start with simpler set of inputs and ask users to use outputs and conditionals to achieve their custom control. If we still see multiple asks, we can take this separately.

src/restore.ts
```diff
@@ -35,22 +35,46 @@ async function run(): Promise<void> {
...
                     primaryKey,
                     ...restoreKeys
                 ].join(", ")}`
             );
-
             return;
         }
-
         // Store the matched cache key
         utils.setCacheState(cacheKey);
```
bishal-pdMSFT, line 65: Wondering if we should not set this state (and the primary key state in L25) if this flow is for `restore` action. Since there is no post step, it does not make sense to set these states.

bishal-pdMSFT, line 65: Can't put a comment on L25 so putting here. The primary key should be set as output as well. This because user mostly would want to reuse same computed primary key in `save` step rather than recompute it again.

bishal-pdMSFT: One thing I am thinking is:
- If an output, set it irrespective of their granular actions or `cache` action
- If a state, set it only for `cache` action

src/save.ts
kotewar: cc - @tanuj077

src/restore.ts
kotewar: Makes sense.
This feature was asked by ~10 users (issues count + thumbs up count) https://github.com/actions/cache/issues/210, https://github.com/actions/cache/issues/955

So should we drop this from our release? As correctly mentioned above, it can be easily achieved by adding an extra step in the workflow. I'm good with dropping it.

restore/action.yml
kotewar: Forgot to reword it, I was thinking of something like `The list of files, directories, and wildcard patterns to restore cache that were used while saving it`

save/action.yml
kotewar: Fixed

save/action.yml
kotewar: Fixed

restore/action.yml
kotewar: Fixed

restore/action.yml
kotewar: Updated

src/restore.ts
kotewar: Which means the wrapper classes would take care of setting as an output or a state. Is that right?

src/restore.ts
bishal-pdMSFT: Then do we still need this input?

src/restore.ts
bishal-pdMSFT: Yes @kotewar

src/restore.ts
kotewar: Yes, `if: always()` should be used with `save` action.
and `Inputs.SaveOnAnyFailure` input should be used in the `cache` action input.

bishal-pdMSFT: Okay, so intention was not clear from the code. Another reason why we need to have separate wrappers for `restore`, `save` and `cache` actions otherwise it going to become unmanageable.

bishal-pdMSFT: Also, what if we do not provide this "save on failure" functionality in cache action and ask users to use the two new actions if they want this. IMO we should not change the behaviour of cache action much for such restore-save dependency related scenarios, and let the two new actions take care of such.
Thoughts?

src/restore.ts
bishal-pdMSFT: Even the non common inputs handling can be part of the wrapper.

src/restore.ts
kotewar: Oh okay, in short we only want to use the `if: always()` with `save` action and not change anything in the `cache` action.
But let's say there's some failure even before anything related to build or cache restore happens, for example the `checkout` action fails, in that case too we will execute the save action, which won't be a desired behaviour right?

src/restore.ts
kotewar: The only issue I see is, some users are not happy with workarounds, they expect the extra workflow steps to be handled by our action, for [example](https://github.com/actions/cache/issues/852#issuecomment-1323305728), one of the user didn't like the idea of using `shell` for creating folders.

src/restore.ts
bishal-pdMSFT: Good point. Can user use `if` conditional to run only if restore was success?
Something like `if: always() && steps.<restore-step>. outcome  == 'success'`? 
Not sure if `steps` context is available in `if`. Worth trying.

bishal-pdMSFT: Looks like it may be available, so definitely worth trying.

bishal-pdMSFT: Btw please note that I am not totally against these additional inputs but they do not feel like must-to-do. Hence better to keep things simple for now and add later if there are additional asks.

src/restore.ts
bishal-pdMSFT: If there are multiple asks after we ship, we would definitely think about doing it. Right now keeping simpler is easy as well.

bishal-pdMSFT:
> Effectively, to summarise @bishal-pdMSFT,
> 
> We can eliminate both `save-on-any-failure` (if we come to conclusion on this [comment](https://github.com/actions/cache/pull/1006#discussion_r1037359759)) and `exit-on-cache-miss` as these use cases can be achieved by using either the split actions or using output of our action.
> 
> So ultimately we will only ship two new actions, nothing else. Is that correct?

Yep, simpler is better! We can any day revisit these inputs based on feedback.

src/restoreOnly.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Outputs } from "./constants";
+import run from "./restoreImpl";
+import * as utils from "./utils/actionUtils";
+
+async function restoreOnly(): Promise<void> {
+    const cacheKey = await run();
+    if (cacheKey) {
+        // Store the matched cache key in output
+        core.setOutput(Outputs.Key, utils.getCacheState());
```
bishal-pdMSFT, line 11: Don't we need to set CacheHit output in this case?

bishal-pdMSFT: We do! as per restore action.yml

src/restore.ts
cdce8p: > We will still let those users achieve this by using the two actions in a certain way. So we are still solving their problem.

True, there is a usable workaround for it. The main appeal IMO would be that the action could provide a much better error message and reduce the boilerplate code required to get up and running. Especially if a workflow with 10+ jobs requires that the check is added in each one, it's just easy too miss or add a typo.

That all for a feature which would be fairly easy to implement.

restore/action.yml
```diff
@@ -0,0 +1,23 @@
...
+    required: true
+  restore-keys:
+    description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
+    required: false
+outputs:
+  cache-hit:
+    description: 'A boolean value to indicate an exact match was found for the primary key'
+runs:
+  using: 'node16'
+  main: '../dist/restore/index.js'
```
bishal-pdMSFT, line 19: Shouldn't it point to restoreOnly.js?

src/saveImpl.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        // If restore has stored a primary key in state, reuse that
+        // Else re-evaluate from inputs
+        const primaryKey =
+            stateProvider.getState(State.CachePrimaryKey) ||
+            core.getInput(Inputs.Key);
+
+        if (!primaryKey) {
+            utils.logWarning(`Error retrieving key from state.`);
```
bishal-pdMSFT, line 35: This is not an accurate error message anymore as save-only will not get it from state. Just say that `key cannot be empty`

src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export class StateProvider extends StateProviderBase {
+    //setOutput = core.setOutput;
+    setState = core.saveState;
+    getState = core.getState;
+}
+
+export class NullStateProvider extends StateProviderBase {
+    //setOutput = core.setOutput;
```
bishal-pdMSFT, line 38: remove the comments

src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
+    setState = (key: string, value: string) => {};
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getState = (key: string) => "";
+}
+
+export class StateProvider extends StateProviderBase {
+    //setOutput = core.setOutput;
```
bishal-pdMSFT, line 32: remove comments

src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
+import * as core from "@actions/core";
+
+import { State } from "./constants";
+
+export interface IStateProvider {
+    //setOutput(key: string, value: string): void;
```
bishal-pdMSFT, line 6: remove comments

__tests__/actionUtils.test.ts
```diff
@@ -79,83 +79,6 @@ test("isExactKeyMatch with same key and different casing returns true", () => {
...
-        return cacheKey;
-    });
-
-    const state = actionUtils.getCacheState();
-
-    expect(state).toEqual(cacheKey);
-
-    expect(getStateMock).toHaveBeenCalledWith(State.CacheMatchedKey);
-    expect(getStateMock).toHaveBeenCalledTimes(1);
-});
```
bishal-pdMSFT, line 157: These tests should move to tests for stateProvider

__tests__/restore.test.ts
```diff
@@ -116,7 +116,7 @@ test("restore on GHES with AC available ", async () => {
 
     expect(stateMock).toHaveBeenCalledWith("CACHE_KEY", key);
```
bishal-pdMSFT, line 117: We should have test for restoreOnly where it should be checked that setOutput has been called with CACHE_KEY

src/constants.ts
```diff
@@ -21,3 +23,8 @@ export enum Events {
 }
 
 export const RefKey = "GITHUB_REF";
+
+export const stateToOutputMap = new Map<string, string>([
```
bishal-pdMSFT, line 27: This can be moved to NullStateProvider as that is right owner/consumer for this map.

src/saveImpl.ts
```diff
@@ -0,0 +1,70 @@
...
+        const primaryKey =
+            stateProvider.getState(State.CachePrimaryKey) ||
+            core.getInput(Inputs.Key);
+
+        if (!primaryKey) {
+            if (stateProvider instanceof StateProvider) {
+                utils.logWarning(`Error retrieving key from state.`);
+            } else {
+                utils.logWarning(`Error retrieving key from input.`);
+            }
```
bishal-pdMSFT, line 39: Do we need this differentiation? Or should keep simple and say "key is not specified"

kotewar: Makes sense, I was just thinking to make the logs more clearer based on the use case but for now better to keep it simple. This has been implemented.

src/constants.ts
kotewar: Moved

src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
+import * as core from "@actions/core";
+
+import { State } from "./constants";
+
+export interface IStateProvider {
```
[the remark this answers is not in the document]
kotewar: Added

__tests__/restore.test.ts
```diff
@@ -317,8 +165,7 @@ test("restore with cache found for restore key", async () => {
...
```
bishal-pdMSFT, line 166: Should there be a validation for "CACHE_RESULT" state as well? I know this is an existing test but if there missing validations then we can add them as we are overall re-looking at all states and outputs in this PR.

__tests__/stateProvider.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        .mockImplementation((key, value) => {
+            return jest.requireActual("@actions/core").saveState(key, value);
+        });
+
+    const cacheMatchedKey = "node-cache";
+
+    const stateProvider: IStateProvider = new StateProvider();
+    stateProvider.setState("stateKey", "stateValue");
+    stateProvider.setState(State.CacheMatchedKey, cacheMatchedKey);
+    stateProvider.getState("stateKey");
```
bishal-pdMSFT, line 44: Should validate that the value fetched is same as value set in L42

__tests__/stateProvider.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        );
+    const setOutputMock = jest
+        .spyOn(core, "setOutput")
+        .mockImplementation((key, value) => {
+            return jest.requireActual("@actions/core").setOutput(key, value);
+        });
+    const cacheMatchedKey = "node-cache";
+    const nullStateProvider: IStateProvider = new NullStateProvider();
+    nullStateProvider.setState(State.CacheMatchedKey, "outputValue");
+    nullStateProvider.setState(State.CachePrimaryKey, cacheMatchedKey);
```
bishal-pdMSFT, line 65: Use `core.getOutput()` to verify that the value matches with those being set here.

bishal-pdMSFT: Ah but there is no getOutput(). Wondering how to validate this?

bishal-pdMSFT: Yeah the mock assert you are doing seems good enough.

restore/action.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Restore Only Cache'
```
bishal-pdMSFT, line 1:
```suggestion
name: 'Restore Cache'
```

restore/action.yml
```diff
@@ -0,0 +1,26 @@
...
+description: 'Restore Cache artifacts like dependencies and build outputs to improve workflow execution time'
+author: 'GitHub'
+inputs:
+  path:
+    description: 'The same list of files, directories, and wildcard patterns to restore cache that were used while saving it'
```
bishal-pdMSFT, line 6: Too detailed and may get confusing for an user.

```suggestion
    description: 'A list of files, directories, and wildcard patterns to restore'
```

restore/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    description: 'An explicit key for restoring the cache'
+    required: true
+  restore-keys:
+    description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
+    required: false
+outputs:
+  cache-hit:
+    description: 'A boolean value to indicate an exact match was found for the primary key'
+  cache-primary-key:
+    description: 'Cache primary key passed in the input to use in subsequent steps of the workflow'
```
bishal-pdMSFT, line 18:
```suggestion
    description: 'A resolved primary cache key for which cache match was attempted'
```

restore/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  cache-restore-key:
+    description: 'Cache key restored'
```
bishal-pdMSFT, line 20:
```suggestion
    description: 'Restore key which was used to restore the cache. It will not be set in case there was an exact match with primary key itself'
```

save/action.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Save Only Cache'
```
bishal-pdMSFT, line 1:
```suggestion
name: 'Save a cache'
```

__tests__/stateProvider.test.ts
kotewar: I tried that, but unfortunately the values stored in the `state` are not accessible in the same step as we discovered last time, hence can't validate it. :/

__tests__/stateProvider.test.ts
kotewar: I added more mock asserts to make it clearer as the values can't be accessed in the same step of the action.

hazem454 (approved):

hazem454 (approved):

__tests__/restore.test.ts
kotewar: Added test validation for this as well.

__tests__/stateProvider.test.ts
bishal-pdMSFT: Ah yes, probably we should mock `core.setState` and `core.getState` to achieve this. Currently you are still loading core and using that. May be instead just save in a memory hash and then serve it back with getState. Should be fine to do so as we are not testing `core` functionality here.

But for now it is okay. You can take this if you want in separate PR.

bishal-pdMSFT (approved):

__tests__/stateProvider.test.ts
kotewar: Yes, will take this in a separate PR.

kotewar: Raised a PR for this - https://github.com/actions/cache/pull/1029. 
You were looking for something like this right? 🤔

PR #1017, opened 2022-12-06
Readme updates for granular control

kotewar:
## Description
Readme changes for the granular cache control initiative

## Motivation and Context

These docs will give the users more clarity about the two new actions

## How Has This Been Tested?

NA, tested in workflows

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows).
 
+[Restore action](./restore/README.md)
+
+[Save action](./save/README.md)
+
 ## What's New
 ### v3
 * Added support for caching from GHES 3.5.
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ This action allows caching dependencies and build outputs to improve workflow ex
 In addition to `actions/cache` action, other two actions are also available
 
 [Restore action](./restore/README.md)
+
 [Save action](./save/README.md)
 
 [![Tests](https://github.com/actions/cache/actions/workflows/workflow.yml/badge.svg)](https://github.com/actions/cache/actions/workflows/workflow.yml)
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This action allows caching dependencies and build outputs to improve workflow execution time.
 
-In addition to `actions/cache` action, other two actions are also available
+In addition to this `cache` action, other two actions are also available
 
 [Restore action](./restore/README.md)
 
--- a/README.md
+++ b/README.md
@@ -2,15 +2,17 @@
 
 This action allows caching dependencies and build outputs to improve workflow execution time.
 
+In addition to `actions/cache` action, other two actions are also available
+
+[Restore action](./restore/README.md)
+[Save action](./save/README.md)
+
 [![Tests](https://github.com/actions/cache/actions/workflows/workflow.yml/badge.svg)](https://github.com/actions/cache/actions/workflows/workflow.yml)
 
 ## Documentation
 
 See ["Caching dependencies to speed up workflows"](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows).
 
-[Restore action](./restore/README.md)
-
-[Save action](./save/README.md)
 
 ## What's New
 ### v3
--- a/restore/README.md
+++ b/restore/README.md
@@ -16,8 +16,8 @@ It is very important to use the same `key` and `path` that were used by either `
 ## Outputs
 
 * `cache-hit` - A boolean value to indicate an exact match was found for the key. 
-* `cache-primary-key` - Cache primary key passed in the input to use in subsequent steps of the workflow
-* `cache-restore-key` - Cache key restored
+* `cache-primary-key` - Cache primary key passed in the input to use in subsequent steps of the workflow.
+* `cache-matched-key` -  Key of the cache that was restored, it could either be the primary key on cache-hit or a partial/complete match of one of the restore keys.
 
 > **Note**
 `cache-hit` will be set to `true` only when cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`.
@@ -124,3 +124,7 @@ Since this action comes with its own set of pros and cons, we are listing some o
 ### Reusing primary key and restored key in the save action
 
 One of the limitation you might experience is passing the same input in both `actions/cache/restore` and `actions/cache/save` action. To avoid this, you can make use of the `outputs` from the restore action to reuse the same primary key and also the key of the cache that was restored. This way changing any key in the restore action will automatically reflect in the subsequent actions where the same input is being used.
+
+### Using restore action outputs to make save action behave just like the cache action
+
+The outputs `cache-primary-key` and `cache-matched-key` can be used to check if the restored cache is same as the given primary key. Alternatively, the `cache-hit` output can also be used to check if the restored was a complete match or a partially restored cache.
--- a/restore/README.md
+++ b/restore/README.md
@@ -116,3 +116,11 @@ steps:
   - name: Build
     run: /build.sh
 ```
+
+## Tips
+
+Since this action comes with its own set of pros and cons, we are listing some of the ways by which you can tackle the limitations.
+
+### Reusing primary key and restored key in the save action
+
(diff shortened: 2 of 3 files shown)


restore/README.md
```diff
@@ -0,0 +1,117 @@
...
+> **Note**
+It is very important to use the same `key` and `path` that were used by either `actions/cache` or `actions/cache/save` while saving the cache.
+
+### Environment Variables
+* `SEGMENT_DOWNLOAD_TIMEOUT_MINS` - Segment download timeout (in minutes, default `60`) to abort download of the segment if not completed in the defined number of minutes. [Read more](https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout)
+
+## Outputs
+
+* `cache-hit` - A boolean value to indicate an exact match was found for the key. 
+* `key` -
```
kotewar, line 19: Will add appropriate info here

restore/README.md
```diff
@@ -0,0 +1,126 @@
...
+It is very important to use the same `key` and `path` that were used by either `actions/cache` or `actions/cache/save` while saving the cache.
+
+### Environment Variables
+* `SEGMENT_DOWNLOAD_TIMEOUT_MINS` - Segment download timeout (in minutes, default `60`) to abort download of the segment if not completed in the defined number of minutes. [Read more](https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout)
+
+## Outputs
+
+* `cache-hit` - A boolean value to indicate an exact match was found for the key. 
+* `cache-primary-key` - Cache primary key passed in the input to use in subsequent steps of the workflow
+* `cache-restore-key` - Cache key restored
```
kotewar, line 20: Will update this to a new name

README.md
```diff
@@ -8,6 +8,10 @@ This action allows caching dependencies and build outputs to improve workflow ex
 
 See ["Caching dependencies to speed up workflows"](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows).
 
+[Restore action](./restore/README.md)
+
+[Save action](./save/README.md)
+
```
bishal-pdMSFT, line 14: This can be moved to previous section, where you can call out that in addition to cache actions, other two actions are also available.

restore/README.md
```diff
@@ -0,0 +1,130 @@
+# Restore action
+
+The restore action restores cache just like the main `cache` action, except it doesn't save the cache. So for cases where users only want to restore cache and not save anything in return, this action is pretty helpful. Moreover, this action accepts the same set of inputs as the `cache` action.
```
bishal-pdMSFT, line 3:
```suggestion
The restore action, as the name suggest, restores a cache. It acts similar to the`cache` action except that it doesn't have a post step to save the cache. This action can provide you a granular control to only restore a cache without having to necessarily save it.  It accepts the same set of inputs as the `cache` action.
```

restore/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Inputs
+
+* `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
+* `key` - String used while saving cache for restoring the cache
+* `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key.
+> **Note**
```
bishal-pdMSFT, line 10: Do we need this? It is good for user to use same keys but they can use any key they want.
This statement may scare them.

restore/README.md
```diff
@@ -0,0 +1,130 @@
...
+It is very important to use the same `key` and `path` that were used by either `actions/cache` or `actions/cache/save` while saving the cache.
+
+### Environment Variables
+* `SEGMENT_DOWNLOAD_TIMEOUT_MINS` - Segment download timeout (in minutes, default `60`) to abort download of the segment if not completed in the defined number of minutes. [Read more](https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout)
+
+## Outputs
```
bishal-pdMSFT, line 16: May be outputs section should come before environment variables one.

restore/README.md
```diff
@@ -0,0 +1,130 @@
...
+    if: steps.cache.outputs.cache-hit != 'true'
+    run: exit 1
+
+  - name: Build
+    run: /build.sh
+```
+
+## Tips
+
+Since this action comes with its own set of pros and cons, we are listing some of the ways by which you can tackle the limitations.
```
bishal-pdMSFT, line 122: Don't call them limitations. This whole statement is not useful
```suggestion
```

restore/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Reusing primary key and restored key in the save action
+
+One of the limitation you might experience is passing the same input in both `actions/cache/restore` and `actions/cache/save` action. To avoid this, you can make use of the `outputs` from the restore action to reuse the same primary key and also the key of the cache that was restored. This way changing any key in the restore action will automatically reflect in the subsequent actions where the same input is being used.
```
bishal-pdMSFT, line 126:
```suggestion
Usually you may want to use same `key` in both actions/cache/restore` and `actions/cache/save` action. To achieve this, use `outputs` from the restore action to reuse the same primary key (or the key of the cache that was restored).
```

save/README.md
```diff
@@ -0,0 +1,87 @@
+# Save action
+
+The save action saves cache just like the post step of the `cache` action, except it can be run individually now without having to call the restore step. The save action needs to be called with required inputs, the `key` with which the cache needs to be stored and the `path` that needs to be cached.
```
bishal-pdMSFT, line 3:
```suggestion
The save action, as the name suggest, saves a cache. It acts similar to the `cache` action except that it doesn't necessarily first do a restore. This action can provide you a granular control to only save a cache without having to necessarily restore it, or to do a restore anywhere in the workflow job and not only in post phase.
```

save/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `path` - 'A list of files, directories, and wildcard patterns to cache'
+* `upload-chunk-size` - 'The chunk size used to split up large files during upload, in bytes'
+
+## Outputs
+
+This action has no outputs.
+
+## Use cases
+
+As this is a newly introduced action to give users more control in their workflows, below are some use cases where one can use this action.
```
bishal-pdMSFT, line 17: Not needed
```suggestion
```

save/README.md
```diff
@@ -0,0 +1,87 @@
...
+  - uses: actions/cache/save@v3
+    id: cache
+    with:
+      path: path/to/dependencies
+      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
+
+### Re-evaluate cache key while saving
+
+Some technologies like dot-net generate the lockfiles during the build time, due to which the already evaluated `${{ hashFiles('**/lockfiles') }}` hash doesn't match the actual hash. Using save action with the same key will not re-evaluate the key as the hash was generated during restores. However passing the expression to re-evaluate the hash would generate a new key in the save action now as save would be called after the build step.
```
bishal-pdMSFT, line 43: Too verbose. Just mention that this action allows re-evaluation of key.

save/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+In other words, you can reuse the output of `restore` action in the input of the `save` action to input the same key. Otherwise, you can pass the hash logic for the key (ex. `${{ hashFiles('**/lockfiles') }}`) and let the save action re-evaluate the key just before executing.
```
bishal-pdMSFT, line 45: This can be removed as well

restore/README.md
kotewar: We are following the same sequence in the cache action readme as well.

restore/README.md
kotewar: Can be removed, existing users would be knowing this already. But I just added this for the new users to know that the key+path combination has to be present in both actions to restore/save the same cache. 

We don't want users to file bugs for these trivial cases due to lack of info. 😅

save/README.md
kotewar: Done

restore/README.md
bishal-pdMSFT: I was thinking from user perspective. Output is more common need than env. In a doc we should start with more commonly useful info at top and then go down to less common useful ones.

bishal-pdMSFT: I was thinking from user perspective. Output is more common need than env. In a doc we should start with more commonly useful info at top and then go down to less common useful ones.

bishal-pdMSFT (approved):

restore/README.md
kotewar: Updated

PR #1018, opened 2022-12-06
Update save with a wrapper layer

bishal-pdMSFT:
## Description

- Added wrapper layers for save
- Renamed StateOutputter to StateManager and removed getOutput. Output will be managed directly using core lib. This class will only handle `state`

## Types of changes

## Checklist:


src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+    //setOutput = core.setOutput;
+    setState = core.saveState;
+    getState = core.getState;
+}
+
+export class NullStateProvider extends StateProviderBase {
+    //setOutput = core.setOutput;
+    setState = core.setOutput;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getState = (key: string) => "";
```
kotewar, line 41: Should `getState` be set to `undefined`?

src/saveImpl.ts
```diff
@@ -0,0 +1,65 @@
...
+            core.getInput(Inputs.Key);
+
+        if (!primaryKey) {
+            utils.logWarning(`Error retrieving key from state.`);
+            return;
+        }
+
+        // If matched restore key is same as primary key, then do not save cache
+        // NO-OP in case of SaveOnly action
+        const state = stateProvider.getCacheState();
```
kotewar, line 41: If my restore was successful in my restore-only action, I would still get `undefined` here right?

kotewar:
```
const state = stateProvider.getCacheState() || core.getInput(Inputs.MatchedKey);
```

`MatchKey` will be output from `restore-only` action

kotewar: Just noting things down...

In Save-only action, user can give the following inputs-
1. cache key - from output of `restore` action or same cache `key` as input to `restore` to allow re-evaluation. (can 
be user input or fetched from output of restore-action depending on use case.)
2. cache matched key - this key is the one that was actually restored, could be exact key or restored from restore keys to determine the `cache-hit`. (This has to be taken as input from the output of the previous restore action)
3. path - same as input to restore. (Always has to be inputted by the user)

src/stateProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+class StateProviderBase implements IStateProvider {
+    getCacheState(): string | undefined {
+        const cacheKey = this.getState(State.CacheMatchedKey);
+        if (cacheKey) {
+            core.debug(`Cache state/key: ${cacheKey}`);
+            return cacheKey;
+        }
+
+        return undefined;
+    }
```
kotewar, line 22: Coming from this [comment](https://github.com/actions/cache/pull/1018/files#r1041929081), 

it seems like this is just a `getState()` implementation with a parameter. We can remove this and depend on `getState` method of `StateProvider` and `NullStateProvider`

src/saveImpl.ts
bishal-pdMSFT: > cache key - from output of restore action or same cache key as input to restore to allow re-evaluation. (can be user input or fetched from output of restore-action depending on use case

@kotewar can we directly read output of earlier step1 with code in step2? Or user still needs to set the input for step2 as output of step1?

kotewar: @bishal-pdMSFT We can if we store it as an environment variable.

PR #1024, opened 2022-12-09
Granular cache control save related changes and new inputs

kotewar:
## Types of changes

## Checklist:

### DIFF
--- a/__tests__/restoreOnly.test.ts
+++ b/__tests__/restoreOnly.test.ts
@@ -1,0 +1,322 @@
+import * as cache from "@actions/cache";
+import * as core from "@actions/core";
+
+import { Events, Inputs, RefKey } from "../src/constants";
+import run from "../src/restoreOnly";
+import * as actionUtils from "../src/utils/actionUtils";
+import * as testUtils from "../src/utils/testUtils";
+
+jest.mock("../src/utils/actionUtils");
+
+beforeAll(() => {
+    jest.spyOn(actionUtils, "isExactKeyMatch").mockImplementation(
+        (key, cacheResult) => {
+            const actualUtils = jest.requireActual("../src/utils/actionUtils");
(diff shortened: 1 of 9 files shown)


__tests__/restoreOnly.test.ts
```diff
@@ -0,0 +1,322 @@
+import * as cache from "@actions/cache";
```
bishal-pdMSFT, line 1: Are these tests same as restore.tests.ts. If yes then they are duplicate. We should move these tests to restoreImpl.tests.ts and keep only very minimal tests in restore.tests.ts and restoreOnly.tests.ts. For example these tests can verify that output is set in case of restoreOnly instead of state and vice versa.

src/constants.ts
```diff
@@ -6,7 +6,9 @@ export enum Inputs {
 }
 
 export enum Outputs {
-    CacheHit = "cache-hit"
+    CacheHit = "cache-hit",
+    Key = "key",
+    MatchedKey = "matched-key"
```
bishal-pdMSFT, line 11: Why are these two needed?

save/action.yml
```diff
@@ -11,9 +11,12 @@ inputs:
   upload-chunk-size:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
+  matched-key:
```
bishal-pdMSFT, line 14: This is not correct. isn't it? Both primary key and matched key will use same value as State constants. No?
https://github.com/actions/cache/blob/be72d0432d4f55dc4892a372e19e129b600b26b5/src/constants.ts#L12

src/constants.ts
kotewar: So that they can be outputted from restore-only action and reused in the save action. 🤔

save/action.yml
kotewar: Good point, we will use different names so that the inputs/states/outputs don't coincide.

PR #1034, opened 2022-12-20
Warning added for save only action

kotewar:
## Description

For save only action, it is important for users to know that the save has failed due to any reason, hence we've added a warning to show the same.

## Motivation and Context

Make user experience better by highlighting any warnings/issues in the workflow run

## How Has This Been Tested?

Local workflow 

Save action - https://github.com/kotewar/test-continue-on-error-for-cache/actions/runs/3742275202
Default cache action - https://github.com/kotewar/test-continue-on-error-for-cache/actions/runs/3742531272/jobs/6353592058#step:13:665

NA 

## Types of changes

## Checklist:

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "npm install && npm run build"
+	"postCreateCommand": "npm install"
 	// Configure tool-specific properties.
 	// "customizations": {},
 	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
--- a/dist/save/index.js
+++ b/dist/save/index.js
@@ -41063,14 +41063,14 @@ const utils = __importStar(__webpack_require__(443));
 process.on("uncaughtException", e => utils.logWarning(e.message));
 function saveImpl(stateProvider) {
     return __awaiter(this, void 0, void 0, function* () {
-        let cacheId;
+        let cacheId = -1;
         try {
             if (!utils.isCacheFeatureAvailable()) {
-                return -2; //-2 refers as safe to ignore for the caller
+                return;
             }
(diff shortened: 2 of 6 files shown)


.devcontainer/devcontainer.json
```diff
@@ -6,7 +6,7 @@
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "npm install && npm run build"
+	"postCreateCommand": "npm install"
```
kotewar, line 9: npm run builds take a very long time while setting up the codespace, hence removing it from here.

dist/save/index.js
```diff
@@ -41063,33 +41063,34 @@ const utils = __importStar(__webpack_require__(443));
 process.on("uncaughtException", e => utils.logWarning(e.message));
 function saveImpl(stateProvider) {
     return __awaiter(this, void 0, void 0, function* () {
+        let cacheId;
         try {
             if (!utils.isCacheFeatureAvailable()) {
-                return;
+                return 0;
```
kotewar, line 41069: We will return 0 when we don't want to log a warning, and -1 when we want to log a warning.

bishal-pdMSFT, line 41069: 0 may be a real cacheId as DB entry starts at 0. Should we use something like -2

kotewar: Yes we can. I was thinking, in the future we can also map the negative numbers to some error types to decide how to log/show error to the user.

kotewar: Used -2 now.

bishal-pdMSFT: Lets do that now itself. We can define an enum on cache return type.

bishal-pdMSFT: Actually re-thinking about this. Why do we need to return id here. Can we return error here. Any error and SaveOnly should print warning message. No?

kotewar: Does this mean we also need to implement error handling in save.ts?

bishal-pdMSFT (approved):

PR #1036, opened 2022-12-22
Add `fail-on-cache-miss` option

### LINKED ISSUE #955: optional action failure on cache miss

cdce8p:
## Description
Add an advanced option to fail job if cache miss occurs.
Based on the work from @kotewar which was reverted in https://github.com/actions/cache/pull/1006/commits/11ab7ccfa2803e57082f50b6bcf6d627d72f6b26.

## Motivation and Context

Fixes #955 and partially https://github.com/actions/cache/discussions/1020#discussioncomment-4341251

This is already possible via the `cache-hit` output. However, especially for advanced workflows, this quickly becomes cumbersome to add every time. It's also an additional 3 line for each job and doesn't include a good error message.
```yaml
  - name: Check cache hit
    run: exit 1
```

As mentioned in https://github.com/actions/cache/issues/955#issuecomment-1362681576, the goal to simplify the configuration doesn't exclude the possibility to help with more advanced use cases. Including the new option will make it easier to use, help keep workflows more readable, and enable better error messages.

## How Has This Been Tested?

Added test cases and run a test workflow using the feature branch.

 --

## Types of changes

## Checklist:

### DIFF
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -290,3 +290,43 @@ test("restore when fail on cache miss is enabled and primary key doesn't match r
     );
     expect(failedMock).toHaveBeenCalledTimes(0);
 });
+
+test("restore with fail on cache miss disabled and no cache found", async () => {
+    const path = "node_modules";
+    const key = "node-test";
+    const restoreKey = "node-";
+    testUtils.setInputs({
+        path: path,
+        key,
+        restoreKeys: [restoreKey],
+        failOnCacheMiss: false
+    });
+
+    const infoMock = jest.spyOn(core, "info");
+    const failedMock = jest.spyOn(core, "setFailed");
+    const stateMock = jest.spyOn(core, "saveState");
+    const restoreCacheMock = jest
+        .spyOn(cache, "restoreCache")
+        .mockImplementationOnce(() => {
+            return Promise.resolve(undefined);
+        });
+
+    await run();
+
+    expect(restoreCacheMock).toHaveBeenCalledTimes(1);
+    expect(restoreCacheMock).toHaveBeenCalledWith(
+        [path],
+        key,
+        [restoreKey],
+        {},
+        false
+    );
+
+    expect(stateMock).toHaveBeenCalledWith("CACHE_KEY", key);
+    expect(stateMock).toHaveBeenCalledTimes(1);
+
+    expect(infoMock).toHaveBeenCalledWith(
+        `Cache not found for input keys: ${key}, ${restoreKey}`
+    );
+    expect(failedMock).toHaveBeenCalledTimes(0);
+});
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -206,7 +206,7 @@ test("restore with cache found for restore key", async () => {
     expect(failedMock).toHaveBeenCalledTimes(0);
 });
 
-test("Fail restore when fail on cache miss is enabled and primary key not found", async () => {
+test("Fail restore when fail on cache miss is enabled and primary + restore keys not found", async () => {
     const path = "node_modules";
     const key = "node-test";
     const restoreKey = "node-";
@@ -246,7 +246,7 @@ test("Fail restore when fail on cache miss is enabled and primary key not found"
     expect(failedMock).toHaveBeenCalledTimes(1);
 });
 
-test("Fail restore when fail on cache miss is enabled and primary key doesn't match restored key", async () => {
+test("restore when fail on cache miss is enabled and primary key doesn't match restored key", async () => {
     const path = "node_modules";
     const key = "node-test";
     const restoreKey = "node-";
@@ -257,6 +257,7 @@ test("Fail restore when fail on cache miss is enabled and primary key doesn't ma
         failOnCacheMiss: true
     });
 
+    const infoMock = jest.spyOn(core, "info");
     const failedMock = jest.spyOn(core, "setFailed");
     const stateMock = jest.spyOn(core, "saveState");
     const setCacheHitOutputMock = jest.spyOn(core, "setOutput");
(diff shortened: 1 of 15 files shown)


cdce8p:
@kotewar I rebased all my PRs to resolve the merge conflicts and update the tests after the recent changes. Is there a timeline when this can move forward? It would be awesome if I wouldn't have to resolve merge conflicts every other week.

restore/action.yml
```diff
@@ -15,6 +15,10 @@ inputs:
     description: 'An optional boolean when enabled, allows windows runners to restore caches that were saved on other platforms'
     default: 'false'
     required: false
+  fail-on-cache-miss:
+    description: 'Fail the workflow if the cache is not found for the primary key'
```
bishal-pdMSFT, line 19: This would cause failure even if `restore-key` matched. I can imagine another case for failure should to happen when that also did not match. That would be true cache "miss".

cdce8p: I agree. That's an implicit result if no cache is found for the primary key. I've updated the description to make it a bit more clear.

bishal-pdMSFT: Apologies I meant a different thing. What happen if a restore-key matched? The job should still proceed as there is a partial match. I can agree to fail if there is no match, but failing on partial match seems too strict to me. This looks like very small use case.
And we don't want to end up with many inputs like `fail-on-no-or-partial-match` and `fail-on-no-match` etc.

action.yml
```diff
@@ -18,6 +18,10 @@ inputs:
     description: 'An optional boolean when enabled, allows windows runners to save or restore caches that can be restored or saved respectively on other platforms'
     default: 'false'
     required: false
+  fail-on-cache-miss:
+    description: 'Fail the workflow if no cache entry is not found'
```
bishal-pdMSFT, line 22:
```suggestion
    description: 'Fail the workflow if cache entry is not found'
```

restore/README.md
```diff
@@ -95,7 +96,7 @@ steps:
 
 ### Exit workflow on cache miss
 
-You can use the output of this action to exit the workflow on cache miss. This way you can restrict your workflow to only initiate the build when `cache-hit` occurs, in other words, cache with exact key is found.
+You can use `fail-on-cache-miss: true` to exit the workflow on a cache miss. This way you can restrict your workflow to only initiate the build when a cache with the exact key is found. Make sure to leave `restore-keys` empty!
```
bishal-pdMSFT, line 99:
```suggestion
You can use `fail-on-cache-miss: true` to exit the workflow on a cache miss. This way you can restrict your workflow to only initiate the build when a cache is matched. Also, if you want to fail if cache did not match primary key, additionally leave `restore-keys` empty!
```

kotewar (changes requested):

__tests__/restore.test.ts
```diff
@@ -205,3 +205,88 @@ test("restore with cache found for restore key", async () => {
...
+
+test("Fail restore when fail on cache miss is enabled and primary + restore keys not found", async () => {
+    const path = "node_modules";
+    const key = "node-test";
+    const restoreKey = "node-";
+    testUtils.setInputs({
+        path: path,
+        key,
+        restoreKeys: [restoreKey],
+        failOnCacheMiss: true
```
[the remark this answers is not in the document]
cdce8p: 👍🏻 Done. Additionally, since I used `utils.getInputAsBool(Inputs.FailOnCacheMiss)` to get the input, basically all tests where `failOnCacheMiss` isn't explicitly set to `true` will test any remaining cases.

https://github.com/actions/cache/blob/8e3048d0f72c72babe53747a8f10e16a74c19ba1/src/utils/actionUtils.ts#L55-L60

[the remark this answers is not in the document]
kotewar: Great thanks :)

kotewar:
Looks good to me, before we can release this out, we need a few additions
1. Please update the patch version in [package.json](https://github.com/cdce8p/cache/blob/fail-on-cache-miss/package.json)
2. Please update [RELEASES.md](https://github.com/cdce8p/cache/blob/fail-on-cache-miss/RELEASES.md) and the `Whats new` section of [README.md](https://github.com/cdce8p/cache/blob/fail-on-cache-miss/README.md) with appropriate release info.

I'll approve, merge the PR and release the new version once the info is updated! Thanks :)

kotewar:
> Whats new section of [README.md](https://github.com/cdce8p/cache/blob/fail-on-cache-miss/README.md) with appropriate release info.

This also please. Just one liner is enough.

cdce8p:
> > Whats new section of [README.md](https://github.com/cdce8p/cache/blob/fail-on-cache-miss/README.md) with appropriate release info.
> 
> This also please. Just one liner is enough.

Sorry. Missed that the first time. Should be good now.

kotewar (approved):
LGTM

kotewar:
Thanks @cdce8p for your contribution, the changes are live 🚀 

Please do test the positive and negative responses once with the latest tag `actions/cache@v3` and `actions/cache/restore@v3`.

cdce8p:
> Thanks @cdce8p for your contribution, the changes are [live](https://github.com/actions/cache/tree/v3.2.4) 🚀

Awesome 🎉 Thanks for your help getting it ready.

> Please do test the positive and negative responses once with the latest tag `actions/cache@v3` and `actions/cache/restore@v3`.

`actions/cache@v3.2.4` does seem to work as expected from my initial testing. However, the [`v3`](https://github.com/actions/cache/tree/v3) tag doesn't seem to have been updated. It still points to `3.2.3` (58c146cc91c5b9e778e71775dfe9bf1442ad9a12) not `3.2.4` (58c146cc91c5b9e778e71775dfe9bf1442ad9a12).
Thus I just get `Unexpected input(s) 'fail-on-cache-miss'` for that.

kotewar:
Let me check

kotewar:
My bad, I forgot to tag the version to `v3`. Please check now.

cdce8p:
> My bad, I forgot to tag the version to `v3`. Please check now.

It works now! Thanks

PR #1055, opened 2023-01-04
Cache recipes for cache, restore and save actions

kotewar:
## Description
Setting up context for the cache recipes

## Motivation and Context
As we plan to add recipes for all the caches, its important to set up context in the beginning so that the users understand the workflows better.

## How Has This Been Tested?
NA - doc change.

## Types of changes

## Checklist:

### DIFF
--- a/CACHING.md
+++ b/CACHING.md
@@ -21,7 +21,7 @@ In your workflows, you can use different strategies to name your key depending o
 
 ### Updating cache for any change in the dependencies
 
-One of the most common use case is to use hash for lockfile as key. This way same cache will be restored for same lockfiles until there's any change in the lockfile/dependencies.
+One of the most common use case is to use hash for lockfile as key. This way, same cache will be restored for a lockfile until there's a change in dependencies listed in lockfile.
 
 ```yaml
   - uses: actions/cache@v3
--- a/CACHING.md
+++ b/CACHING.md
@@ -17,7 +17,7 @@ jobs:
 
 In your workflows, you can use different strategies to name your key depending on your use case so that the cache is scoped properly based on the need. If you wish to create OS specific caches, or caches based on the lockfiles, commit SHA, workflow run id, etc. then you can generate the keys dynamically at run-time. Below are some of the tips to strategically name your cache using the [cache](https://github.com/actions/cache) or [restore](https://github.com/actions/cache/tree/main/restore) action.
 
-**Updating cache for any change in the dependencies**
+### Updating cache for any change in the dependencies
 One of the most common use case is to use hash for lockfile as key. This way same cache will be restored for same lockfiles until there's any change in the lockfile/dependencies.
 ```yaml
   - uses: actions/cache@v3
@@ -28,7 +28,7 @@ One of the most common use case is to use hash for lockfile as key. This way sam
       key: cache-${{ hashFiles('**/lockfiles') }}
 ```
 
-**Using restore keys to download the closest matching cache**
+### Using restore keys to download the closest matching cache
 If cache is not found matching the primary key, restore keys will be used to download the closest matching cache that was recently created. This way most of the dependencies can be downloaded from the restore key hence saving some build time.
 
 ```yaml
@@ -44,8 +44,8 @@ If cache is not found matching the primary key, restore keys will be used to dow
 
 The restore keys can be provided as a complete name, or a prefix, read more [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) on how a cache key is matched using restore keys.
 
-**Separating the caches by Operating System**
-Caches can be stored separately for different Operating Systems. This can be used in combination with hashfiles in case multiple caches are being generated per OS. 
+### Separate caches by Operating System
+In case of workflows with matrix running for multiple Operating Systems, the caches can be stored separately for each of them. This can be used in combination with hashfiles in case multiple caches are being generated per OS. 
 ```yaml
   - uses: actions/cache@v3
     with:
@@ -55,13 +55,13 @@ Caches can be stored separately for different Operating Systems. This can be use
       key: ${{ runner.os }}-cache
 ```
 
-**Limiting cache to only the current workflow/attempt**
+### Limiting cache to only the current workflow/attempt
 Caches scoped to the particular workflow run id or run attempt can be stored and referred by using the run id/attempt
 ```yaml
     key: cache-${{ github.run_id }}-${{ github.run_attempt }}
 ```
 
-**Limiting cache for a particular commit** 
+### Limiting cache for a particular commit
 For very short term or isolated use cases, where cache is supposed to be short lived, commit sha can be used.
 ```yaml
   - uses: actions/cache@v3
@@ -72,7 +72,7 @@ For very short term or isolated use cases, where cache is supposed to be short l
       key: cache-${{ github.sha }}
 ```
 
-**Using multiple factors while forming a key depening on the need**
+### Using multiple factors while forming a key depening on the need
 Cache key can be formed by combination of more than one metadata, evaluated info.
 ```yaml
   - uses: actions/cache@v3
@@ -207,7 +207,7 @@ steps:
 
 ## Restoring and saving caches
 
-## Restoring and saving cache using a single action
+### Restoring and saving cache using a single action
 
 The [cache](https://github.com/actions/cache/tree/main#cache) action allows caching dependencies and restoring them using a single action. It has a `main` step and a `post` step. In the `main` step, the cache is restored if it exists for the input `key`, `path` combination (refer [scope](#scope)). If cache is not found for the given `key` input, then cache is restored using [restore keys](#restore-keys) . If the cache doesn't exist or is restored using `restore-keys`, the cache is saved in the `post` step of this action.
 
@@ -223,6 +223,22 @@ The [cache](https://github.com/actions/cache/tree/main#cache) action allows cach
 ```
 
 The `cache` action provides one output `cache-hit` which is set to `true` when cache is restored using primary key and `false` when cache is restored using `restore-keys` or no cache is restored.
+
+### Download remaining dependencies in case of cache miss
+
+In case cache gets download using restore keys, there's a chance that some dependencies might be missing. It might also be possible that no cache was restored because there was no match. In such cases, the output `cache-hit` is set to `false`. We can make use of this output to download the remaining dependencies.
+
+```yaml
+  - uses: actions/cache@v3
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+  - name: Install Dependencies
+    if: steps.cache.outputs.cache-hit != 'true'
+    run: ./install-dependencies.sh
+```
 
 ### Using combination of restore and save actions
 
@@ -253,7 +269,7 @@ steps:
   - uses: actions/checkout@v3
 
   - name: Build
-    run: /build-parent-module.sh
+    run: ./build-parent-module.sh
 
   - uses: actions/cache/save@v3
     id: cache
@@ -275,11 +291,11 @@ steps:
 
   - name: Install Dependencies
     if: steps.cache.outputs.cache-hit != 'true'
-    run: /install.sh
+    run: ./install.sh
       
   - name: Build
-    run: /build-child-module.sh
+    run: ./build-child-module.sh
 
   - name: Publish package to public
-    run: /publish.sh
+    run: ./publish.sh
 ```
--- a/CACHING.md
+++ b/CACHING.md
@@ -2,8 +2,9 @@
 
 This document lists some of the strategies (and example workflows if possible) which can be used
 
-- to solve some common use cases
-- to effectively leverage the step inputs and outputs
+- to use an effective cache key and/or path
+- to solve some common use cases around saving and restoring caches
+- to leverage the step inputs and outputs more effectively
 
 ## Strategically using keys
 
--- a/CACHING.md
+++ b/CACHING.md
@@ -10,6 +10,7 @@ The cache action works using the following set of inputs and outputs as mentione
 
 ### Keys
 
+A `key`, also referred as `primary key` is a value with which cache is restored or saved. If cache is not found with the primary key, the same key is used to save the cache in the `actions/cache` action.
 #### Static keys
 
 ```yaml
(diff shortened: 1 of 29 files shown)


CACHING.md
```diff
@@ -0,0 +1,143 @@
+# Cache Recipes
+
+This document focuses on gathering all the common use cases that will help the users of `actions/cache`
+- Know the use cases that can be tackled and sample workflows on how to tackle them
+- Optimise their workflows to use the cache inputs and outputs better
```
bishal-pdMSFT, line 5: I don't think we can list **all** use cases. 
```suggestion
This document lists some of the strategies (and example workflows if possible) which can be used
- to solve some common use cases
- to effectively leverage the step inputs and outputs
```

CACHING.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Actions Cache Basics
+
+The cache action works using the following set of inputs and outputs as mentioned [here](https://github.com/actions/cache#inputs). However these inputs are self explaining below are some ways in which the inputs can be used in a better way.
+
+### Keys
+
+#### Static keys
```
bishal-pdMSFT, line 13: rather than minor cases like static keys, focus on the most important use cases. For example,

- One of the most common use case is to use hash for lockfile as a key. This way same set of cached dependencies can be easily restored.
- Another use case could be to create different caches for different OS. In that case runner os can be used as a key
- For very short term (or isolated) use cases where cache is supposed to be short lived commit sha can be used. Similarly, for a cache which is scoped to a specific workflow run, the run id can be used.

CACHING.md
```diff
@@ -0,0 +1,143 @@
...
+Version = hash(Compression tool, Path(s) to be cached)
+
+### Branch
+
+Whenever a cache is saved, the repository `branch` where it was generated is also stored along with it. This is done mainly to avoid caches from `feature` branches to interact with jobs running on the `default` branch. 
+
+### Scope
+
+The cache is scoped to a key, version and branch. The default branch cache is accessible to all other branches, but not the other way round. This means if you have a cache with key matching (completely or partially) and (exact) version in the default branch, you will be able to restore the cache in any of the branches. However if you create a cache in a feature branch, it cannot be restored in any other branch.
+Refer [matching the key](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) for more info on how keys are matched and restored.
```
bishal-pdMSFT, line 81: These should not be needed in this doc. If needed we should update the relevant [doc](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action).

CACHING.md
```diff
@@ -0,0 +1,143 @@
...
+    key: cache-${{ hashFiles('**/lockfiles') }}
+```
+Cache key by combination of multiple options: 
+```yaml
+    key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
+
+The [GitHub Context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) can be used to create keys using the workflows metadata.
+
+### Paths
```
bishal-pdMSFT, line 59: I don't think we should explain what is Path. We should tell that a selected list of paths can be used. For example for node, we can cache npm-cache directory. Similarly few more examples.

CACHING.md
```diff
@@ -0,0 +1,143 @@
...
+-->
+
+## Save followed by restore
+<!--
+- Use case for saving first and restoring across child/subsequent jobs
+- A complex workflow with save and restore actions called multiple times.
+- Save cache on any failure.
+-->
+
+## Storage & CLI
```
bishal-pdMSFT, line 124: This section should not be in this doc

CACHING.md
```diff
@@ -0,0 +1,143 @@
...
+Basic restore save example
+- Using default cache action
+- Using restore and save actions together with possibilities.
+Restore save example with advanced controls
+- Using if conditions to achieve use cases that we solved 
+- Fail workflow on cache miss
+- Force rewrite cache by deleting cache between steps using cli
+-->
+
+## Save followed by restore
```
bishal-pdMSFT, line 146: Focus on the user problem which is being solved. For example, here we can say that - `Save a cache in a job and reuse in multiple jobs`

Aiasp1985P:
Me puedes traducir al español

CACHING.md
kotewar: I plan to use the information given here in the optimisation section below. Is it okay to assume the readers would be aware of cache management UI and CLI?

CACHING.md
kotewar: Sure, I'll update the headings as I update these sections. 👍🏽

CACHING.md
kotewar: The intention of adding these sections was to ensure everyone understands the basics of actions cache as these terminologies will be used frequently in the coming sections of the doc especially for the optimisation use cases.

CACHING.md
kotewar: I've added these from line no. 37 to 58. I can update them with the descriptions that you have provided in the comment.

CACHING.md
bishal-pdMSFT: While I agree that the reader should have good understanding of these terms, I think this doc is not the right place to explain these. Such explanation should go to docs where we have sections for inputs and scope etc. If those sections are missing details, then they should be enhanced.
What we can do here is to just tel user to ensure that they understand various aspects of caching and provide links to doc.

serroopoint (commented):
28016959

CACHING.md
```diff
@@ -0,0 +1,351 @@
+# Cache Recipes
+
+This document lists some of the strategies (and example workflows if possible) which can be used
+
+- to solve some common use cases
+- to effectively leverage the step inputs and outputs
```
bishal-pdMSFT, line 6:
```suggestion
- to use an effective cache key and/or path
- to solve some common use cases around saving and restoring caches
- to leverage the step inputs and outputs more effectively
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Strategically using keys
```
bishal-pdMSFT, line 8: Choosing right key

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+```yaml
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    - uses: actions/cache@v3
+      with:
+        key: ${{ some-metadata }}-cache
+```
+
+In your workflows, you can use different strategies to name your key depending on your use case so that the cache is scoped properly based on the need. If you wish to create OS specific caches, or caches based on the lockfiles, commit SHA, workflow run id, etc. then you can generate the keys dynamically at run-time. Below are some of the tips to strategically name your cache using the [cache](https://github.com/actions/cache) or [restore](https://github.com/actions/cache/tree/main/restore) action.
```
bishal-pdMSFT, line 19: Shortening the sentence to try to keep it simple. Removed reference to actions as that is quite obvious. No?
```suggestion
In your workflows, you can use different strategies to name your key depending on your use case so that the cache is scoped appropriately for your need. For example, you can have cache specific to OS, or based on the lockfile or commit SHA or even workflow run.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: cache-${{ hashFiles('**/lockfiles') }}
+```
+
+### Using restore keys to download the closest matching cache
+
+If cache is not found matching the primary key, restore keys will be used to download the closest matching cache that was recently created. This way most of the dependencies can be downloaded from the restore key hence saving some build time.
```
bishal-pdMSFT, line 36:
```suggestion
If cache is not found matching the primary key, restore keys can be used to download the closest matching cache that was recently created. This ensures that the build/install step will need to additionally fetch just a handful of newer dependencies, and hence saving build time.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ runner.os }}-cache
+```
+
+### Limiting cache to only the current workflow/attempt
+
+Caches scoped to the particular workflow run id or run attempt can be stored and referred by using the run id/attempt
```
bishal-pdMSFT, line 66:
```suggestion
Caches scoped to the particular workflow run id or run attempt can be stored and referred by using the run id/attempt. This is an effective way to keep a short lived cache.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+```yaml
+  - uses: actions/cache@v3
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ runner.os }}-cache
+```
+
+### Limiting cache to only the current workflow/attempt
```
bishal-pdMSFT, line 64:
```suggestion
### Creating a short lived cache
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+Caches scoped to the particular workflow run id or run attempt can be stored and referred by using the run id/attempt
+
+```yaml
+    key: cache-${{ github.run_id }}-${{ github.run_attempt }}
+```
+
+### Limiting cache for a particular commit
+
+For very short term or isolated use cases, where cache is supposed to be short lived, commit sha can be used.
```
bishal-pdMSFT, line 74:
```suggestion
On similar lines, commit sha can be used to create a very specialized and short lived cache.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
+
+The [GitHub Context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) can be used to create keys using the workflows metadata.
+
+## Restoring Cache
+
+### Know paths better
```
bishal-pdMSFT, line 102:
```suggestion
### Understand path
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+While setting paths for caching dependencies it is important to give correct path depending on the image you are using and if you are running the action inside or outside the container. Below are some paths one should keep in mind while writing their workflow files.
```
bishal-pdMSFT, line 104:
```suggestion
While setting paths for caching dependencies it is important to give correct path depending on the hosted runner you are using or whether the action is running in a container job. Assigning different `path` for save and restore will result in cache miss. 

Below are GiHub hosted runner specific paths one should take care of when writing a workflow which saves/restores caches across OS.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+#### Ubuntu Paths
+
+```yaml
+process.env['RUNNER_TEMP']=/home/runner/work/_temp 
+process.cwd()=/home/runner/work/repo/repo
+Home directory = /home/runner
+${{ github.workspace }} = /home/runner/work/repo/repo
```
bishal-pdMSFT, line 112: Changed the order a bit based on priority of usage
```suggestion
Home directory (`~/`) = `/home/runner`
`${{ github.workspace }}` = `/home/runner/work/repo/repo`
`process.env['RUNNER_TEMP']`=`/home/runner/work/_temp` 
`process.cwd()` = `/home/runner/work/repo/repo`
```

bishal-pdMSFT: If okay, then change similar for other OSes as well

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+  - name: Build
+    run: /build.sh
+```
+
+## Saving cache
+
+### Reusing primary key from restore cache as input to save action
+
+If you want to avoid re-writing the cache key again in `save` action, the outputs from `restore` action can be used as input to the `restore` action.
```
bishal-pdMSFT, line 189:
```suggestion
If you want to avoid re-computing the cache key again in `save` action, the outputs from `restore` action can be used as input to the `restore` action.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ steps.restore-cache.outputs.key }}
+```
+
+### Re-evaluate cache key while saving cache
+
+With save action, the key can now be re-evaluated while executing the action. This helps in cases where the lockfiles are generated during the build.
```
bishal-pdMSFT, line 212:
```suggestion
On the other hand, the key can also be explicitly re-computed while executing the save action. This helps in cases where the lockfiles are generated during the build.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+
+```yaml
+uses: actions/cache/save@v3
+with:
+    key: npm-cache-${{hashfiles(package-lock.json)}}
+```
+
+### Saving cache even if the build fails
+
+There are instances where some flaky test cases would fail the entire workflow and users would get frustrated because the builds would run for hours and the cache couldn't get saved as the workflow failed in between. For such use-cases, users would now have the ability to use `actions/cache/save` action to save the cache by using `if: always()` condition. This way the cache will always be saved if generated, or a warning will be thrown that nothing is found on the cache path. Users can also use the `if` condition to only execute the `actions/cache/save` action depending on the output of the previous steps. This way they get more control on when to save the cache.
```
bishal-pdMSFT, line 241:
```suggestion
There can be cases where a cache should be saved even if the build job fails. For example, a job can fail due to flaky tests but the caches can still be re-used. You can use `actions/cache/save` action to save the cache by using `if: always()` condition. 

Similarly, `actions/cache/save` action can be conditionally used based on the output of the previous steps. This way you get more control on when to save the cache.
```

CACHING.md
```diff
@@ -0,0 +1,351 @@
...
+      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+  - name: build
+    run: ./install.sh
+  - uses: actions/cache/save@v3
+    with:
+      path: |
+        path/to/dependencies
+        some/other/dependencies
+      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
+```
```
bishal-pdMSFT, line 308: This looks like should already be covered in `README as it talking about basics of how to use cache and other new actions.

bishal-pdMSFT (approved):

CACHING.md
kotewar: The order makes sense, I'll change at all places.

CACHING.md
kotewar: Its not there in README, I'll shift it there.

kotewar: Moved.

PR #1049, opened 2022-12-27
Revert compression changes related to windows but keep version logging

Phantsure:
due to symlink issues

## Description

Revert compression changes related to windows but keep version logging

## Motivation and Context

Some issues have been reported due to `linux/mac` cache not being recovered on windows because of symlink issue. We would be testing possible solutions for them till then reverting this change.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation (add or update README or docs)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #1010, opened 2022-11-30
Actions/cache breaks symlink on Windows runners


FabienPean:
## Summary 

Github actions/cache@v3 does not restore symlink properly on Windows by default. See for example restoring Intel MKL below:

```
11/28/2022  10:54 PM    <DIR>          .
11/28/2022  11:20 PM    <DIR>          ..
11/28/2022  10:54 PM    <DIR>          2022.0.3
11/28/2022  11:20 PM    <SYMLINKD>     latest [\??\C:\Program Files (x86)\Intel\oneAPI\mkl\2022.0.3]
```

## Context

This occurred during an action to cache Intel MKL. The minimal example is reproduced in the following [repository](https://github.com/FabienPean/test-actions-cache/) with the following [script](https://github.com/FabienPean/test-actions-cache/blob/main/.github/workflows/cache.yml)

Digging previous issues, I could find one possible solution by adding the following step:

```
    steps:
    - name: Use GNU Tar from Msys
      if: ${{ runner.os == 'Windows' }}
      run: |
        echo "C:\msys64\usr\bin" >> $env:GITHUB_PATH
```
However this is hardly convenient and was time consuming to first identify the problem and then to find it.

## Differences

Logs are there: https://github.com/FabienPean/test-actions-cache/actions/runs/3586177236/jobs/6035270661

### Caching difference

`C:\Windows\System32\tar.exe --posix -z -cf cache.tgz --exclude cache.tgz -P -C D:/a/test-actions-cache/test-actions-cache --files-from manifest.txt`

vs

`C:\msys64\usr\bin\tar.exe --posix --use-compress-program "zstd -T0" -cf cache.tzst --exclude cache.tzst -P -C D:/a/test-actions-cache/test-actions-cache --files-from manifest.txt --force-local`

### Restoring difference

`C:\Windows\System32\tar.exe -z -xf D:/a/_temp/79e339ad-c303-4c67-aa2c-d0533f9a6395/cache.tgz -P -C D:/a/test-actions-cache/test-actions-cache`

vs 

`C:\msys64\usr\bin\tar.exe --use-compress-program "zstd -d" -xf D:/a/_temp/e365a544-fe05-44e5-86cd-035de603e3e4/cache.tzst -P -C D:/a/test-actions-cache/test-actions-cache --force-local`

## Desired solution

By order of preference:

1. Work by default
2. Input to the action
3. Add documentation

Phantsure:
[...]

PR #1056, opened 2023-01-04
Add support to opt-in enable cross-os caching on windows

Phantsure:
## Description

Enabling cross os support as an opt-in feature on windows. Previously it was given as default in https://github.com/actions/cache/pull/1049. And fixing symlink issue pointed out in here: https://github.com/actions/cache/issues/1010

## Motivation and Context

This change is done to allow using saving and restoring cache on Windows which can be restored or is saved on Linux or Mac runners as **opt-in** feature rather than **default**. This also bring a fix for symlink issue pointed out by a customer here: https://github.com/actions/cache/issues/1010.

## Types of changes

It would invalidate previous caches present on windows runners due to version mismatch.
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -15,7 +15,7 @@ inputs:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
   enableCrossOsArchive:
-    description: 'An optional boolean when enabled, allows windows runners to save/restore caches that can be restored/saved respectively on other platforms'
+    description: 'An optional boolean when enabled, allows windows runners to save or restore caches that can be restored or saved respectively on other platforms'
     default: 'false'
     required: false
 outputs:
--- a/action.yml
+++ b/action.yml
@@ -15,7 +15,7 @@ inputs:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
   enableCrossOsArchive:
-    description: 'An optional boolean enabled to save and restore cache on windows which could be restored and saved on any platform'
+    description: 'An optional boolean when enabled, allows windows runners to save/restore caches that can be restored/saved respectively on other platforms'
     default: 'false'
     required: false
 outputs:
--- a/action.yml
+++ b/action.yml
@@ -14,6 +14,10 @@ inputs:
   upload-chunk-size:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
+  enableCrossOsArchive:
+    description: 'An optional boolean enabled to save and restore cache on windows which could be restored and saved on any platform'
+    default: 'false'
+    required: false
 outputs:
   cache-hit:
     description: 'A boolean value to indicate an exact match was found for the primary key'
--- a/restore/action.yml
+++ b/restore/action.yml
@@ -12,7 +12,7 @@ inputs:
     description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
     required: false
   enableCrossOsArchive:
-    description: 'An optional boolean when enabled, allows windows runners to restore caches that may be saved on other platforms'
+    description: 'An optional boolean when enabled, allows windows runners to restore caches that were saved on other platforms'
     default: 'false'
     required: false
 outputs:
--- a/restore/action.yml
+++ b/restore/action.yml
@@ -12,7 +12,7 @@ inputs:
     description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
     required: false
   enableCrossOsArchive:
-    description: 'An optional boolean enabled to restore cache on windows which could be saved on any platform'
+    description: 'An optional boolean to enable restoring a cache on Windows which was created on another platform'
     default: 'false'
     required: false
 outputs:
--- a/restore/action.yml
+++ b/restore/action.yml
@@ -12,7 +12,7 @@ inputs:
     description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
     required: false
   enableCrossOsArchive:
-    description: 'An optional boolean to enable restoring a cache on Windows which was created on another platform'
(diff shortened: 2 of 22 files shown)


action.yml
```diff
@@ -14,6 +14,10 @@ inputs:
   upload-chunk-size:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
+  enableCrossOsArchive:
+    description: 'An optional boolean enabled to save and restore cache on windows which could be restored and saved on any platform'
```
pdotl, line 18: Suggestion for description: Optional boolean when enabled, allows windows to restore caches created on other platforms.

src/restoreImpl.ts
```diff
@@ -31,11 +31,16 @@ async function restoreImpl(
...
+        const enableCrossOsArchive = utils.getInputAsBool(
+            Inputs.EnableCrossOsArchive
+        );
 
         const cacheKey = await cache.restoreCache(
             cachePaths,
             primaryKey,
-            restoreKeys
+            restoreKeys,
+            {},
```
pdotl, line 42: Just curious, why do we have `{}` here?

action.yml
Phantsure:
```suggestion
    description: 'An optional boolean when enabled, allows to save and restore caches on windows restored and saved on other platforms'
```
This sound better? We need to have save and restore both

src/restoreImpl.ts
Phantsure: Because there is a download options which is not being passed. https://github.com/actions/toolkit/blob/main/packages/cache/src/options.ts#L24

pdotl: Got it. So I think optional parameters such as DownloadOptions need to stay at the end so we don't need to do this. You might have to change the parameter position of `enableCrossOsArchive` to come before this.

action.yml
pdotl:
```suggestion
    description: 'An optional boolean when enabled, allows windows runners to save/restore caches that can be restored/saved respectively on other platforms'
```
True, we need save/restore both. How about this?

Phantsure: Sounds good

src/restoreImpl.ts
Phantsure: That would need change in toolkit. Does this mean we don't want `enableCrossOsArchive` to be optional?

Phantsure: This could possibly be a breaking change for projects using toolkit. Then that would be minor version release

pdotl: Depends on how are we calling it. So there are two types of parameters we can have, 
1. Mandatory parameters that have a default value so explicitly setting them is optional
2. Optional parameters that can be empty

If both are optional parameters that can be empty then this problem is unavoidable.

Phantsure: Going ahead with current approach

action.yml
bishal-pdMSFT, line 18: `Windows` with caps W

restore/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
   restore-keys:
     description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
     required: false
+  enableCrossOsArchive:
+    description: 'An optional boolean enabled to restore cache on windows which could be saved on any platform'
```
bishal-pdMSFT, line 15:
```suggestion
    description: 'An optional boolean to enable restoring a cache on windows which was created on another platform'
```

src/restoreImpl.ts
```diff
@@ -31,11 +31,16 @@ async function restoreImpl(
         const cachePaths = utils.getInputAsArray(Inputs.Path, {
             required: true
         });
+        const enableCrossOsArchive = utils.getInputAsBool(
```
bishal-pdMSFT, line 34: Does `utils.getInputAsBool` assign the default value in case it is not specified by user. Please ensure you have tested the back compat scenario where existing workflows do not have this input set.

wolflikeme200 (approved):

src/restoreImpl.ts
Phantsure: I have added default in `action.yml` so that should work. Would test as well

action.yml
Phantsure: @bishal-pdMSFT Does this sound better or this https://github.com/actions/cache/pull/1056#discussion_r1061408475 ?

restore/action.yml
Phantsure: Would you take a look at above comment by Lovepreet as well?

action.yml
bishal-pdMSFT: Looks good. The word "respectively" does not seem right fit as we are using `/`. But I will agree my english is not so good.

restore/action.yml
bishal-pdMSFT: replied.

restore/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
   restore-keys:
     description: 'An ordered list of keys to use for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.'
     required: false
+  enableCrossOsArchive:
+    description: 'An optional boolean to enable restoring a cache on Windows which was created on another platform'
```
bishal-pdMSFT, line 15: This should change according to the one used for top level action

save/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
   upload-chunk-size:
     description: 'The chunk size used to split up large files during upload, in bytes'
     required: false
+  enableCrossOsArchive:
+    description: 'An optional boolean to enable saving a cache on Windows which may be restored later on another platform'
```
bishal-pdMSFT, line 15: This should change according to the one used for top level action

bishal-pdMSFT (approved):

action.yml
pdotl: We can remove respectively, I'm also not too sure if this would be grammatically correct.

PR #1060, opened 2023-01-06
Release support for cross-os caching as opt-in feature

Phantsure:
Add documentation for cross-os caching

## Description

Make new release for support of cross-os caching as opt-in feature and add documentation around cross-os caching.

## Types of changes

## Checklist:


tips-and-workarounds.md
```diff
@@ -19,23 +19,11 @@ A cache today is immutable and cannot be updated. But some use cases require the
...
-
-The above workaround is also needed if you wish to use cross-os caching since difference of compression algorithms will result in different cache versions for the same cache key. So the above workaround will ensure `zstd` is used for caching on all platforms thus resulting in the same cache version for the same cache key.
+## Cross OS cache
+From `v3.2.3` cache is cross os compatible when `enableCrossOsArchive` input is passed as true. This means that a cache created on `ubuntu-latest` can be used by `windows-latest` provided the workflow which runs on `windows-latest` have input `enableCrossOsArchive` as true. This is useful when you want to cache dependencies which are available on different the OSs. This will help reduce the consumption of the cache quota. Things to keep in mind while using this feature:
+- Only cache those files which are compatible across OSs.
+- Caching symlinks might cause issues while restoration as they work differently on different OSs. 
+- Put runner OS in the cache key if that cache is not meant to be used across OSs to prevent cross-os.
```
pdotl, line 26: Should not have this recommendation here as it might confuse the user with the usage of `enableCrossOsArchive` input. IMO, this could go in the recipes or common use cases @kotewar is working on.

pdotl (commented):
Minor comments.

Aiasp1985P:
*https://commons.m.wikimedia.org/wiki/Archive
<https://commons.m.wikimedia.org/wiki/Archive>*

Aiasp1985P:
8

tips-and-workarounds.md
```diff
@@ -19,23 +19,11 @@ A cache today is immutable and cannot be updated. But some use cases require the
...
-      run: |
-        echo "Adding GNU tar to PATH"
-        echo C:\Program Files\Git\usr\bin>>"%GITHUB_PATH%"
-```
-
-The `cache` action will use GNU tar instead of bsd tar on Windows. This should work on all Github Hosted runners as it is. For self-hosted runners, please ensure you have GNU tar and `zstd` installed.
-
-The above workaround is also needed if you wish to use cross-os caching since difference of compression algorithms will result in different cache versions for the same cache key. So the above workaround will ensure `zstd` is used for caching on all platforms thus resulting in the same cache version for the same cache key.
+## Cross OS cache
+From `v3.2.3` cache is cross os compatible when `enableCrossOsArchive` input is passed as true. This means that a cache created on `ubuntu-latest` can be used by `windows-latest` provided the workflow which runs on `windows-latest` have input `enableCrossOsArchive` as true. This is useful when you want to cache dependencies which are available on different the OSs. This will help reduce the consumption of the cache quota. Things to keep in mind while using this feature:
```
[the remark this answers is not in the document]
Phantsure: Have added that

tips-and-workarounds.md
Phantsure:
```suggestion

```

README.md
```diff
@@ -43,6 +44,7 @@ If you are using this inside a container, a POSIX-compliant `tar` needs to be in
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
 * `key` - An explicit key for restoring and saving the cache
 * `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key.
+* `enableCrossOsArchive` - An optional boolean when enabled, allows Windows runners to save or restore caches that can be restored or saved respectively on other platforms.
```
pdotl, line 47: Nitpicking here but it would good to explicitly mention that default value for this will be false.

pdotl (approved):
LGTM. Just one comment.

README.md
Phantsure: Good catch

PR #1084, opened 2023-01-22
Update caching-strategies.md

kpfleming:
Correct typo; `restore` should be `save'.


kotewar (approved):
Thanks for the correction, I'll merge it when all checks are passed. LGTM.

PR #1095, opened 2023-01-30
Add `check` action

### LINKED ISSUE #831: Allow optionally testing for `cache-hit` without restoring old cache

### LINKED ISSUE #901: Skip download option (allow only to check if cache exists for the given key)

cdce8p:
## Description
Alternative or in addition to #1041 as suggested by @kotewar in https://github.com/actions/cache/pull/1041#issuecomment-1408219377.

Add a dedicated action to check if a cache entry exists without actually downloading it.

> **Note**
> Requires https://github.com/actions/toolkit/pull/1286
>
> This PR currently links a custom version of `actions/toolkit` with the PR changes to be able to test it.

## Motivation and Context
Fixes #901
Fixes #831
https://github.com/actions/cache/discussions/1020#discussioncomment-4332778
https://github.com/actions/cache/discussions/1020#discussioncomment-4341251

If dependency install / cache setup and the actual tests are split across multiple jobs, it's often unnecessary to restore the cache if an exact cache-hit occurred. Similarly, if the cache is created from scratch, it's only necessary to know if a cache entry already exists.

Using `dry-run` in these cases saves time and resources, especially for larger caches which don't need to be restored.

## How Has This Been Tested?
Added test case and run a test workflow using the feature branch.

--

## Types of changes

## Checklist:


kotewar:
@cdce8p, I'll review this shortly and also discuss with my team regarding this PR and your comment on the other one.

This might take some time, but thanks for your contribution again! 😃

cdce8p:
Hi @kotewar 👋🏻, any update? Just saw a new issue pop up #1113.

kotewar:
Sorry for the delay, I'll review the PR today. @cdce8p

cdce8p:
> Sorry for the delay, I'll review the PR today. @cdce8p

Awesome, thanks! Just updated all PRs to include the zstd fix.

PR #1118, opened 2023-02-20
Fix zstd not being used after zstd version upgrade to 1.5.4 on hosted runners

### LINKED ISSUE #1110: Linux -> Mac & enableCrossOsArchive -> Cache not found for input keys
When creating a simple cache of downloaded content in my Linux steps, and caching it, each Linux machine has no problem mounting the created cache and reusing the content.

When the workflow gets to the OSX machine, it does not register a cache hit, and instead creates a new cache instance with the identical key.  

Config:

Linux Machines (happy to use the cache):
```yaml
   - name: Ansible Lint -- Mount Ansible Cache
      uses: actions/cache@v3
      with:
        enableCrossOsArchive: true
        key: ansible-${{ hashFiles('./template/{{cookiecutter.profile_slug}}/profile/requirements.yml') }}-${{ env.CACHE_TTL }}
        path: ansible_cache
```
OSX Machine (creates it's own cache):

```yaml
    - name: Apply Profile -- Mount Ansible Cache
       uses: actions/cache@v3
       with:
         enableCrossOsArchive: true
         key: "ansible-${{ hashFiles('./template/{{cookiecutter.profile_slug}}/profile/requirements.yml') }}-${{ env.CACHE_TTL }}"
          path: ansible_cache
```
Am I assuming functionality that does not exist?

Example Workflow Run: https://github.com/osx-provisioner/profile-generator/actions/runs/4197900167/jobs/7280858264

### LINKED ISSUE #1114: Unexpected cache misses since Feb 17th
I am getting unexpected cache misses using `actions/cache@v3` since yesterday on a scheduled job that runs every 2 hours.

```yaml
    - uses: actions/cache@v3
      with:
        path: folder
        key: folder-${{ env.SUFFIX }}
        restore-keys: |
          folder-
        fail-on-cache-miss: true
```

I am using the option `fail-on-cache-miss: true`, because restoring the cache is critical to the job running, so the action fails with the error:

```
Error: Failed to restore cache entry. Exiting as fail-on-cache-miss is set. Input key: folder-73489456
```

I have not deleted cache entries and I have verified that prefix-matched cache entries exist with:

```shell
gh api \
  -H "Accept: application/vnd.github+json" \
  /repos/OWNER/REPO/actions/caches
```

Further details here: https://github.com/actions/runner/issues/2448
(I wasn't sure where to create this issue as I don't fully understand where the error is, sorry! I'm reporting here as suggested by [this comment](https://github.com/actions/runner/issues/2448#issuecomment-1435668085). If there is a more appropriate place to report this issue please let me know.)

### LINKED ISSUE #1116: "Cache not found" and outputs cache-hit is set to empty string instead of boolean
Restoring claims that cache has not been found:

kuva

But it is found:

kuva

Also as you can see from the comparison, `${{ steps.yarn-cache.outputs.cache-hit }}` is being evaluated to empty string instead of `false` or `true`.

kuva

Runner is ubuntu-22.04 and the cache was created by ubuntu-22.04. It is pure linux-for-linux caching.

pdotl:
## Description

This PR adds changes to fix zstd not being used after zstd version upgrade to 1.5.4 on hosted runners. Please see this [PR](https://github.com/actions/toolkit/pull/1353) in `actions/toolkit` for changes.

## Motivation and Context

Fixes #1110 
Fixes #1114 
Potentially fixes #1116 

## How Has This Been Tested?

Tested with all hosted runners using workflows and ensured all test cases pass on CI.

## Types of changes

## Checklist:

### DIFF
--- a/dist/restore-only/index.js
+++ b/dist/restore-only/index.js
@@ -1152,12 +1152,13 @@ function unlinkFile(filePath) {
     });
 }
 exports.unlinkFile = unlinkFile;
-function getVersion(app) {
+function getVersion(app, additionalArgs = []) {
     return __awaiter(this, void 0, void 0, function* () {
-        core.debug(`Checking ${app} --version`);
         let versionOutput = '';
+        additionalArgs.push('--version');
+        core.debug(`Checking ${app} ${additionalArgs.join(' ')}`);
         try {
-            yield exec.exec(`${app} --version`, [], {
+            yield exec.exec(`${app}`, additionalArgs, {
                 ignoreReturnCode: true,
                 silent: true,
                 listeners: {
@@ -1177,19 +1178,14 @@ function getVersion(app) {
 // Use zstandard if possible to maximize cache performance
 function getCompressionMethod() {
     return __awaiter(this, void 0, void 0, function* () {
-        const versionOutput = yield getVersion('zstd');
+        const versionOutput = yield getVersion('zstd', ['--quiet']);
         const version = semver.clean(versionOutput);
-        if (!versionOutput.toLowerCase().includes('zstd command line interface')) {
-            // zstd is not installed
+        core.debug(`zstd version: ${version}`);
+        if (versionOutput === '') {
             return constants_1.CompressionMethod.Gzip;
         }
-        else if (!version || semver.lt(version, 'v1.3.2')) {
-            // zstd is installed but using a version earlier than v1.3.2
-            // v1.3.2 is required to use the `--long` options in zstd
-            return constants_1.CompressionMethod.ZstdWithoutLong;
-        }
         else {
-            return constants_1.CompressionMethod.Zstd;
+            return constants_1.CompressionMethod.ZstdWithoutLong;
         }
     });
 }
--- a/.licenses/npm/@actions/cache.dep.yml
+++ b/.licenses/npm/@actions/cache.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: "@actions/cache"
-version: 3.1.3
+version: 3.1.4
 type: npm
 summary:
 homepage:
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ See ["Caching dependencies to speed up workflows"](https://docs.github.com/en/ac
 * New actions are available for granular control over caches - [restore](restore/action.yml) and [save](save/action.yml).
 * Support cross-os caching as an opt-in feature. See [Cross OS caching](./tips-and-workarounds.md#cross-os-cache) for more info.
 * Added option to fail job on cache miss. See [Exit workflow on cache miss](./restore/README.md#exit-workflow-on-cache-miss) for more info.
+* Fix zstd not being used after zstd version upgrade to 1.5.4 on hosted runners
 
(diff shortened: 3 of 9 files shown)


bishal-pdMSFT (approved):

DerTim1:
@pdotl Can you please tag that version and also move tag `v3` to this release?

pdotl:
@DerTim1 I have tagged the version and moved it to `v3`.

chetbox:
I re-ran the job that was failing, which seems to have picked up a newer of `action/cache@v3`, but I am still having the same problem.

```
Download action repository 'actions/cache@v3' (SHA:95b455a0fba5a5e1770cc10ded51e5fe99a6737b)
``` 

```
Error: Failed to restore cache entry. Exiting as fail-on-cache-miss is set
```

I tried again with `actions/cache@69d9d449aced6a2ede0bc19182fadc3a0a42d2b0` and it worked.

@pdotl it looks to me like the `v3` tag needs to be moved to `69d9d449aced6a2ede0bc19182fadc3a0a42d2b0`?

pdotl:
@chetbox fixed! 👍

chetbox:
Thanks @pdotl, looks good now. Much appreciated.

atsju:
I had this problem first time today after the fix  : 
`Download action repository 'actions/cache@v3' (SHA:69d9d449aced6a2ede0bc19182fadc3a0a42d2b0)`
`Cache not found for input keys: Windows-mingw810_64`
https://github.com/atsju/DFTFringe/actions/runs/4233014638/jobs/7353437839
And the cache existed. 
Are caches supposed to be invalidated automatically when the action is updated ?

vmarkovtsev:
Our CI has started to fail after the new release. It looks like either some or all files are not restored from the cache now. [Example.](https://github.com/athenianco/athenian-api/actions/runs/4233880928/jobs/7355889511)
We tried to drop all the caches. That didn't help.

cdce8p:
> Are caches supposed to be invalidated automatically when the action is updated ?

That is expected. You'll need to rebuild all caches with `v3.2.6`.
That's because now the action correctly looks for `zstd` caches if it is installed and only if not falls back to the `glib` defaults. Over the last few days the zstd version parsing was broken which caused only `glib` caches to be created.

dist/restore-only/index.js
```diff
@@ -1177,19 +1178,14 @@ function getVersion(app) {
...
             return constants_1.CompressionMethod.Gzip;
         }
-        else if (!version || semver.lt(version, 'v1.3.2')) {
-            // zstd is installed but using a version earlier than v1.3.2
-            // v1.3.2 is required to use the `--long` options in zstd
-            return constants_1.CompressionMethod.ZstdWithoutLong;
-        }
         else {
-            return constants_1.CompressionMethod.Zstd;
+            return constants_1.CompressionMethod.ZstdWithoutLong;
```
JannikWo, line 1188: Sorry for digging out this old PR, I got here following a bug. 
But why are we not checking for the used zstd version here? Shouldn't be  `constants_1.CompressionMethod.Zstd` still used when zstd is above v1.3.2?

PR #876, opened 2022-07-26
fix: fix cache-hit value when cache not found

bchen1029:
related to https://github.com/github/docs/pull/18524

set `cache-hit` to `false` when there is no cache matched

PR #1160, opened 2023-04-24
Mention empty possibility for `outputs.cache-hit`

rikhuijzer:
The `outputs.cache-hit` used to indicate that the value is a boolean, which means `true` or `false`. However, `outputs.cache-hit` is not set when there is no cache restored. I discovered this the hard way and also @aparna-ravindra mentioned it in https://github.com/github/docs/pull/18524#issuecomment-1161626882:

> However, if no cache was restored (either on the primary key or the restore-keys), then the cache-hit output value is not set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 
 ### Outputs
 
-* `cache-hit` - A boolean value to indicate an exact match was found for the key.
+* `cache-hit` - A boolean value to indicate an exact match was found for the key, or empty (not set) if no cache was restored.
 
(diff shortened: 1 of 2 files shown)


rgrunber:
I also discovered this the hard way :rofl: It looks like there's https://github.com/actions/cache/pull/876 to correct the issue ?

Although it's true the documentation at https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-output-of-the-cache-action speaks only about `true`,  a consumer of the action might easily interpret that as just because that's the example that was chosen. Using `true` brings along the implicit assumption that the negation of the precondition (no cache hit) will be `false`. If it's not the case, then that probably deserves a special note in the documentation.

PR #1162, opened 2023-04-27
Bump actions/core to 1.10.0

rentziass:
These solves the workflow warnings related to the use of `set-output` and `save-state`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "author": "GitHub",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   "author": "GitHub",
```
thboop, line 24: please bump the "version" field here on line 2 as appropriate in this file and package-lock.json

thboop (approved):
lgtm

Mater2211 (commented):
[]()

PR #1167, opened 2023-05-04
Include .crates.toml in Rust example

tommy-gilligan:
## Description
Using the existing example, cargo install will error out after restoring a cache:
```
cargo install cargo-tarpaulin
    Updating crates.io index
error: binary `cargo-tarpaulin` already exists in destination
Add --force to overwrite
Error: Process completed with exit code 101.
```

`--force` is undesirable because it will recompile the binary.  Cargo is scared of the existing binary because it doesn't have enough information about it.  Including `~/.cargo/.crates.toml` coaxes it into being okay with the situation.

## Motivation and Context
Caching `cargo install PACKAGE` errors with

```
error: binary `PACKAGE` already exists in destination
```

This can be circumvented by using `--force` but this negates any benefit of using the cache.

## How Has This Been Tested?
I am using the example with this change on my own private repo successfully.

## Types of changes

## Checklist:


LucasFA:
The cargo book also suggests caching `.crates2.json`, too [book](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci). 

Either cache it all or don't include the `bin` directory

PR #1242, opened 2023-09-13
feat: save-always flag

to-s:
This adds the ability to select if the post action step should be run, even if there was another failing step before.

## Description
```
      - name: Load ccache cache
        uses: to-s/cache@main
        with:
          save-always: true
          path: ~/.ccache
          key: ${{ github.job }}-ccache-cache-${{ github.head_ref || github.ref_name }}-${{ github.sha }}
          restore-keys: |
            ${{ github.job }}-ccache-cache-${{ github.head_ref || github.ref_name }}-
            ${{ github.job }}-ccache-cache-
```

## Motivation and Context
If you use ccache to build a bun of unit tests and after that one of these tests fails, the ccache is not stored and everything has to be build again with by the next run.

This can help to save a lot of billing time and increase the sustainability.

## How Has This Been Tested?
By using the code block above.

## Types of changes

## Checklist:


action.yml
```diff
@@ -26,14 +26,18 @@ inputs:
...
+    required: false    
 outputs:
   cache-hit:
     description: 'A boolean value to indicate an exact match was found for the primary key'
 runs:
   using: 'node16'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: success()
+  post-if: "success() || github.event.inputs.save-always"
```
janisz, line 40:
```suggestion
  post-if: "github.event.inputs.save-always"
```

janisz: I'm sorry I should give more context. 
I think we need to revert this change. I was thinking we can pass `success()` as an arg but it will not work. I found a workaround for my case and manually run save restore action with different conditions.

to-s: Reverted with https://github.com/actions/cache/pull/1242/commits/0e9aed0d0d8123bb9d12426b86a761cf9d0c4cff

to-s:
@janisz, anything missing here to get this PR merged?

IanButterworth:
It'd be good to know whether or not/when this is likely to be merged, otherwise we are considering moving away from this action and duplicating its functionality.

IanButterworth:
bump @robherley (based on a new version being released just now)

robherley:
👋 Hey folks, going to merge this in and it should get picked up in the next release

IanButterworth:
Great! Thanks!

Issue #1281, opened 2023-11-15
Add `--remove-files`to avoid running out of space


jlanga:
Adding the `--remove-files` flag would prevent the following error:

```
Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/owner/repo --files-from manifest.txt --use-compress-program zstdmt
zstd: error 70 : Write error : cannot write block : No space left on device 
/usr/bin/tar: cache.tzst: Wrote only 4096 of 10240 bytes
/usr/bin/tar: Child returned status 70
/usr/bin/tar: Error is not recoverable: exiting now
Warning: Failed to save: "/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2
```

This would remove files as they are being added to the tarball.

Cheers!

PR #1282, opened 2023-11-17
Add --remove-files on tarball creation

jlanga:
Add `--remove-files` on tarball creation

## Description
Just added `--remove-files` flag to `tar` on the save section. This flag removes files on the fly as they are being added to the tarball.

## Motivation and Context

When the cache size is too big, the elements to be cached and the tar.gz coexist at the same time, filling the disk space, and failing to produce an object.

This PR solves #1281 .

## How Has This Been Tested?

By using `tar` on a linux machine.
I ran `npm run test` and it passed.

I created a folder and mock files, and after running tar, the folder and the files weren't there.

## Types of changes

## Checklist:

### DIFF
--- a/dist/save-only/index.js
+++ b/dist/save-only/index.js
@@ -36635,7 +36635,7 @@ function getTarArgs(tarPath, compressionMethod, type, archivePath = '') {
         // Method specific args
         switch (type) {
             case 'create':
-                args.push('--posix', '-cf', BSD_TAR_ZSTD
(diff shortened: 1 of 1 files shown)


jlanga:
@bethanyj28  review please?

adazem009:
I'm having the out of space problem as well.
I think you should make this the default when the cache action is used (which automatically saves cache after the workflow completes successfully), but make it optional when the restore and save actions are used because removing the files might break steps after saving.

PR #1308, opened 2024-01-22
Add force-overwrite for cache

PrinsFrank:
## Description
This PR adds a `force-overwrite` options to force overwrite an already existing cache.

## Motivation and Context
When working with very long workflows and incremental changes any tiny change in cache can be helpful. That's why currently I delete a cache when it exists with the following workflow:

```yml
- name: Pull PHPStan cache
  id: pull-phpstan-cache
  uses: actions/cache/restore@v3
  with:
    path: ${{ env.PHPSTAN_CACHE_DIR }}
    key: phpstan-${{ github.ref_name }}
    restore-keys: |
      phpstan

- name: Run PHPStan
  run: composer run stan

- name: Install gh-actions cache plugin
  if: always()
  run: gh extension install actions/gh-actions-cache

- name: Clear previous version of PHPStan branch cache
  if: always()
  continue-on-error: true
  run: gh actions-cache delete phpstan-${{ github.ref_name }} -B ${{ github.ref }} --confirm

- name: Push PHPStan branch cache
  if: always()
  uses: actions/cache/save@v3
  with:
    path: ${{ env.PHPSTAN_CACHE_DIR }}
    key: phpstan-${{ github.ref_name }}
```

For context: without any caching the job will take about 20 minutes. With a main cache and a 'branch cache' that is never updated again, the job will take a few minutes after the first commit, but every next change results in the cache being further out of date and the job taking longer and longer. That's why deleting the existing cache will speed things up significantly.

## How Has This Been Tested?
N/A

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/saveImpl.ts
+++ b/src/saveImpl.ts
@@ -8,6 +8,7 @@ import {
     StateProvider
 } from "./stateProvider";
 import * as utils from "./utils/actionUtils";
+import { issueCommand } from "@actions/core/lib/command";
 
 // Catch and log any unhandled exceptions.  These exceptions can leak out of the uploadChunk method in
 // @actions/toolkit when a failed upload closes the file descriptor causing any in-process reads to
@@ -47,11 +48,16 @@ export async function saveImpl(
         // NO-OP in case of SaveOnly action
         const restoredKey = stateProvider.getCacheState();
 
-        if (utils.isExactKeyMatch(primaryKey, restoredKey)) {
+        const forceOverwrite = utils.getInputAsBool(Inputs.ForceOverwrite);
+        if (utils.isExactKeyMatch(primaryKey, restoredKey) && !forceOverwrite) {
             core.info(
-                `Cache hit occurred on the primary key ${primaryKey}, not saving cache.`
+                `Cache hit occurred on the primary key ${primaryKey} and force-overwrite is disabled, not saving cache.`
             );
             return;
+        }
+
+        if (utils.isExactKeyMatch(primaryKey, restoredKey) && forceOverwrite) {
+            await issueCommand('actions-cache delete', {}, primaryKey);
         }
 
         const cachePaths = utils.getInputAsArray(Inputs.Path, {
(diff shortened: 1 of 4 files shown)


ro0NL (approved):
-

PrinsFrank:
@joshmgross Would you mind reviewing this, or could you point me to the right person for that? This would close #342 which has hundreds of upvotes, mentions and a lot of comments.

szepeviktor (approved):
Many need this.

src/saveImpl.ts
```diff
@@ -47,13 +48,18 @@ export async function saveImpl(
...
+        if (utils.isExactKeyMatch(primaryKey, restoredKey) && !forceOverwrite) {
             core.info(
-                `Cache hit occurred on the primary key ${primaryKey}, not saving cache.`
+                `Cache hit occurred on the primary key ${primaryKey} and force-overwrite is disabled, not saving cache.`
             );
             return;
         }
 
+        if (utils.isExactKeyMatch(primaryKey, restoredKey) && forceOverwrite) {
+            await issueCommand('actions-cache delete', {}, primaryKey);
```
joshmgross, line 60: @PrinsFrank were you able to validate these changes in your own repository?

I'm not entirely sure if the backend supporting the cache action will allow overwriting an existing cache entry like this, and as far as I know this `actions-cache delete` command is not a real command.

PrinsFrank: Sorry, I got confused with the CI options that are available in the actions-cache plugin, but not in the toolkit yet. I use that here: https://github.com/PrinsFrank/CI-PHP/blob/main/.github/workflows/quality.yml#L134, so I can confirm this workflow works, the code just currently doesn't. I wasn't sure how to run this code, but I figured it out and can see this doesn't work in the current form.

If I understand correctly there's currently not any code for this at all in upstream packages, so I tried implementing it in https://github.com/actions/toolkit/pull/1783. As I mentioned before, I'm not a typescript developer so any pointers would be very much appreciated.

ro0NL (approved):

ro0NL:
we want to always save generated cache on main, and inherit in PR's

PR #1325, opened 2024-02-09
Fix save-always/post-if with an output

### LINKED ISSUE #1315: v4.0 save always not working
I have setup the save always to true such that the cache step should always fail even if one of the tasks fail. See screenshots below:
Save always set to true in step:
image

Then further in the job a step fails but the post cache step is not run unlike the post repositry setup:
image

Related to 
https://github.com/actions/cache/pull/1242

bearfriend:
## Description
In v4, the `save-always` input was [added](https://github.com/actions/cache/pull/1242). Unfortunately, it doesn't work because the `inputs` (and most other [contexts](https://github.com/BrightspaceUI/testing/actions/runs/7809704267/job/21303659163#step:16:3) or their [keys](https://github.com/BrightspaceUI/testing/actions/runs/7809704267/job/21303405681#step:16:7)) are [not available](https://github.com/orgs/community/discussions/65593) to `post-if`. Currently, the only thing available to it that can be used to fix this feature is, seemingly, step [outputs](https://github.com/BrightspaceUI/testing/actions/runs/7816145528/job/21322468245#step:16:9).

**Note**: Because outputs require an `id`, so too will using `save-always`. However, the feature was never properly documented and never worked, so I might argue this makes the fix not a breaking change

## Motivation and Context
The `save-always` input to control the `post-if` does not work.
Fixes #1315

## How Has This Been Tested?

All relevant tests have been updated and one new test has been added to test the new output specifically.

The changes have also been tested by running the action from this fork on one of my repos, where the job is intentionally failed, but the cache is [successfully saved](https://github.com/BrightspaceUI/testing/actions/runs/7837422795/job/21394673358#step:15:6), and then [restored](https://github.com/BrightspaceUI/testing/actions/runs/7837422795/job/21394733949#step:3:21) in the next run attempt.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,7 +9,6 @@ on:
     branches:
       - main
       - releases/**
-      - save-always-output
 
 jobs:
   # Build and unit test
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - releases/**
+      - save-always-output
 
 jobs:
   # Build and unit test
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - releases/**
+      - save-always-output
 
 jobs:
   # Build and unit test
@@ -46,10 +47,10 @@ jobs:
       uses: actions/checkout@v3
     - name: Generate files in working directory
       shell: bash
-      run: __tests__/create-cache-files.sh ${{ runner.os }} test-cache
+      run: __tests__/create-cache-files.sh ${{ runner.os }}-save test-cache
     - name: Generate files outside working directory
       shell: bash
-      run: __tests__/create-cache-files.sh ${{ runner.os }} ~/test-cache
+      run: __tests__/create-cache-files.sh ${{ runner.os }}-save ~/test-cache
     - name: Save cache
       uses: ./
       with:
@@ -76,10 +77,51 @@ jobs:
           ~/test-cache
     - name: Verify cache files in working directory
       shell: bash
-      run: __tests__/verify-cache-files.sh ${{ runner.os }} test-cache
+      run: __tests__/verify-cache-files.sh ${{ runner.os }}-save test-cache
     - name: Verify cache files outside working directory
       shell: bash
-      run: __tests__/verify-cache-files.sh ${{ runner.os }} ~/test-cache
+      run: __tests__/verify-cache-files.sh ${{ runner.os }}-save ~/test-cache
+
+  # End to end with save-always
+  test-save-always:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Save cache
+        id: save-always
+        uses: ./
+        with:
+          key: test-${{ runner.os }}-${{ github.run_id }}.${{ github.run_attempt }}
+          path: test-cache
+          save-always: true
+      - name: Generate files
+        shell: bash
+        run: |
+          __tests__/create-cache-files.sh ${{ runner.os }}-save-always test-cache
+          exit 1
+  test-save-always-restore:
+    needs: test-save-always
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Restore cache
+        uses: ./
+        with:
+          key: test-${{ runner.os }}-${{ github.run_id }}.${{ github.run_attempt }}
+          path: test-cache
+      - name: Verify cache
+        run: __tests__/verify-cache-files.sh ${{ runner.os }}-save-always test-cache
 
   # End to end with proxy
   test-proxy-save:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -82,7 +82,7 @@ jobs:
       run: __tests__/verify-cache-files.sh ${{ runner.os }}-save ~/test-cache
 
   # End to end with save-always
-  test-save-always:
+  test-save-always-with-failure:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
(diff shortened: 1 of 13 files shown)


action.yml
```diff
@@ -33,11 +33,13 @@ inputs:
...
   cache-hit:
     description: 'A boolean value to indicate an exact match was found for the primary key'
+  save-always-d18d746b9:
+    description: "Run the post step to save the cache even if another step before fails"
 runs:
   using: 'node20'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: "success() || github.event.inputs.save-always"
+  post-if: "success() || (contains(steps.*.outputs.save-always-d18d746b9, 'true') && !contains(steps.*.outputs.save-always-d18d746b9, 'false'))"
```
bearfriend, line 42: We can't know the step `id`, so check all steps for our unique output, and make sure it has at least one `'true'` value and no `'false'` values.

action.yml
```diff
@@ -33,11 +33,13 @@ inputs:
 outputs:
   cache-hit:
     description: 'A boolean value to indicate an exact match was found for the primary key'
+  save-always-d18d746b9:
```
bearfriend, line 36: Give the new output a unique name to reduce conflicts when checked in `post-if`.

src/restoreImpl.ts
```diff
@@ -12,6 +12,11 @@ import * as utils from "./utils/actionUtils";
 export async function restoreImpl(
     stateProvider: IStateProvider
 ): Promise<string | undefined> {
+    core.setOutput(
+        Outputs.SaveAlways,
+        core.getInput(Inputs.SaveAlways) || "false"
```
bearfriend, line 17: Always fall back to setting an explicit `"false"` value to prevent hijacking.

ddelange:
cc @to-s @bethanyj28

IanButterworth:
Bump. At least CI should be run here?
Making this new feature actually work would be appreciated.

IanButterworth:
bump

IanButterworth:
bump. Please review this. The feature that the breaking change v4.0 was made for does not work. This purports to fix it.

bearfriend:
@bethanyj28 Do you have any idea why we're ok with this feature sitting broken for so long? Can we either get some movement on this PR or just close it? Is there someone else I can reach out to?

Thanks

ruffsl:
@bearfriend , I couldn't get 3e72a6ebe141e2d806ccb28d2696f2680fd84c32 to work. Must this be merged for action version namespace pinning to take effect?

image

bearfriend:
> @bearfriend , I couldn't get [3e72a6e](https://github.com/actions/cache/commit/3e72a6ebe141e2d806ccb28d2696f2680fd84c32) to work. Must this be merged for action version namespace pinning to take effect?

It needs to go through the build process. This branch has done that and should be usable, though I haven't checked in a while:
[bearfriend/cache@save-always-output-build](https://github.com/bearfriend/cache/tree/save-always-output-build)

IanButterworth:
Bump

IanButterworth:
I've opened a community discussion to understand why this is getting no attention https://github.com/orgs/community/discussions/121811

ruffsl:
> If there are any urgent bug fixes or security issues, please let me know and I'll make sure they're prioritized.

- https://github.com/actions/cache/issues/1401#issuecomment-2152904398

@joshmgross , this PR could use some love. Thanks!

joshmgross (commented):
Thanks for your contribution!

Would you be up for adding an integration test in a workflow to validate that `save-always` is working? That would ensure this feature gets fixed and doesn't break in the future.

I'm thinking something along the lines of the `actions/github-script` integration tests that reference the action with `./` to validate the current branch's version.
https://github.com/actions/github-script/blob/main/.github/workflows/integration.yml

Let me know if you want any help setting that up.

bearfriend:
@joshmgross Looks like there were already a similar set of tests in `/.github/workflows/workflow.yml` that they're calling e2e tests. I added one there, but I'm getting an odd result.Perhaps it's expected before merging for a reason I'm not seeing, but the post step [fails to run](https://github.com/bearfriend/cache/actions/runs/9410630913/job/25922622556) on this branch, but [works as expected](https://github.com/bearfriend/cache/actions/runs/9410630913/job/25922622556#step:7:4) on another branch that has had `build` run on it.

Even if I add `needs: build` on this branch it still fails. Any insight into what might be happening?

.github/workflows/workflow.yml
```diff
@@ -76,10 +76,52 @@ jobs:
...
       shell: bash
-      run: __tests__/verify-cache-files.sh ${{ runner.os }} test-cache
+      run: __tests__/verify-cache-files.sh ${{ runner.os }}-save test-cache
     - name: Verify cache files outside working directory
       shell: bash
-      run: __tests__/verify-cache-files.sh ${{ runner.os }} ~/test-cache
+      run: __tests__/verify-cache-files.sh ${{ runner.os }}-save ~/test-cache
+
+  # End to end with save-always
+  test-save-always:
```
joshmgross, line 85:
```suggestion
  # End to end with save-always
  test-save-always-with-failure:
```

What do you think about changing the name here to make it clear that these jobs are expected to fail?

I want to make it clear that the three failing checks are expected

PR Merge box with three failing checks from this job

IanButterworth: Is there any way to do something like a post hook to check that failure occurred and mark as passing? It does seem prone to confusion to have CI marked intentionally as failing

bearfriend: Yep, completely agree.

bearfriend: Yeah I'm a little unclear on how this will ultimately look through the github UI. The status of the workflow in the Actions tab is is passing, but the indicator above for checks on the commit show failing. Maybe that's technically accurate, but still feels a little weird. Either way that's a bigger issue than this bug or this repo.

I don't _think_ there is a way to force the step conclusions to pass without altering the post step to get a `success`, but perhaps Josh knows more.

joshmgross: Agreed it's confusing - I don't think we support anyway to hide these. If you wanted a step to fail but not fail the job itself, you can add `continue-on-error` to the step but that wouldn't allow us to test the `post-if` case where `success()` is false.

action.yml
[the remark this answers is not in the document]
IanButterworth: We use `actions/cache` in a composite action, where we rely on `actions/cache` injecting the post step after the user's code. Moving to require an explicit `actions/cache/save` I believe would not be possible with a composite action, requiring the user to change their workflows, which wouldn't be compatible with our needs.

Here is our PR that's lined up for this feature working https://github.com/julia-actions/cache/pull/107/files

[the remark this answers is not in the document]
ruffsl: > use actions/cache in a composite action, where we rely on actions/cache injecting the post step

I had the same goal, but we may have other issues in common as well:
- https://github.com/actions/runner/issues/1478#issuecomment-1532027544

joshmgross:
https://github.com/actions/cache/issues/1315#issuecomment-2397517875

PR #1217, opened 2023-08-07
Consume latest toolkit and fix dangling promise bug

chkimes:
This PR updates to the latest actions/cache version published from the toolkit, which contains https://github.com/actions/toolkit/pull/1484 and should fix #810.

## Description
See https://github.com/actions/toolkit/pull/1484 for details on the Azure blob fixes. The cache action contains an additional fix for dangling promises that prevents the process from ending despite the cache being fully restored.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (add or update README or docs)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1327, opened 2024-02-15
Fix`fail-on-cache-miss` not working

### LINKED ISSUE #1265: `fail-on-cache-miss` for restore action not failing the workflow
My interpretation of the `fail-on-cache-miss` feature on the `actions/cache` and `actions/cache/restore` actions is that the workflow should stop and be considered a fail if there is no cache found with the given key.

I'm seeing evidence in workflows where using the `actions/cache/restore` action will not fail even if this option is set to `true`.

Snippet from the workflow file:
```yaml
  Acceptance-Linux:
    name: "Acceptance tests (Linux)"
    needs: [Build-Snapshot-Artifacts]
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 #v4.1.0

      - name: Bootstrap environment
        uses: ./.github/actions/bootstrap

      - name: Download snapshot build
        uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 #v3.3.2
        with:
          path: snapshot
          fail-on-cache-miss: true
          key: snapshot-build-${{ github.run_id }}

      - name: Run comparison tests (Linux)
        run: make compare-linux
```

Evidence that the workflow continues past the failed cache restore (failing on the subsequent step, which it should have never reached.
Screenshot 2023-10-20 at 1 27 58 PM

cdce8p:
## Description
`process.exit` overwrites the exit code set by `core.setFailed`, e.g. the workflow would continue even though `fail-on-cache-miss` is set and no cache is found.

Originally introduced in 
- #1217 by @chkimes

/CC  @kgrzebien

## Motivation and Context
Fixes #1265

## How Has This Been Tested?
Created a workflow which should restore a non-existing cache. With `v4.0.0` the workflow continues, with this PR it fails as expected (with `fail-on-cache-miss` set).

## Types of changes

## Checklist:


cdce8p:
/CC @bethanyj28 Would you mind taking a look at this one? It fixes a bug which was just recently introduced.

bethanyj28:
@cdce8p - thanks for the contribution! This seems reasonable to me. Could you add a test case in https://github.com/actions/cache/blob/main/__tests__/restoreImpl.test.ts to validate? Thank you!

cdce8p:
> This seems reasonable to me. Could you add a test case in https://github.com/actions/cache/blob/main/__tests__/restoreImpl.test.ts to validate? Thank you!

I thought about that but I'm not really sure how to do it. The only thing not tested is `process.exit(1);` however from what I've read online it's difficult to test that with `jest`. Tbh I wouldn't consider myself an expert in that area. If you've any pointers for me that would be appreciated.

I already looked at #1217 which first introduced the `process.exit` but AFAICT it didn't add any tests either.

bethanyj28:
👋🏻 @cdce8p - sorry for the delay! While the initial change might've not had tests, it would be really nice to validate in the future that we don't regress here. I believe mocking `process.exit` isn't too bad in jest. Could you do something like this?

```js
test("restore failure with earlyExit should call process exit", async () => {
    // mock `process.exit`
    const processExitMock = jest.spyOn(process, "exit").mockImplementation();

    // simulate a failure
    testUtils.setInput(Inputs.Path, "node_modules");
    const failedMock = jest.spyOn(core, "setFailed");
    const restoreCacheMock = jest.spyOn(cache, "restoreCache");

    // set true for `earlyExit`
    await restoreImpl(new StateProvider(), true);
    expect(restoreCacheMock).toHaveBeenCalledTimes(0);
    expect(failedMock).toHaveBeenCalledWith(
        "Input required and not supplied: key"
    );

    // validate `process.exit` called
    expect(processExitMock).toHaveBeenCalledWith(1);
});
```

Thank you!

bethanyj28 (approved):
Thank you!

cdce8p:
@bethanyj28 Would you mind tagging the `4.0.2` release? It would be awesome if this bug would finally be fixed.
I already updated the version in `package.json` and the changelog in this PR.

bethanyj28:
Definitely! I appreciate you updating the version 🙏🏻

Emmanuelruina:
CACHING ACTION

PR #1340, opened 2024-02-29
Update actions/cache publish flow

bethanyj28:
## Description

Updating the flow to update major versions when a new release is published. This will follow a more paved path for publishing.


robherley (approved):
MVP

PR #1395, opened 2024-05-10
examples: hash go.mod, not go.sum

peterbourgon:
go.sum isn't a lock file and doesn't represent dependencies. See [this issue](https://github.com/dependabot/dependabot-core/issues/4740) for more detail.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -190,7 +190,7 @@ steps:
     path: |
       ~/.cache/go-build
       ~/go/pkg/mod
-    key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+    key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
     restore-keys: |
       ${{ runner.os }}-go-
 ```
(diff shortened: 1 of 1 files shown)


peterbourgon:
Friendly ping :) /cc @bethanyj28 @cdce8p

peterbourgon:
Most ideal would be to define cache keys like [what is described in this comment](https://github.com/actions/setup-go/issues/483#issuecomment-2180656736), but that may be step 2 after this more basic PR 😄

joshmgross:
Thanks for your contribution @peterbourgon!

I'll need to do a bit of research and I'll get back to you on this change.

joshmgross:
While I understand that `go.sum` is not a lockfile, it does contain module checksums that could prevent tampering. From that perspective, it provides a more complete picture of the dependencies than `go.mod`.

```shell
❯ go help modules
Modules are how Go manages dependencies.

A module is a collection of packages that are released, versioned, and
distributed together. Modules may be downloaded directly from version control
repositories or from module proxy servers.

For a series of tutorials on modules, see
https://golang.org/doc/tutorial/create-module.

For a detailed reference on modules, see https://golang.org/ref/mod.

By default, the go command may download modules from https://proxy.golang.org.
It may authenticate modules using the checksum database at
https://sum.golang.org. Both services are operated by the Go team at Google.
The privacy policies for these services are available at
https://proxy.golang.org/privacy and https://sum.golang.org/privacy,
respectively.

The go command's download behavior may be configured using GOPROXY, GOSUMDB,
GOPRIVATE, and other environment variables. See 'go help environment'
and https://golang.org/ref/mod#private-module-privacy for more information.
```

Issue #1334, opened 2024-02-23
`cache-hit` undefined on cache miss?


aryairani:
The README says:

> **Note** `cache-hit` will only be set to `true` when a cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`.

but `outputs.cache-hit == 'false'` doesn't seem to be true when I get a cache miss, and I noticed that the examples all use `!= 'true'`, not `== 'false'`.

I think [here](https://github.com/actions/cache/blob/main/src/restoreImpl.ts#L65) we can see `restoreImpl` return early on a miss, without setting any value for `cache-hit` at all. I'm not a TypeScript or Actions expert though.

In the meantime, I'll switch my `== 'false'` to `!= 'true'` but I feel dirty. :)

pddg:
This problem does not seem to occur in v3, only in v4. In fact, the repositories that had been delayed updating to v4 were working fine.
In workflows that saving the cache when `outputs.cache-hit == 'false'`, the cache was no longer being saved. I recognized this too late because it simply skips a step and does not result in an error.
Is this a bug? Or is it a breaking change?

RalfJung:
I'm seeing a potentially related problem, where our steps that have `${{ steps.cache.outputs.cache-hit != 'true' }}` now always get executed since the switch to v4. With v3, they used to be properly skipped on a cache hit, but I have not seen them skipped a single time since then.

aryairani:
@RalfJung: I was going to say don't use `if: ${{ steps.cache.outputs.cache-hit != 'true' }}` because that's a string that will always evaluate to `true`, not a boolean expression, but if it eventually worked for you, then I have no idea how anything works.

aryairani:
@RalfJung: I was going to say don't use `if: ${{ steps.cache.outputs.cache-hit != 'true' }}` because that's a string that will always evaluate to `true`, not a boolean expression, and that you should use `if: steps.cache.outputs.cache-hit != 'true` instead.

But if it eventually worked for you, then I have no idea how anything works.

no-yan:
It looks like there are document revisions in progress related to this issue.
https://github.com/actions/cache/pull/1263

joshmgross:
This is being reverted in #1467 as workflows are already dependent on `cache-hit` being undefined - #1466

PR #1263, opened 2023-10-17
README: clarify that `cache-hit` returns a string and not a boolean

andreasabel:
Rationale: if `cache-hit` was a boolean, then `!` could be used to get the opposite.
However, it is really a string, and this should be pointed out clearly in the documentation.
Because both `!'true'` and `!'false'` are `false`, so customers can shoot themselves in the foot here.

Closes #1262.

PR #1404, opened 2024-05-22
Fix cache-hit output when cache missed

fchimpan:
## Description

fix https://github.com/actions/cache/issues/1334

According to the [README](https://github.com/actions/cache?tab=readme-ov-file#outputs), if the cache was not found, the output should be `'false'`. However, it actually returned an empty string. This seems to be due to the behavior of GitHub Actions to return an empty string if the output is not defined. Please see [these conversations](https://github.com/actions/cache/pull/1263).

I would like to change the code to return `'false'` when the cache is not found, per the README.

## Motivation and Context

In PR #1263, a README correction is suggested. This is a great initiative, thank you. However, to avoid confusion, how about changing the implementation instead?

## How Has This Been Tested?

The issue was caused by not defining an output for `cache-hit` when the cache was not found, and then returning. Therefore, I will set the output before returning.

## Types of changes

## Checklist:

### DIFF
--- a/src/restoreImpl.ts
+++ b/src/restoreImpl.ts
@@ -51,6 +51,7 @@ export async function restoreImpl(
         );
 
         if (!cacheKey) {
+            core.setOutput(Outputs.CacheHit, false.toString());
             if (failOnCacheMiss) {
                 throw new Error(
                     `Failed to restore cache entry. Exiting as fail-on-cache-miss is set. Input key: ${primaryKey}`
@@ -62,7 +63,6 @@ export async function restoreImpl(
                     ...restoreKeys
                 ].join(", ")}`
             );
-            core.setOutput(Outputs.CacheHit, false.toString());
             return;
         }
 
--- a/src/restoreImpl.ts
+++ b/src/restoreImpl.ts
@@ -62,7 +62,7 @@ export async function restoreImpl(
                     ...restoreKeys
                 ].join(", ")}`
             );
-
+            core.setOutput(Outputs.CacheHit, false.toString());
             return;
         }
 
--- a/__tests__/restore.test.ts
+++ b/__tests__/restore.test.ts
@@ -260,7 +260,7 @@ test("Fail restore when fail on cache miss is enabled and primary + restore keys
     );
 
     expect(stateMock).toHaveBeenCalledWith("CACHE_KEY", key);
-    expect(setCacheHitOutputMock).toHaveBeenCalledTimes(0);
+    expect(setCacheHitOutputMock).toHaveBeenCalledTimes(1);
 
     expect(failedMock).toHaveBeenCalledWith(
         `Failed to restore cache entry. Exiting as fail-on-cache-miss is set. Input key: ${key}`
--- a/__tests__/restoreOnly.test.ts
+++ b/__tests__/restoreOnly.test.ts
@@ -86,7 +86,8 @@ test("restore with no cache found", async () => {
     );
 
     expect(outputMock).toHaveBeenCalledWith("cache-primary-key", key);
-    expect(outputMock).toHaveBeenCalledTimes(1);
+    expect(outputMock).toHaveBeenCalledWith("cache-hit", "false");
+    expect(outputMock).toHaveBeenCalledTimes(2);
     expect(failedMock).toHaveBeenCalledTimes(0);
 
     expect(infoMock).toHaveBeenCalledWith(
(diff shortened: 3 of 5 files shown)


fchimpan:
Hi, @bethanyj28 
I would like to request a reviewer for this pull request. If you are not the maintainer, could you please recommend a suitable reviewer?

Thank you very much!

fchimpan:
@joshmgross
I would like to request a reviewer for this pull request. If you are not the reviewer, could you please recommend a suitable reviewer?

Thank you very much!

src/restoreImpl.ts
```diff
@@ -62,7 +62,7 @@ export async function restoreImpl(
                     ...restoreKeys
                 ].join(", ")}`
             );
-
+            core.setOutput(Outputs.CacheHit, false.toString());
```
joshmgross, line 65: Should we set this output before the `if (failOnCacheMiss)` case above?

That would ensure it's set even if `fail-on-cache-miss` is true.

fchimpan: Hi, @joshmgross 

Thanks for your response.

Indeed, since it applies even in the case of exceptions, it seems good to output before `failOnCacheMiss`. What do you think?
[6f5b5ee](https://github.com/actions/cache/pull/1404/commits/6f5b5ee7292482e1a430a02aa545bd3f446712fe)

joshmgross: Thanks @fchimpan!

Looks like the `dist/` is out of date - could you re-run `npm run build`?

https://github.com/actions/cache/actions/runs/9599939755/job/26475812160?pr=1404

fchimpan: Thanks @joshmgross !
[8addae5](https://github.com/actions/cache/pull/1404/commits/8addae5ffffa871c88fbd2d8a9a43af45f069161)
I have pushed the dist.

fchimpan:
I tested the implementation of this PR by actually using it. As expected, it was confirmed that ‘false’ is returned when the cache is not found.

v4 test : https://github.com/fchimpan/cache-pr-1404-test/actions/runs/9606510227
This PR test : https://github.com/fchimpan/cache-pr-1404-test/actions/runs/9606510228

fchimpan:
Hi @joshmgross ,
I’ve made the necessary changes. Could you please review the code?

Thank you!

v4hn:
Comment: The ternary logic which this patch removes was a useful (but -granted- undocumented and broken) API to detect whether `restore-keys` was used.
E.g.,
```
steps:
- name: Restore Cache
  id: restore
  uses: actions/cache/restore@v4
  with:
    path: ...
    key: abc-${{ github.sha }}-${{ github.run_id }}-${{ github.run_attempt }}
    restore-keys: |
      abc-${{ github.sha }}-${{ github.run_id }}
- name: Run if any cache was restored
  if: steps.restore.outputs.cache-hit != ''
  ...
```

The correct way to get the same check seems to be
`if: steps.restore.outputs.cache-matched-key != ''`.
Maybe it makes sense to add that use-case in the README?

Jason3S:
I think it would have been better to change the documentation. This was a major breaking change.

Jason3S:
Before landing this type of change, it would be useful to check how it is used:
https://github.com/search?q=.cache-hit+language%3AYaml+&type=code

Ari-Kishon:
This was a major change and should have been released as such.
The change in return value cause several critical steps in our workflows to be skipped and caused them all to fail.

This change could have potentially disrupted development for my whole team had I not caught it before going on vacation.

We rely heavily on actions from this organization because we perceive them be stable and professionally maintained.

stof:
The readme says that `chache-hit` is a boolean value: https://github.com/actions/cache?tab=readme-ov-file#outputs

The string `'false'` is **not** a boolean value (and it is a truthy value). Please revert this broken change

v4hn:
> The readme says that `chache-hit` is a boolean value: https://github.com/actions/cache?tab=readme-ov-file#outputs
> 
> The string `'false'` is **not** a boolean value (and it is a truthy value). Please revert this broken change

In that sense the API was inconsistent with the docs already before.
When a `restore-keys` entry was found instead of the primary key, `'false'` was already returned before and a falsy value was only returned in case neither key was found.

stof:
Then the readme correction proposed in https://github.com/actions/cache/pull/1263 is still relevant, even though the description of this PR says that it is meant to fix the implementation instead of having to clarify the readme (the PR of course needs to be adapted to account for the change done here)

nati-levi:
This broke our CI.
Should have been major release. 
Its a breaking change.

AnimMouse:
This basically broke all of my composite actions.
It's sad that it's 2024, and GitHub Actions still can't make their [type system](https://en.wikipedia.org/wiki/Type_system) right!
`true` and `false` are booleans, not strings!

joshmgross:
👋 I will work on reverting this change

PR #1434, opened 2024-07-18
Clarify that the `restore-keys` input is a string in the docs 

### LINKED ISSUE #1173: Incorrect documentation for `restore-keys`
https://github.com/actions/cache/blob/04f198bf0b2a39f7230a4304bf07747a0bddf146/README.md#L52

This turns out to be slightly but importantly wrong: the format is actually a _multi-line string_ with one key per line; attempting to supply an actual list of keys will fail to validate (c.f. https://github.com/python/cpython/pull/104468#pullrequestreview-1425935982).  I think the just updating the docs is probably better than switching to the currently-documented behavior.

x612skm:
Previously the description of `restore-keys` confusingly implied that it might accept a YAML list, which is wrong. This patch corrects that to say, it's a string. 

Closes #1173

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 
 * `key` - An explicit key for a cache entry. See [creating a cache key](#creating-a-cache-key).
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
-* `restore-keys` - An ordered multiline string listing the prefix-matched keys,that are used for restoring stale cache if no cache hit occurred for key.
+* `restore-keys` - An ordered multiline string listing the prefix-matched keys, that are used for restoring stale cache if no cache hit occurred for key.
 * `enableCrossOsArchive` - An optional boolean when enabled, allows Windows runners to save or restore caches that can be restored or saved respectively on other platforms. Default: `false`
 * `fail-on-cache-miss` - Fail the workflow if cache entry is not found. Default: `false`
 * `lookup-only` - If true, only checks if cache entry exists and skips download. Does not change save cache behavior. Default: `false`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 
 * `key` - An explicit key for a cache entry. See [creating a cache key](#creating-a-cache-key).
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
-* `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key.
+* `restore-keys` - An ordered multiline string listing the prefix-matched keys,that are used for restoring stale cache if no cache hit occurred for key.
```
webknjaz, line 57:
```suggestion
* `restore-keys` - An ordered multiline string listing the prefix-matched keys, that are used for restoring stale cache if no cache hit occurred for key.
```

x612skm: Addressed it here : https://github.com/actions/cache/pull/1434/commits/8c968f5777efc8aeef46f5b639df80fa502b294c

x612skm:
@bethanyj28, mind taking a look at this PR?

webknjaz (approved):

joshmgross (approved):

README.md
```diff
@@ -54,7 +54,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 
 * `key` - An explicit key for a cache entry. See [creating a cache key](#creating-a-cache-key).
 * `path` - A list of files, directories, and wildcard patterns to cache and restore. See [`@actions/glob`](https://github.com/actions/toolkit/tree/main/packages/glob) for supported patterns.
-* `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key.
+* `restore-keys` - An ordered multiline string listing the prefix-matched keys, that are used for restoring stale cache if no cache hit occurred for key.
```
joshmgross, line 57: Can you update https://github.com/actions/cache/blob/e47d9f9ec8be3ac771b0b03c8e50f86e1c2fdea8/restore/action.yml#L11-L13 and https://github.com/actions/cache/blob/e47d9f9ec8be3ac771b0b03c8e50f86e1c2fdea8/action.yml#L11-L13 to match this new description?

x612skm: Done @joshmgross Mind taking a relook?

joshmgross (approved):

PR #1439, opened 2024-07-23
examples: Add opam example

rusty-key:
Successfully set up opam dependency caching, sharing the config in case it is useful to someone else

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -24,6 +24,7 @@
 - [Node - Yarn](#node---yarn)
 - [Node - Yarn 2](#node---yarn-2)
 - [OCaml/Reason - esy](#ocamlreason---esy)
+- [OCaml/Reason - opam](#ocamlreason---opam)
 - [PHP - Composer](#php---composer)
 - [Python - pip](#python---pip)
   - [Simple example](#simple-example)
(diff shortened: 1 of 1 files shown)


examples.md
```diff
@@ -416,6 +417,30 @@ Esy allows you to export built dependencies and import pre-built dependencies.
       if: steps.restore-cache.outputs.cache-hit != 'true'
 ```
 
+## OCaml/Reason - opam
+
+This example requires you to have `opam.locked` file which you can generate with `opam lock .`.
+
+It's worth noting that even if lock file was change, it might be optimal to restore previous cache, so `hashFiles` in the key and if statement in "Save opam cache" step are optional.
+```yaml
```
joshmgross, line 425: Are the separate steps for `restore` and `save` needed here?

The default action will add a save post step automatically for you and handle checking the `cache-hit` output.

```yaml
    - name: Cache opam
      uses: actions/cache@v4
      with:
        path: _opam
        key: ${{ runner.os }}-opam-${{ hashFiles('opam.locked') }}

    - name: Install dependencies
      run: opam install . --locked --deps-only -y
```

PR #1450, opened 2024-08-13
fix-save-always

barnuri:
fix save-always condition

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -37,7 +37,7 @@ runs:
   using: 'node20'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: "success() || github.event.inputs.save-always"
+  post-if: "success() || github.event.inputs.save-always == 'true' || github.event.inputs.save-always == true"
 branding:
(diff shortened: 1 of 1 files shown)


barnuri:
@joshmgross can you please look ? the condition is not working

action.yml
```diff
@@ -37,7 +37,7 @@ runs:
   using: 'node20'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: "success() || github.event.inputs.save-always"
+  post-if: "success() || github.event.inputs.save-always == 'true' || github.event.inputs.save-always == true"
```
joshmgross, line 40: This is not the reason `save-always` is broken.

`github.event.inputs.save-always` is not a valid context access.

See https://github.com/actions/cache/pull/1325#discussion_r1636968780 and https://github.com/actions/cache/issues/1315#issuecomment-2163822166 for more details.

joshmgross:
https://github.com/actions/cache/issues/1315#issuecomment-2397517875

PR #1456, opened 2024-08-29
Add Bun example

idleberg:
## Types of changes

## Checklist:

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -1,5 +1,6 @@
 # Examples
 
+- [Bun](#bun)
 - [C# - NuGet](#c---nuget)
 - [Clojure - Lein Deps](#clojure---lein-deps)
 - [D - DUB](#d---dub)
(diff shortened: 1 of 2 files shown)


examples.md
```diff
@@ -41,6 +42,27 @@
...
+
+### Windows
+
+```yaml
+- uses: actions/cache@v4
+  with:
+    path: |
+      ~\.deno
+      ~\AppData\Local\deno
+    key: ${{ runner.os }}-deno-${{ hashFiles('**/deps.ts') }}
```
joshmgross, line 63: Does Bun on Windows not have a lockfile?

examples.md
```diff
@@ -41,6 +42,26 @@
 - [Swift - Mint](#swift---mint)
 - [* - Bazel](#---bazel)
 
+## Bun
+
+```yaml
+- uses: actions/cache@v4
+  with:
+    path: |
+      ~/.bun/install/cache
```
PaulRBerg, line 51: Why not add `node_modules` to the list of paths, @idleberg @joshmgross?

RiskyMH: @PaulRBerg `node_modules` with Bun is just hardlinks to the aforementioned `~/.bun/install/cache` like PNPM to use less space on disk.

However Bun is usually faster than the cache system, so I wouldn't recommend using it: https://github.com/oven-sh/setup-bun/issues/14#issuecomment-1714116221

PaulRBerg: haha interesting, thanks for your answer!

Issue #1466, opened 2024-10-07
Cache-Hit set to string, not boolean, as of v4.1.0


bigdogwillfeed:
After a recent upgrade, our runners started behaving differently on cache miss.

As of 0c45773b623bea8c8e75f6c82b208c3cf94ea4f9, a step defined with `if: steps.cache-release.outputs.cache-hit` would be skipped if the "cache release" job missed the cache, as intended. As of 2cdf405574d6ef1f33a1d12acccd3ae82f47b3f2, the job now runs on cache miss, despite clearly missing the cache.

Here's a relevant snippet of the run with debug logging enabled
```
Cache not found for input keys: release-0e0ceb0ce2360a93ec77b898ac06177b7429666b
##[debug]Node Action run completed with exit code 0
##[debug]Save intra-action state CACHE_KEY = release-0e0ceb0ce2360a93ec77b898ac06177b7429666b
##[debug]Set output cache-hit = false
##[debug]Finishing: Cache release folder
<snip>
##[debug]Evaluating condition for step: 'Panic if build already exists'
##[debug]Evaluating: (success() && steps.cache-release.outputs.cache-hit)
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating Index:
##[debug]....Evaluating Index:
##[debug]......Evaluating Index:
##[debug]........Evaluating steps:
##[debug]........=> Object
##[debug]........Evaluating String:
##[debug]........=> 'cache-release'
##[debug]......=> Object
##[debug]......Evaluating String:
##[debug]......=> 'outputs'
##[debug]....=> Object
##[debug]....Evaluating String:
##[debug]....=> 'cache-hit'
##[debug]..=> 'false'
##[debug]=> 'false'
##[debug]Expanded: (true && 'false')
##[debug]Result: 'false'
##[debug]Starting: Panic if build already exists
```

madebr:
I'm seeing the same bad behavior.
As a workaround, I currently added a string comparison to my workflows:
```
      if: ${{ (!steps.cache-restore.outputs.cache-hit || steps.cache-restore.outputs.cache-hit == 'false') }}
```

Jason3S:
This is cause by #1404, which was a major breaking change.
It is not clear why the change was necessary.

joshmgross:
This should be fixed in https://github.com/actions/cache/releases/tag/v4.1.1, the `v4` tag will be updated soon.

[...]

Issue #1262, opened 2023-10-16
`!cache-hit` is `false` when `cache-hit` is `false`


andreasabel:
This is the workflow I caught red-handed: https://github.com/andreasabel/hasktorch/actions/runs/6536754694/job/17749144021
1. Version: `actions/cache@v3` SHA:704facf57e6136b1bc63b828d79edcd491f0ee84 https://github.com/andreasabel/hasktorch/actions/runs/6536754694/job/17749144021
2. Cache restored from `restore-key` https://github.com/andreasabel/hasktorch/actions/runs/6536754694/job/17749144021#step:7:4
   ```
    Run actions/cache/restore@v3
    with:
      path: ~/.cabal/store
      key: macOS-cabal-3.6.2.0-ghc-9.2.8-plan-813f23bfbcb4c800c8178d743c3df6e501ec38580a651c180dcd99b0abc9319d
      restore-keys: macOS-cabal-3.6.2.0-ghc-9.2.8-
    ...
   Cache restored from key: macOS-cabal-3.6.2.0-ghc-9.2.8-813f23bfbcb4c800c8178d743c3df6e501ec38580a651c180dcd99b0abc9319d
   ```
   Note that this key is not the primary key, which has an additional `-plan` before the SHA.
3. The step "Cache dependencies" (`if: !steps.cache.outputs.cache-hit`) is skipped", it's code is https://github.com/andreasabel/hasktorch/actions/runs/6536754694/workflow
   ```yaml
    - name: Cache dependencies
      if:   ${{ !steps.cache.outputs.cache-hit }}
      uses: actions/cache/save@v3
      with:
        path: ~/.cabal/store
        key: ${{ steps.cache.outputs.cache-primary-key }}
   ```

According to the docs, `outputs.cache-hit` should not be `true` if the cache was restored with a fallback key.
This is stressed here:
https://github.com/actions/cache/blob/704facf57e6136b1bc63b828d79edcd491f0ee84/README.md?plain=1#L63-L65
And again for the `restore` action:
https://github.com/actions/cache/blob/704facf57e6136b1bc63b828d79edcd491f0ee84/restore/README.md?plain=1#L15-L22

andreasabel:
Update: Debug printing shows that actually, it is not that `cache-hit` is `true`:
https://github.com/andreasabel/hasktorch/actions/runs/6537791407/job/17757232711#step:7:8229
```
[...]

PR #1467, opened 2024-10-08
Restore original behavior of `cache-hit` output

joshmgross:
- #1262
- #1466

As described in #1466, #1404 broke workflows that relied on the previous behavior of `cache-hit` being an empty string when there was a cache miss:

```
if: ${{ steps.cache-restore.outputs.cache-hit }}
```

If `cache-hit` is an empty string, this would work as expected. When cache-hit is `false`, this is a non-empty string and treated as "true".

Actions outputs do not support proper boolean values, and we can't guarantee how users are interpreting this string output.

This PR reverts #1404 and updates the README to clarify all possible values of `cache-hit`:
1. 'true' if there's an exact match
2. 'false' if there's a cache hit with a restore key
3. '' if there's a cache miss

This is likely confusing to users, but we should maintain the existing behavior to avoid breaking existing workflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 
 ### Outputs
 
-* `cache-hit` - A boolean value to indicate an exact match was found for the key.
-
-    > **Note** `cache-hit` will only be set to `true` when a cache hit occurs for the exact `key` match. For a partial key match via `restore-keys` or a cache miss, it will be set to `false`.
+* `cache-hit` - A string value to indicate an exact match was found for the key.
+  * If there's a cache hit, this will be 'true' or 'false' to indicate if there's an exact match for `key`.
+  * If there's a cache miss, this will be an empty string.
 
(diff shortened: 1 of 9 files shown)


robherley (approved):

Jason3S:
@joshmgross 
Thank you 🙏

AnimMouse:
@joshmgross
Thank you!

ulgens:
> This is likely confusing to users, but we should maintain the existing behavior to avoid breaking existing workflows.

I can see the reasoning here but this shouldn't mean a weird behaviour should be kept just because it's breaking. IMO having it in the release notes and making the necessary change in version number should be enough, which was already done. Having this PR merged feels like a step in the wrong direction and there should be a plan to properly handle that breaking change.

joshmgross:
> IMO having it in the release notes and making the necessary change in version number should be enough, which was already done.

Per https://semver.org/, a breaking change needs to be a major version bump (i.e. `v5` of this action).

I agreed that we should fix this confusing behavior, but we're not ready to create a new major version of this action.
https://github.com/actions/cache/issues/1466#issuecomment-2400480945

PR #1486, opened 2024-10-22
Add releaseNewActionVersion environment gate to prevent accidental releases

Link-:
## Description

This pull request includes changes to the GitHub Actions workflow configuration to improve the publishing process. The most important changes are the addition of an environment for the publish job and some formatting adjustments.

## Motivation and Context

We don't want accidental and non-gated releases of immutable action packages. `releaseNewActionVersion` is a gated environment and will require approval before release.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/publish-immutable-actions.yml
+++ b/.github/workflows/publish-immutable-actions.yml
@@ -6,6 +6,8 @@ on:
 
 jobs:
   publish:
+    environment:
+      name: releaseNewActionVersion
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -15,6 +17,7 @@ jobs:
     steps:
       - name: Checking out
         uses: actions/checkout@v4
+      
       - name: Publish
         id: publish
         uses: actions/publish-immutable-action@0.0.3


joshmgross:
cc @Jcambass - I'm curious if we should standardize this process across the actions org

Jcambass:
👋🏻 It's the first time I'm hearing about this.

Can you elaborate a bit how the UX looks like with these changes?

Ideally the don't need this as we really want each release to correspond to an immutable package. So, if anything, this extra confirmation should be required before creating the release.

That being said, I do understand that a release can be deleted and re-published while and immutable action version cannot.


I'll leave this decision up to you. I'm also unsure about making this standard. First, looking at most first party actions, I don't see a real way to standardise things. Configs will drift, so we at most could make it a best practise.

More importantly though I wonder if the risk of accidental publishing is really greater than the risk of not realising that publishing the release will create this action that needs approval. This could lead to a drift in available release vs available immutable actions. Something we definitely need to avoid for the time being.

Happy to discuss though as I don't have a good understanding of how the UX of this would look like :) 

Also sorry for the late response here. This got lost in my notifications.

PR #1525, opened 2024-12-28
docs: Make the "always save prime numbers" example more clear

Tobbe:
## Description
Update the ["Always save cache" example](https://github.com/actions/cache/tree/main/save#always-save-cache) to make it clear that it's talking about prime numbers

## Motivation and Context
I was dropped into that docs page without much context from our GitHub actions CI output
```
Warning: Input 'save-always' has been deprecated with message: save-always does not work as intended and will be removed in a future release.
A separate `actions/cache/restore` step should be used instead.
See https://github.com/actions/cache/tree/main/save#always-save-cache for more details.
```
And when I just looked at the example it wasn't immediately clear to me what "Primes" was. If it was "primes" as in "This action 'primes' the cache" or if it was "primes" as in "prime numbers". 
Reading it through a few times I'm guessing it's the latter; prime numbers. So this PR updates the docs to remove the ambiguity.

## How Has This Been Tested?

Not tested

N/A

## Types of changes

## Checklist:

### DIFF
--- a/save/README.md
+++ b/save/README.md
@@ -79,8 +79,10 @@ To avoid saving a cache that already exists, the `cache-hit` output from a resto
 The `cache-primary-key` output from the restore step should also be used to ensure
 the cache key does not change during the build if it's calculated based on file contents.
 
+Here's an example where we imagine we're calculating a lot of prime numbers and want to cache them:
+
 ```yaml
-name: Always Caching Primes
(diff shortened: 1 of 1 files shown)


joshmgross (approved):
Thanks!

I can definitely see the confusion between the verb "primes" and the noun "primes" here.

Issue #1570, opened 2025-03-11
Feat: Control whether caches are written


helly25:
In some cases you want to read but not save a new cache. For instance when you run workflows on a tag action, then you probably want to read caches from the main/default branch, but you most likely do not want to create new caches. The current solution is to separate restore and save actions which is cumbersome.

In https://github.com/actions/cache/pull/1452 the `always_save` feature was removed. 

I propose to add a new config `save-on-success` that defaults to true and can be set false to prevent cache writing. Now this new value can be set from available context - just like you could when separating the steps - granted the decision must be available upfront. That above described decision is in fact available immediately on job creation. So we can use the new setting to simplify the setup.

helly25:
A proposal to implement this is in https://github.com/actions/cache/pull/1571

helly25:
See also https://github.com/actions/cache/discussions/1598

iwahbe:
This would be very helpful to have.

PR #1571, opened 2025-03-11
Add new setting save-on-success.

helly25:
Proposal to implement https://github.com/actions/cache/issues/1570

## Description

Allows to control whether new caches can be written in post action on success.

## Motivation and Context

In some cases you want to read but not save a new cache. For instance when you run workflows on a tag action, then you probably want to read caches from the main/default branch, but you most likely do not want to create new caches. The current solution is to separate restore and save actions which is cumbersome.

In #1452 the always_save feature was removed.

I propose to add a new config save-on-success that defaults to true and can be set false to prevent cache writing. Now this new value can be set from available context - just like you could when separating the steps - granted the decision must be available upfront. That above described decision is in fact available immediately on job creation. So we can use the new setting to simplify the setup.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 * `enableCrossOsArchive` - An optional boolean when enabled, allows Windows runners to save or restore caches that can be restored or saved respectively on other platforms. Default: `false`
 * `fail-on-cache-miss` - Fail the workflow if cache entry is not found. Default: `false`
 * `lookup-only` - If true, only checks if cache entry exists and skips download. Does not change save cache behavior. Default: `false`
+* `save-on-success` - If true, then the cache is written in the post action on success, or (if false) the cache will only be restored if available.
 
 #### Environment Variables
 
--- a/action.yml
+++ b/action.yml
@@ -26,13 +26,18 @@ inputs:
     description: 'Check if a cache entry exists for the given input(s) (key, restore-keys) without downloading the cache'
     default: 'false'
     required: false
+  save-on-success:
+    description: 'Whether the cache is written in the post action on success or (if false) is only restored'
+    default: 'true'
+    required: false
   save-always:
     description: 'Run the post step to save the cache even if another step before fails'
     default: 'false'
     required: false
     deprecationMessage: |
       save-always does not work as intended and will be removed in a future release.
-      A separate `actions/cache/restore` step should be used instead.
+      If you only want to control whether a new cache will be written use `save-always` instead.
+      Otherwise a separate `actions/cache/restore` step should be used instead.
       See https://github.com/actions/cache/tree/main/save#always-save-cache for more details.
 outputs:
   cache-hit:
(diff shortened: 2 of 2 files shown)


eirnym (approved):

Rossbro2:
What is the status of this?  This could really improve our workflows!

README.md
```diff
@@ -76,6 +76,7 @@ If you are using a `self-hosted` Windows runner, `GNU tar` and `zstd` are requir
 * `enableCrossOsArchive` - An optional boolean when enabled, allows Windows runners to save or restore caches that can be restored or saved respectively on other platforms. Default: `false`
 * `fail-on-cache-miss` - Fail the workflow if cache entry is not found. Default: `false`
 * `lookup-only` - If true, only checks if cache entry exists and skips download. Does not change save cache behavior. Default: `false`
+* `save-on-success` - If true, then the cache is written in the post action on success, or (if false) the cache will only be restored if available.
```
Rossbro2, line 79:
```suggestion
* `save-on-success` - If true, then the cache is written in the post action on success, or (if false) the cache will only be restored if available. Default: `true`
```

action.yml
```diff
@@ -26,13 +26,18 @@ inputs:
...
+    default: 'true'
+    required: false
   save-always:
     description: 'Run the post step to save the cache even if another step before fails'
     default: 'false'
     required: false
     deprecationMessage: |
       save-always does not work as intended and will be removed in a future release.
-      A separate `actions/cache/restore` step should be used instead.
+      If you only want to control whether a new cache will be written use `save-always` instead.
```
Rossbro2, line 39:
```suggestion
      If you only want to control whether a new cache will be written use `save-on-success` instead.
```

helly25:
> What is the status of this? This could really improve our workflows!

An owner needs to approve the workflows and merge on green :-)

Rossbro2:
@salmanmkc any chance you could approve, merge & release this?  I've tested this on a fork and this is working great!

eirnym:
@GhadimiR hey, could you please review this PR?

action.yml
```diff
@@ -41,7 +46,7 @@ runs:
   using: 'node20'
   main: 'dist/restore/index.js'
   post: 'dist/save/index.js'
-  post-if: "success()"
+  post-if: "success() && github.event.inputs.save-on-success"
```
joshmgross, line 49: This will not work for the same reason that `save-always` was removed - #1452.

`post-if` does not have access to the input context, so this will always be false and completely break saving the cache.

To enable this type of workflow, you need to either:
1. Enable accessing the input context in `post-if` (requires `actions/runner` changes)
2. Use two separate actions for restore and save (this should already work)

eirnym:
Hi, thank you for your comments. Is it correct to say, that the only way is to write a custom action like it's done in Python setup?

GhadimiR:
@eirnym that _might_ be the best way to accomplish your use-case, if you want to apply complicated logic to your caching, however for precisely what is described above, it sounds like you could accomplish this with

```
actions/cache/save@v4
```
and 
```
actions/cache/restore@v4
```

You can see some examples that may help you [here](https://github.com/actions/cache/blob/main/caching-strategies.md#saving-cache) and combine that with [simple conditions](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions#example-only-run-job-for-a-specific-repository) based on branch name or whatever else you'd like to act on.

PR #1576, opened 2025-03-18
docs: fix condition for always saving

MatthijsBurgh:
## Description

The condition for the always save step was incorrect. I fixed the README.

## Motivation and Context

The example was not working for me.

## How Has This Been Tested?

Tested in my own repo.

## Types of changes

- [X] Documentation (add or update README or docs)

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/save/README.md
+++ b/save/README.md
@@ -106,7 +106,7 @@ jobs:
 
     - name: Always Save Prime Numbers
       id: cache-prime-numbers-save
-      if: always() && steps.cache-prime-numbers-restore.outputs.cache-hit != 'true'
+      if: ${{ always() && steps.cache-prime-numbers-restore.outputs.cache-hit != 'true' }}
       uses: actions/cache/save@v4
       with:
         key: ${{ steps.cache-prime-numbers-restore.outputs.cache-primary-key }}


joshmgross:
> The example was not working for me.

👋 @MatthijsBurgh could you share what failure you were seeing?

My understanding is that `${{ }}` is optional for `if` statements 


> The exception to this rule is when you are using expressions in an `if` clause, where, optionally, you can usually omit `${{` and `}}`. 
https://docs.github.com/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#about-expressions

MatthijsBurgh:
I had a better look at my CI. I was combining it with a check inside brackets. Which doesn't combine with logic outside of the brackets.

Doesn't work:

```yaml
if: ${{ runner.os != 'Windows' }} && steps.yarn-cache-restore.outputs.cache-hit != 'true' && always()
```

Works:

```yaml
if: ${{ runner.os != 'Windows'  && steps.yarn-cache-restore.outputs.cache-hit != 'true' && always() }}
if: runner.os != 'Windows'  && steps.yarn-cache-restore.outputs.cache-hit != 'true' && always()
```

So the PR can be closed.

PR #1639, opened 2025-08-08
Offer a strategy to reduce cache churn

atoulme:
## Description
This is a documentation change offering a new caching strategy meant to reduce cache churn by selecting on which runs a cache can be created.

## Motivation and Context
I am trying to check my reasoning and offer this best practice ahead of offering that `actions/cache` offers an option to perform cache save conditionally.

## How Has This Been Tested?
No tests - this is a doc change.

## Types of changes

## Checklist:

- [] I have added tests to cover my changes.

### DIFF
--- a/caching-strategies.md
+++ b/caching-strategies.md
@@ -304,3 +304,26 @@ steps:
   - name: Publish package to public
     run: ./publish.sh
 ```
+
+### Saving cache only if the build runs on the default branch
+
+Workflow runs can restore caches created in either the current branch or the default branch (usually `main`) [Reference](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache).
+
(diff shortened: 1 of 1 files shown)


OhanaAhyInspect-Owner:
- name: Cache
  uses: actions/cache@v4.2.4
  with:
    # A list of files, directories, and wildcard patterns to cache and restore
    path: 
    # An explicit key for restoring and saving the cache
    key: 
    # An ordered multiline string listing the prefix-matched keys, that are used for restoring stale cache if no cache hit occurred for key. Note `cache-hit` returns false in this case.
    restore-keys: # optional
    # The chunk size used to split up large files during upload, in bytes
    upload-chunk-size: # optional
    # An optional boolean when enabled, allows windows runners to save or restore caches that can be restored or saved respectively on other platforms
    enableCrossOsArchive: # optional, default is false
    # Fail the workflow if cache entry is not found
    fail-on-cache-miss: # optional, default is false
    # Check if a cache entry exists for the given input(s) (key, restore-keys) without downloading the cache
    lookup-only: # optional, default is false
    # Run the post step to save the cache even if another step before fails
    save-always: # optional, default is false

PR #1682, opened 2025-12-08
docs: Update checkout action version from v4 to v6

Chris53897:
get the latest version for actions/checkout


Link-:
Addressed in another PR

PR #1726, opened 2026-03-01
Fixed Cargo example not working for multiple archs

Melchizedek6809:
Added runner.arch to the cache key in the Cargo example since otherwise an arm runner might pull an x86 cache which breaks the build. I got the following error without this fix:

```
/home/runner/work/_temp/98f299c6-8f8f-4624-b041-55b989a1682b.sh: line 1: /home/runner/.cargo/bin/cargo: cannot execute binary file: Exec format error
```

## Description
It's just a change to the Rust/Cargo example since the action works quite well it's just that the key used in the example does not consider the architecture of the runner which at least for Rust/Linux is quite important.

## Motivation and Context
Should help others since they might run into the same error as I did when just copy-pasting the example code, I for example just blindly assumed that runner.os might be something like "ubuntu-24.04" or "ubuntu-24.04-arm" to distinguish between the different architecture.

## How Has This Been Tested?
Added the runner.arch part to the cache key for my Rust project and now 

## Types of changes

## Checklist:

Didn't run the test suite since I only changed a single line in the docs / EXAMPLES.md file.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -618,7 +618,7 @@ whenever possible:
       ~/.cargo/registry/cache/
       ~/.cargo/git/db/
       target/
-    key: ${{ runner.os }}-{{ runner.arch }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    key: ${{ runner.os }}-${{ runner.arch }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 ```
 
 ## Scala - SBT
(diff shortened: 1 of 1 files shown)


herrernst:
Shouldn't `runner.arch` be added to the cache key in all examples that build native objects, e.g. also Swift?

Melchizedek6809:
@herrernst Probably, but I haven't done any Swift development and I didn't want to add untested instructions

Melchizedek6809:
Could anyone review/merge this? Seems others ran into similar problems from the 2 coderabbit mentions 😅

PR #1757, opened 2026-05-12
Disable default save action for pull_request_target

### LINKED ISSUE #1756: Cache/save should not run automatically from a pull_request_trigger
The current behavior of actions/cache is to always do both a restore and a save step. Even though its documented to never run save from an untrusted source. 

Cache poisoning attacks were used against an nom package yesterday, the attack used a pull_request_target workflow to accomplish it. 

In order to be secure by default, I propose to update actions/cache to take its own recommended practices in account and to make the save step conditional on the trigger not being prt.

I understand there are probably still other ways to accomplish cache poinsening, but this would go a long way in reducing yet another angle of the pull_request_target triggers dangers.

Alternatively, cache/save or cache could use pick another branch to cache against, not the target branch, main in this case, but the source branch, pulls/123.

jessehouwing:
Disables the default save action on actions/cache when the event is a pull_request_target to prevent unnecessary executions. This change optimizes the workflow by ensuring that the save action only occurs for relevant events.

This plug a cache poisoning attack vector used in a number of supply chain attacks.

For users who REALLY want to save the cache on a PRT workflow, they can still use actions/cache/save dircetly. The secure by default principle should apply to the standard  actions/cache action.

Technically this *is* a breaking change. A major version bump might be advised. But I'd personally backport this to all previous versions of actions/cache instead.

Fixes https://github.com/actions/cache/issues/1756

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,8 @@ jobs:
 
 > **Note**
 > You must use the `cache` or `restore` action in your workflow before you need to use the files that might be restored from the cache. If the provided `key` matches an existing cache, a new cache is not created and if the provided `key` doesn't match an existing cache, a new cache is automatically created provided the job completes successfully.
+>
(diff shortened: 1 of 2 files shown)


jessehouwing:
It might be even better to solve this in the `@actions/cache` package itself and just turn off `/save` when the trigger is a PRT.

janellecanon01-lab:
yes

janellecanon01-lab (approved):

janellecanon01-lab (approved):

PR #1781, opened 2026-07-13
feat: adopt @actions/cache 6.2.0 (cache-mode skip + read-denied warning)

philip-gai:
## Description

Adopts `@actions/cache` 6.2.0 and rebuilds the vendored `dist/` bundles. The new toolkit version adds two cache-mode security behaviors:

- ACTIONS_CACHE_MODE skip: when the runner sets `ACTIONS_CACHE_MODE`, the action skips operations the effective mode does not permit. Restore is skipped for `none` and `write-only`; save is skipped for `none` and `read`. Each skip emits a single non-fatal info line and returns a cache miss on restore or a no-op on save. It never throws.
- Read/write denied warnings: when the cache service denies an operation because the token has no readable or writable scopes, the action surfaces a clear warning instead of a generic error, and continues without failing the run.

## Motivation and Context

The cache service can restrict what a job's token is allowed to do (read, write, both, or neither). Previously a denied read or write surfaced as an opaque error. These changes make the client honor the mode up front and report policy denials clearly.

Regression safety: when `ACTIONS_CACHE_MODE` is unset or set to an unrecognized value, behavior is identical to today. The skip logic only triggers on the explicit restrictive values.

Note on activation: the client-side skip depends on the runner exporting `ACTIONS_CACHE_MODE`, and the denial warnings depend on the cache service returning the corresponding policy responses. Until those roll out, this bump is behavior-preserving.

Toolkit change: actions/toolkit#2447 (published as `@actions/cache` 6.2.0).

## How Has This Been Tested?

- Full `@actions/cache` unit suite passing in the toolkit repo (tsc, jest, prettier, eslint clean).
- End-to-end validation on hosted runners using a test build of this action pinned to the unreleased toolkit code:
  - Env-driven matrix over `none`, `read`, `write`, `write-only`, unknown, and unset confirmed the expected skip or proceed outcome for both restore and save, with exactly one info line per skip and no failures.
  - With cache-mode enforcement enabled server-side, restore under a read-denied token returned the read-denied policy response and the action degraded to a warning plus cache miss; save under a write-denied token returned the write-denied policy response and degraded to a warning. Allowed and unset modes cached and restored normally.

## Types of changes

## Checklist:

### DIFF
--- a/.licenses/npm/@actions/cache.dep.yml
+++ b/.licenses/npm/@actions/cache.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: "@actions/cache"
-version: 6.1.0
+version: 6.2.0
 type: npm
 summary: Actions cache lib
 homepage: https://github.com/actions/toolkit/tree/main/packages/cache
--- a/dist/restore-only/index.js
+++ b/dist/restore-only/index.js
@@ -39021,7 +39021,7 @@ module.exports = { version: packageJson.version }
 /***/ 4012:
 /***/ ((module) => {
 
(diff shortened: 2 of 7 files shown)


boxofyellow (approved):
LGTM: :shipit:

boxofyellow (approved):
LGTM: :shipit:

PR #1797, opened 2026-09-04
fix: always land in GCS, backfill GitHub hits, remember paths for the post step

Jasperhino:
## Why

In cula-platform the pnpm store cache (`Linux-pnpm-store-v3-…`, 1.6 GB) has never been in GCS: every job logs "Cache not found in GCS, falling back to GitHub cache" and downloads it from the GitHub cache instead. Three things in this action combine to make that permanent:

1. **A GitHub-fallback hit was treated like a GCS hit.** `saveImpl` skips saving on an exact key match regardless of which backend answered. So once GCS missed a key a single time, the GitHub copy satisfied every later job and GCS never received it.
2. **The post step could not see its `path`.** When this action runs inside a composite action that is itself inside another one (cula-platform: job → `setup-project` → `install-dependencies` → `Cula-Technologies/cache`), the post step's `path: ${{ steps.store.outputs.path }}` evaluates to empty. Every such job logged `Input required and not supplied: path` and saved nothing. That is 25 warnings in a single develop run.
3. **A GCS save failure did not fall back.** `saveCache` logged "falling back to GitHub cache" on an empty GCS result and then returned `-1` without doing so. (The one job that did populate the GitHub entry hit the *thrown* path instead, after `google-github-actions/auth`'s cleanup had deleted the credentials file before the save ran; that ordering is fixed on the caller side.)

## What

- **Restore records the backend** that served the hit in state (`CACHE_SOURCE`). **Save backfills GCS** when an exact hit came from the GitHub cache and GCS is configured, with the GitHub save disabled for that call (it would only fail on the existing entry). A GCS-served hit still skips as before.
- **Restore records the resolved paths** in state (`CACHE_PATHS`). **Save uses them when the `path` input is empty**, which is the nested-composite post-step case. The `restore` and `save` split actions are unaffected (the null state provider never exposes these).
- `saveCache` falls back to the GitHub cache on both the empty-result and the thrown-error path unless told not to.
- README: a "Google Cloud Storage backend" section describing restore order, save, backfill, the path fallback, and the auth-ordering caveat.
- Tests: the `getState` mocks were positional `mockImplementationOnce` chains that depended on the exact number and order of reads (and leaked across tests; several passed only in-suite). They are keyed by state name now. Three new tests cover GCS hit / GitHub hit backfill / path fallback; count assertions updated for the two extra states.
- `dist/` rebuilt with `npm run build`, line endings kept as CRLF to match the committed files.

## Verified

`npm run lint`, `prettier --check`, `jest` (84 passing), `npm run build`. Not verified: a real run of the rebuilt dist in CI; the follow-up in cula-platform is to bump the pinned SHA in `install-dependencies`, `cache-dependencies` and the workflows once this is merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Y5UfALF3nKcwCLWPikbnJn


Jasperhino:
Opened against the wrong repository by mistake (fork-internal change). Closing.

